PR #1859, opened 2024-06-21
Add VSCode UI tests with webdriver.io

kgartland-rstudio:
## Intent

This PR adds webdriver.io vscode UI testing. The first test includes adding a new Deployment and checking the config.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Checklist


test/vscode-ui/test/specs/extension.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  let workbench: any;
+  let input: any;
+
+  before(async () => {
+    workbench = await browser.getWorkbench();
+    input = await $(".input");
+  });
+
+  it("open extension", async () => {
+    browser.$("aria/Posit Publisher").waitForExist({ timeout: 30000 });
```
sagerb, line 44: What is your approach for the selection criteria's you are using? It would be helpful to identify attributes you can count on not being modified or lost (at least we can be aware of the usage), so that the tests won't need to be modified as much with code changes in the underlying interfaces. Have you just been looking what is set now? Should we start putting a `data-` attribute on the web components  that you may need to access?

test/vscode-ui/test/specs/extension.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    const extension = await browser.$("aria/Posit Publisher");
+    await expect(extension).toExist();
+    await extension.click();
+  });
+
+  it("create first deployment", async () => {
+    await switchToSubframe();
+    // initialize project via button
+    $(".add-deployment-btn").waitForExist({ timeout: 30000 });
+    const init = await $(".add-deployment-btn");
```
sagerb, line 56: This one is a class, which is very fragile...

test/vscode-ui/test/specs/extension.spec.ts
sagerb, line 44: I am also always worried about timeouts.. did you set this one very high because of the unknown nature of our initialization? Could you loop until it appears instead?

sagerb (approved):
This looks like a really good start. I have a few questions regarding the selectors being used and their targets, but I don't see any reason this can't get merged in as is. Those selectors can be addressed as we move forward, as long as these tests are not going to block our CI when they fail?

test/vscode-ui/test/specs/extension.spec.ts
kgartland-rstudio: > What is your approach for the selection criteria's you are using? It would be helpful to identify attributes you can count on not being modified or lost (at least we can be aware of the usage), so that the tests won't need to be modified as much with code changes in the underlying interfaces. Have you just been looking what is set now? Should we start putting a data- attribute on the web components that you may need to access?

`data-` attributes would be great.  I was following the best practices for webdriver here, they say the aria labels should be used for selectors. 
https://webdriver.io/docs/selectors

If we have `data` selectors I think that would be preferable though. I can certainly add them but was having a hard time finding where they should go.

> Those selectors can be addressed as we move forward, as long as these tests are not going to block our CI when they fail?

Right now, they'd be running every in CI on every PR. I can make them run only at night though.

Issue #1851, opened 2024-06-18
Support specifying a path for API requests


dotNomad:
Several APIs need to be able to specify a `path` to support nested folder deploying flows in the VSCode extension.

The proposed solution is to have:
- an optional query param `?dir=path`
- pass an optional path in the body of the request for applicable responses

The APIs that need adjustments:
- `GET /api/configurations`
- All `/api/configurations/$name` API endpoints since non-root named Configurations will need a path alongside to be identified
- ALL `/api/packages` API endpoints since configurations or file creation needs to know the non-root directory
- `POST and DEL /api/deployments/$name` API endpoints need a path for non-root deployments and configurations

**Note**

The `entrypoint` query parameter mentioned in #1847 will not work alongside the `dir` query param.

---

Part of the #1849 epic

dotNomad:
@mmarchetti I'm taking a look and I'm not seeing support for `dir` in the `/api/packages/python/scan` or `/api/packages/r/scan` endpoints. Am I missing it?

mmarchetti:
> Am I missing it?

Nope, I missed it. See https://github.com/posit-dev/publisher/pull/1919

Issue #1844, opened 2024-06-18
Add `parentDir` to API resources


dotNomad:
Currently we have `configurationRelPath` and similar, which has a value of `.posit/publish/python-staging.toml` for example.

For Nested Project Deploying and the strings we will want to display we will need a new attribute in:
- Configurations
- Deployment Records
- ConfigurationDetails (returned by the Inspect API and passed to `PUT /api/configurations/${NAME}`)

`parentDir` will contain the relative-to-the-workspace-root path of the directory containing the `.posit` directory.

---

Part of the #1843 epic.

PR #1873, opened 2024-06-26
Add subdirectory support to deployments API

mmarchetti:
## Intent

This PR adds support for an optional `dir` query parameter to the deployment API endpoints:

* GET /api/deployments
* GET /api/deployments/$NAME
* POST /api/deployments
* POST /api/deployments/$NAME
* PATCH /api/deployments/$NAME
* DELETE /api/deployments/$NAME

The directory is relative to the workspace root directory, and cannot escape it (the API should return a 400 if you provide a path that would be outside of the workspace root).

It also populates the returned configuration DTO with a new `projectDir` field that contains the subdirectory (again, relative to the workspace root).

This is part of https://github.com/posit-dev/publisher/issues/1844 and part of https://github.com/posit-dev/publisher/issues/1851.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Existing tests have been updated to assert that the projectDir is correct when no dir is specified. New tests have been added specifically to test the subdirectory case for each API.

## Directions for Reviewers

Run the agent in a parent directory (one or more levels up from your deployable projects). Access deployment info from subdirectories using the various APIs:

```
$publisher ui -vv --listen=localhost:9001 &
curl localhost:9001/api/deployments?dir=test/sample-content/quarto-proj-py | jq
curl localhost:9001/api/deployments?dir=test/sample-content/quarto-proj-none | jq
```

The `POST /api/deployments/$NAME` endpoint should load the configuration named in the deployment from the same directory path; verify that.

Also run the extension and regression test basic publishing (you can't use subdirectories in the extension yet, but everything should default to the current behavior).

### DIFF
--- a/internal/services/api/delete_deployment.go
+++ b/internal/services/api/delete_deployment.go
@@ -16,8 +16,13 @@ import (
 func DeleteDeploymentHandlerFunc(base util.AbsolutePath, log logging.Logger) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		name := mux.Vars(req)["name"]
-		path := deployment.GetDeploymentPath(base, name)
-		err := path.Remove()
+		projectDir, _, err := ProjectDirFromRequest(base, w, req, log)
+		if err != nil {
+			// Response already returned by ProjectDirFromRequest
+			return
+		}
+		path := deployment.GetDeploymentPath(projectDir, name)
+		err = path.Remove()
 		if err != nil {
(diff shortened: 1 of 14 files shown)


dotNomad (approved):
Tested out the APIs and they all worked as expected. Smoke tested the extension and all of the operations worked as expected as well

Issue #1848, opened 2024-06-18
Add Deployment selection button to open Entrypoint files


dotNomad:
On hold waiting for the API changes in #1847

This PR will add the publish button to the open entrypoint files.

To determine if the button should be added we can call the new `/api/inspect?entrypoint=path` API. If any results are in the response we know its a valid entrypoint.

When the button is clicked:
- Call the new `/api/deployments?entrypoint=path` endpoint
- Guide the user to create a new Deployment / Configuration with that file as the entrypoint if one doesn’t exist
- Select the Deployment / Configuration in the sidebar with that entrypoint if `length === 1`
- Prompt the user to select from the multiple Deployments with the given entrypoint

---

Part of the #1846 epic.

PR #1883, opened 2024-06-26
Add document tracker to set `activeFileEntrypoint` context

dotNomad:
This PR adds a new context to the extension that tracks if the active document is an entrypoint with a `ContentType` that is known (aka not `Unknown`)

If the active document is an entrypoint - found via `/api/inspect` - we show a button which currently is non-functional, but will lead a user through deploying.

Note that `licenses.md` didn't need to change since `vscode-uri` is already included from dependencies.

## Intent

Start of #1848

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach here was to create a tracker that created all of the listeners needed itself, tracked the documents it needed to, and updated the context.

This way the extension could create a `DocumentTracker` and add it to its `diposables` with no other interaction between the two.

The `DocumentTracker` handles which documents need to be tracked via VSCode listeners, and `TrackedEntrypointDocument`s handle updating themselves if they are or are becoming the active document.

Everything else was broken off as a utility function.

## Directions for Reviewers

Test the entrypoint context in a few cases:
- open a non-entrypoint file
- open an entrypoint file
- open a file outside of the workspace entirely (I do this by running `code some-file.txt` from the Desktop for example)

Then do the same from a folder at a higher level. The `dir` and `entrypoint` query params should take care of the nested folder part for us.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -251,9 +251,22 @@
         "title": "View Content Log on Connect",
         "icon": "$(output)",
         "category": "Posit Publisher"
+      },
+      {
+        "command": "posit.publisher.deployWithEntrypoint",
+        "title": "Deploy with this Entrypoint",
+        "icon": "$(cloud-upload)",
+        "category": "Posit Publisher"
       }
     ],
     "menus": {
+      "editor/title": [
+        {
+          "command": "posit.publisher.deployWithEntrypoint",
+          "group": "navigation",
+          "when": "posit.publish.activeFileEntrypoint == true"
+        }
+      ],
       "view/title": [
         {
           "command": "posit.publisher.configurations.add",
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -698,6 +698,7 @@
     "eventsource": "^2.0.2",
     "get-port": "5.1.1",
     "mutexify": "^1.4.0",
-    "retry": "^0.13.1"
+    "retry": "^0.13.1",
+    "vscode-uri": "^3.0.8"
   }
 }
--- a/extensions/vscode/src/entrypointTracker.ts
+++ b/extensions/vscode/src/entrypointTracker.ts
@@ -14,6 +14,7 @@ import { useApi } from "src/api";
 import { getPythonInterpreterPath } from "src/utils/config";
 import { isActiveDocument } from "src/utils/files";
 import { hasKnownContentType } from "src/utils/inspect";
+import { getSummaryStringFromError } from "src/utils/errors";
 
 const ACTIVE_FILE_ENTRYPOINT_CONTEXT = "posit.publish.activeFileEntrypoint";
 
@@ -24,20 +25,29 @@ const ACTIVE_FILE_ENTRYPOINT_CONTEXT = "posit.publish.activeFileEntrypoint";
  * @returns If the text document is an entrypoint
  */
 async function isDocumentEntrypoint(document: TextDocument): Promise<boolean> {
-  const api = await useApi();
-  const python = await getPythonInterpreterPath();
+  try {
+    const api = await useApi();
+    const python = await getPythonInterpreterPath();
 
-  // If the file is outside the workspace, it cannot be an entrypoint
-  const dirname = uriUtils.dirname(document.uri);
-  if (dirname === undefined) {
+    // If the file is outside the workspace, it cannot be an entrypoint
+    const dirname = uriUtils.dirname(document.uri);
+    if (dirname === undefined) {
+      return false;
+    }
+
+    const response = await api.configurations.inspect(python, {
+      dir: workspace.asRelativePath(dirname),
+      entrypoint: uriUtils.basename(document.uri),
+    });
+    return hasKnownContentType(response.data);
+  } catch (error: unknown) {
+    const summary = getSummaryStringFromError(
+      "entrypointTracker::isDocumentEntrypoint",
+      error,
+    );
+    window.showInformationMessage(summary);
     return false;
   }
-
-  const response = await api.configurations.inspect(python, {
-    dir: workspace.asRelativePath(dirname),
-    entrypoint: uriUtils.basename(document.uri),
-  });
-  return hasKnownContentType(response.data);
 }
 
 /**
--- a/extensions/vscode/src/entrypointTracker.ts
+++ b/extensions/vscode/src/entrypointTracker.ts
@@ -1,18 +1,11 @@
-import {
-  Disposable,
-  TextDocument,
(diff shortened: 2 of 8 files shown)


extensions/vscode/src/utils/files.ts
```diff
@@ -231,3 +233,34 @@ export function splitFilesOnInclusion(
...
+    // File is outside of the workspace
+    return undefined;
+  }
+
+  const dirname = uriUtils.dirname(uri);
+
+  // If the file is in the workspace root, return "."
+  if (dirname.fsPath === workspaceFolder.uri.fsPath) {
+    return ".";
+  }
```
dotNomad, line 262: `workspace.asRelativePath(dirname)` returns a path relative to the workspace folder, but if the path _IS_ the workspace folder then it will return an absolute path.

That doesn't work for us, so here I check to see if the file system path is the same and return a `"."`.

extensions/vscode/src/entrypointTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+import { getSummaryStringFromError } from "src/utils/errors";
+
+/**
+ * Determines if a text document is an entrypoint file.
+ *
+ * @param document The text document to inspect
+ * @returns If the text document is an entrypoint
+ */
+async function isDocumentEntrypoint(document: TextDocument): Promise<boolean> {
+  const dir = relativeDir(document.uri);
```
dotNomad, line 27: This was quite tricky digging into what VSCode could do for us and what else is out there. I broke it out into its own utils function so we can get a `dir` path for any API requests.

extensions/vscode/src/entrypointTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+    const isEntrypoint = await isDocumentEntrypoint(document);
+    return new TrackedEntrypointDocument(document, isEntrypoint);
+  }
+
+  /**
+   * Sets the file entrypoint context with this as the active file.
+   * @param options Options for the activation
+   * @param options.forceUpdate Whether to force the entrypoint to update
+   */
+  async activate(options?: { forceUpdate?: boolean }) {
```
dotNomad, line 77: The API calls only happen when we create a `TrackedEntrypointDocument` or when it is activated and needs an updated.

This occurs when the document is saved, or the document is re-opened.

`TrackedEntrypointDocument`s don't update if the file is open in VSCode, another file becomes active, then the original file is active again. That is because nothing has changed since the first API call.

extensions/vscode/src/entrypointTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Listener function for the closing of a text document.
+   * Stops the document from being tracked.
+   *
+   * @param document The closed document
+   */
+  onTextDocumentClosed(document: TextDocument) {
+    this.removeDocument(document);
+  }
```
dotNomad, line 188: Documents are removed when closed since another editor could potentially change the document. This way when a document is re-opened we recheck just in case.

extensions/vscode/src/entrypointTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * Determines if a text document is an entrypoint file.
+ *
+ * @param document The text document to inspect
+ * @returns If the text document is an entrypoint
+ */
+async function isDocumentEntrypoint(document: TextDocument): Promise<boolean> {
+  const dir = relativeDir(document.uri);
+  // If the file is outside the workspace, it cannot be an entrypoint
+  if (dir === undefined) {
```
dotNomad, line 29: If we don't want anything but `.` level entrypoints for now we can easily add that check here.

sagerb: True, but I don't think we need to per our implementation plan.

extensions/vscode/src/extension.ts
```diff
@@ -100,6 +101,14 @@ export async function activate(context: ExtensionContext) {
     ),
   );
   setStateContext(PositPublishState.initialized);
+
+  context.subscriptions.push(
+    new DocumentTracker(),
+    commands.registerCommand(Commands.DeployWithEntrypoint, () => {
+      commands.executeCommand(Commands.HomeView.Focus);
+      console.log("'Deploy with this Entrypoint' button hit!");
```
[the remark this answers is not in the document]
sagerb: Easily built upon! Great!

sagerb (approved):
This looks really good and I think I have a handle on it.

If you could update the command to accept the document URI parameter that will be passed in, per my comment, and then output it in the console log, I think you'll have this PR at a good state. I'll approve it so you aren't blocked by me. Thanks!

extensions/vscode/src/extension.ts
```diff
@@ -100,6 +101,14 @@ export async function activate(context: ExtensionContext) {
     ),
   );
   setStateContext(PositPublishState.initialized);
+
+  context.subscriptions.push(
+    new DocumentTracker(),
+    commands.registerCommand(Commands.DeployWithEntrypoint, () => {
```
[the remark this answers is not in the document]
dotNomad: I wasn't sure what exactly was going to be passed - a `TextDocument`, `TextEditor`, but after some digging it is a `Uri` and that is typed, and logged now in https://github.com/posit-dev/publisher/pull/1883/commits/69d7a74721bbb41c08691c7b3593e78738c896bf

extensions/vscode/package.json
```diff
@@ -251,9 +251,22 @@
         "title": "View Content Log on Connect",
         "icon": "$(output)",
         "category": "Posit Publisher"
+      },
+      {
+        "command": "posit.publisher.deployWithEntrypoint",
+        "title": "Deploy with this Entrypoint",
+        "icon": "$(cloud-upload)",
```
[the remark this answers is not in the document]
dotNomad: I tested this out and it does work, and doesn't look too bad even though it is being scaled down from `450x450` to `16x16`

CleanShot 2024-07-03 at 14 25 44@2x

However since it is a `png` it doesn't get colored using the active theme. So even if I changed it to be white/gray/etc it would look incorrect depending on the theme used.

Personally I'd prefer to avoid the non-native look until we get that fixed.

If others disagree, it can be easily changed while I'm out with the below example:

```json
      {
        "command": "posit.publisher.deployWithEntrypoint",
        "title": "Deploy with this Entrypoint",
        "icon": "assets/img/color/posit-publisher.png",
        "category": "Posit Publisher"
      }
```

PR #1902, opened 2024-06-28
Select entrypoint first in new deployment input

### LINKED ISSUE #1784: Prompt for entrypoint before choosing server via Credential
https://docs.google.com/document/d/1nlHr4pQP_CZKCVxtVA1Ba_fZtuAJRC-BcIxhS7JavEo/edit?disco=AAABOjDGMW4

sagerb:
## Intent

This moves the entry point selection and the title entry up to the beginning of the New Deployment `multiStepInput`.

Resolves #1784 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The diff is strange because of the similar code blocks. I updated the `steps` object to have the new step number calculations and then moved the `inputEntryPointSelection` and the `inputTitle` entry methods to occur ahead of the credential collection.

## Directions for Reviewers

To validate, you should create deployments on the following, making sure a deployment can be created and that the step numbers update appropriately to your selections:
[ ] a project with a single entry point, using an existing credential 
[ ] a project with a single entry point, creating a new credential
[ ] a project with multiple entry points, using an existing credential
[ ] a project with multiple entry points, using a new credential

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -60,134 +60,6 @@ type possibleSteps = {
 };
 
 const steps: Record<string, possibleSteps | undefined> = {
-  pickCredentials: {
-    noCredentials: {
-      singleEntryPoint: {
-        step: 0,
-        totalSteps: 4,
-      },
-      multipleEntryPoints: {
-        step: 0,
-        totalSteps: 5,
-      },
-    },
-    newCredentials: {
-      singleEntryPoint: {
-        step: 1,
-        totalSteps: 5,
-      },
-      multipleEntryPoints: {
-        step: 1,
-        totalSteps: 6,
-      },
-    },
-    existingCredentials: {
-      singleEntryPoint: {
-        step: 1,
-        totalSteps: 2,
-      },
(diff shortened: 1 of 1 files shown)


dotNomad (approved):
Fantastic. Tested all the different options, and the steps always looked right. 

The only weirdness is if you select "Create a New Credential" then click back the steps still show `3/6`

CleanShot 2024-07-01 at 11 26 02@2x

I think this is still valid though.

sagerb:
> Fantastic. Tested all the different options, and the steps always looked right.
> 
> The only weirdness is if you select "Create a New Credential" then click back the steps still show `3/6`
> 
> CleanShot 2024-07-01 at 11 26 02@2x
> 
> I think this is still valid though.

It is difficult to think what is correct here... With the current option being create new, it is valid that it will take 6 steps, but if you change it to using an existing, we don't modify the steps dynamically in order to show that it now will only be 2 or 3 (whichever it is). It would probably be a GOOD change, although I'm unsure how expensive it would be to extend the framework to support it. Let's wait until we get some feedback that it is confusing...

PR #1903, opened 2024-06-28
Only update version in package.json in CI

mmarchetti:
## Intent

When we build a release package in CI (on a tag), our scripts update the version number in the extension's `package.json`. This PR limits this action to only happen when the `CI` environment variable is `true`, so it won't happen locally.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/scripts/set-version.py
+++ b/extensions/vscode/scripts/set-version.py
@@ -1,12 +1,17 @@
 #!/usr/bin/env python3
 
 import json
+import os
 import re
 import sys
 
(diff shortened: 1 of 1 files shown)


sagerb (approved):
Looks great!

dotNomad (approved):
Very nice change 👍

sagerb:
And Verified!

Issue #1906, opened 2024-06-28
Files needing copyright headers not failing CI build


sagerb:
I discovered this w/ files from `main` on June 27th.

From a quick examination, I don't believe any of the workflow steps are calling `just lint`.

PR #1907, opened 2024-06-28
Show restore status in active publishing log display

### LINKED ISSUE #1795: Add restore package progress into Publisher Log
Like what was done for progress indicator window (lower-right popup)

sagerb:
## Intent

Resolves #1795 

This PR updates the publishing log to show environment restoration status when a deployment is active.

It also updates the publishing progress dialog to show active message terminology rather than the older (considered inactive) verbs.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The event message utility method of `eventMsgToString` was updated to provide a verb appropriate to the actual message (in that the verb should reflect if that phase has been completed or is in progress - inActive or active). In doing this, a Record structure was switched to a Map, to better match our other implementations.

To be consistent between the `deployProgress` and the `Publisher Log`, the deployProgress component was switched to subscribe to and display the more generic message of `restoreEnv` rather than the language-specific messages of `restorePythonEnv` and `restoreREnv`.

In addition, two `go` files were updated when `just fix` was run against the branch. These files were unchanged and present within `main` for some time, without being flagged as missing their copyright headers. I'm not currently sure what's going on, but I'll add an issue to the triage board: (#1906)

## Directions for Reviewers

The best way to see the effect is to reset your test environment so that the Python and R environment restorations will require actual installations to occur. 

Should verify deployment progress display and publishing log when a deployment is active as well as completed, for both R and Python projects.


extensions/vscode/src/views/logs.ts
```diff
@@ -229,6 +233,17 @@ export class LogsTreeDataProvider implements TreeDataProvider<LogsTreeItem> {
...
         },
       );
+
+      // Register for some specific events we need to handle differently
+      this._stream.register(
+        "publish/restoreEnv/status",
+        (msg: EventStreamMessage) => {
+          const stage = this.stages.get("publish/restoreEnv");
+          if (stage && isPublishRestoreEnvStatus(msg)) {
+            stage.activeLabel = `${RestoringEnvironmentLabel} - ${restoreMsgToStatusSuffix(msg)}`;
```
dotNomad, line 243: I'm having a bit of trouble understanding what is happening here.

extensions/vscode/src/api/types/events.ts
```diff
@@ -136,15 +136,29 @@ export interface EventSubscriptionTargetCallbackMap {
...
 
+export const restoreMsgToStatusSuffix = (
+  msg:
+    | PublishRestorePythonEnvStatus
+    | PublishRestoreREnvStatus
+    | PublishRestoreEnvStatus,
+) => {
+  let suffix = msg.data.name;
+  if (msg.data.version) {
+    suffix = `${suffix} (${msg.data.version})...`;
```
dotNomad, line 147: I'm failing to see the difference here. I'm not 100% sure this is code is affecting this, but it appears to be with the space but no `msg.data.version` text?

CleanShot 2024-07-01 at 14 35 54@2x

dotNomad (commented):
I had a few questions on this one.

extensions/vscode/src/api/types/events.ts
sagerb: Those messages are coming directly from the Agent with a different event type, so they are not affected by the changes in this PR. They are occurring in the Run Content Section, if I am seeing the same thing as you are:

>local_id=_CkD5fpTfG7mcL0K source=server.log event_op=publish/runContent
time=2024-07-01T14:50:15.785-07:00 level=INFO msg="Determining session server location ..." >local_id=_CkD5fpTfG7mcL0K source=server.log event_op=publish/runContent
time=2024-07-01T14:50:15.785-07:00 level=INFO msg="Connecting to session server http://127.0.0.1:37971 ..."

extensions/vscode/src/views/logs.ts
sagerb: We're updating the label for the Restore Environment stage when we receive a status event for it? To do this, we are updating the child tree item for the stage, which then causes a refresh, which repaints the change. Perhaps we should zoom real quick...

extensions/vscode/src/api/types/events.ts
dotNomad: These three lines were under the "Restore Environment" step not the "Run Content" step.

sagerb:
I've updated this branch with an additional change for Log pluralization (https://github.com/posit-dev/publisher/pull/1907/commits/529f43b3f78ffaa7deb4d569e7739f507e09b1eb), as well as now the expansion state of each stage is maintained as the tree is refreshed with new log entries. (https://github.com/posit-dev/publisher/pull/1907/commits/dec3f1c169e51c284ee5312cc05059efb9de51c0)

This code continues to be very inefficient, as the message data being received is added to the stages object, with then the code calling refresh, which causes every stage and every one of the log messages for each stage to be constructed over and over again each time GetChildren is called by VSCode's infrastructure. This reconstruction of the entire set of TreeItems occurs once for every log message received, plus a couple for the stable stages. While I decided not to implement it here, I have created a new triage item that will need to address this very inefficient approach. I would suggest the code create each TreeItem once, and simply keep passing references back into the VSCode callbacks for getChildren. See #1921

dotNomad (approved):
Pulled it down and messed about with the logs. I wasn't able to get the open and closing expansion state issue with the new changes 🎉

PR #1908, opened 2024-06-28
Publishing Log: expand section on error and add bullets to log message lines

### LINKED ISSUE #1790: Automatically expand publisher log section in Error, when fails
I observed some awkwardness in what to do when the deployment failed, and how to find further details on what really failed. While this will be addressed more completely when we do enhanced error handling, it would be an easy improvement to make sure the section which is in error is opened at the error (if possible).

sagerb:
## Intent

Resolves #1790 

This PR will cause Publishing Log sections to automatically expand themselves upon receiving a failure message.

I've also taken the opportunity to add a small bullet to the beginning of each log message, as well as suppressing newline characters received in the log messages, since we can't display those correctly.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Simple set of changes. When encounter failure and viewing publisher log, you'll now see something like this:

2024-06-28 at 3 01 PM

Final success or failure messages do not have icons as a prefix, as shown in the following screenshot.

2024-06-28 at 3 10 PM

NOTE: This behavior only occurs when the log is actively being updated. If you open the log after the publishing has occurred, the node will be closed by default.

## Directions for Reviewers

Easiest way to force failure that I've found is to either specify the wrong python version (which will fail fast), or add a bogus package requirement to your requirements.txt file.

Please verify:
[ ] a successful deployment - only impact should be seeing bullets in the log messages except for the final success message.
[ ] a failed deployment - which should result in the publishing log section which encountered the error to be automatically expanded (with then the log messages having bullets visible). In this case, the final failure message should not have a bullet.

### DIFF
--- a/extensions/vscode/src/views/logs.ts
+++ b/extensions/vscode/src/views/logs.ts
@@ -17,7 +17,11 @@ import {
 
 import { EventStream, displayEventStreamMessage } from "src/events";
 
-import { EventStreamMessage } from "src/api";
+import {
+  EventStreamMessage,
+  isPublishFailure,
+  isPublishSuccess,
+} from "src/api";
 import { Commands, Views } from "src/constants";
 
 enum LogStageStatus {
@@ -368,7 +372,9 @@ export class LogsTreeLogItem extends TreeItem {
     }
     super(displayEventStreamMessage(msg), state);
     this.tooltip = JSON.stringify(msg);
-    this.iconPath = new ThemeIcon("debug-stackframe-dot");
+    if (!isPublishSuccess(msg) && !isPublishFailure(msg)) {
+      this.iconPath = new ThemeIcon("debug-stackframe-dot");
+    }
 
     if (msg.data.dashboardUrl !== undefined) {
       this.command = {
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/views/logs.ts
```diff
@@ -362,8 +367,14 @@ export class LogsTreeLogItem extends TreeItem {
     msg: EventStreamMessage,
     state: TreeItemCollapsibleState = TreeItemCollapsibleState.None,
   ) {
+    if (msg.data.message) {
+      msg.data.message = msg.data.message.replaceAll("\n", " ");
```
dotNomad, line 371: I still think we should be adjusting this in the backend, but totally fine leaving this in for now.

dotNomad (approved):
Fantastic. The dots make a big difference for parsing this.

extensions/vscode/src/views/logs.ts
sagerb: I'm not sure. It really is a presentation layer decision.. for the CLI, a newline character is perfectly fine. I originally thought about showing it as an additional tree item in the view, but it was too complex with the way we are creating those. I'll add a triage item just so we can discuss and everyone can think about it... but I don't expect that we'll plan on having it resolved anytime soon.

PR #1915, opened 2024-07-01
Add dir parameter to inspect api

### LINKED ISSUE #1844: Add `parentDir` to API resources

### LINKED ISSUE #1851: Support specifying a path for API requests

mmarchetti:
## Intent

This PR adds the new `dir` parameter to the `/api/inspect` endpoint, and adds `projectDir` to the response body. This changes the shape of the response, so there is a minimal frontend change to consume the new type.

Fixes #1851 
Fixes #1844

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

Ensure that creating a new configuration (via inspection works. We do this in 3 places:
* Creating a new deployment via the + button in the Home view
* Creating a new configuration via the + button in the Configurations view
* In the Select Active Configuration for Deployment command (access via the ellipsis menu in the Home view), if you select Create a New Configuration instead of choosing an existing one.


extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -266,7 +266,7 @@ export async function newDeployment(
   let credentialListItems: QuickPickItem[] = [];
 
   let entryPointListItems: QuickPickItemWithIndex[] = [];
-  let configDetails: ConfigurationDetails[] = [];
+  let configDetails: ConfigurationInspectionResult[] = [];
```
sagerb, line 269: I like what you did in `selectConfig.ts` where you renamed `configDetails` to `inspectionResults`. Can you update this to be refactored in the same way? I've been trying to keep the code as similar as possible, so that we can refactor it all into shared code, sometime in the future...

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -903,10 +905,10 @@ export async function newDeployment(
       );
       return;
     }
-    selectedConfigDetails.title = state.data.title;
+    selectedConfigDetails.configuration.title = state.data.title;
     const createResponse = await api.configurations.createOrUpdate(
       configName,
-      selectedConfigDetails,
+      selectedConfigDetails.configuration,
```
sagerb, line 911: With the rename of `configDetails` to `inspectionDetails`, should we also rename `selectedConfigDetails` to `selectedInspectionDetails`?

extensions/vscode/src/multiStepInputs/selectConfig.ts
```diff
@@ -389,17 +391,18 @@ export async function selectConfig(
     // Create the Config File
     try {
       const configName = await untitledConfigurationName();
-      const selectedConfigDetails = configDetails[state.data.entryPoint.index];
+      const selectedConfigDetails =
```
sagerb, line 394: same comment as above:

>With the rename of configDetails to inspectionDetails, should we also rename selectedConfigDetails to selectedInspectionDetails?

sagerb (approved):
Do the Go tests for the API need to be enhanced in any way to handle this change?

I had a few other requests for a bit more renaming, just to keep the code consistent and easy to maintain. Other than that, this looks good! I'll approve it and leave it to you for any additional changes.

sagerb:
I've verified that this code works great!

mmarchetti:
> Do the Go tests for the API need to be enhanced in any way to handle this change?

Ideally, yes. This endpoint is currently wiithout unit tests.

Issue #1845, opened 2024-06-18
Show Relative Paths in the UI


dotNomad:
~On hold waiting for the backend changes in #1844~

Once we have access to the `parentDir` attribute we can use it in our VSCode UI to better ground the user and ourselves as we continue the nested folder deploying work.

If the `parentDir` is not `.` we should show it in:
- The sidebar where we display the entrypoint file for the selected Deployment
- In the Deployment VSCode select inputs
- In the Configuration select inputs
- In the ConfigurationDetails select inputs (inspection when selecting a entrypoint)

---

Part of the #1833 epic.

dotNomad:
Marking this as closed. Reviewed the places where we do (and don't) show paths in the extension and didn't find anything that didn't make sense in nested and non-nested workspace roots.

PR #1922, opened 2024-07-02
Update frontend API client to reflect API changes

dotNomad:
This updates the API module in the frontend to support the new `dir` and `entrypoint` query params.

It also updates the types to support the new `projectDir` attribute.

The unused `accounts` API resource has also been removed.

## Intent

Part of #1845 and #1848

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling


extensions/vscode/src/api/resources/ContentRecords.ts
```diff
@@ -55,35 +64,41 @@ export class ContentRecords {
     targetName: string,
     accountName: string,
     configName: string = "default",
+    params?: { dir?: string },
   ) {
-    const params = {
+    const data = {
```
dotNomad, line 69: Changes the name here and above from `params` to `data` to better reflect the naming Axios suspects and to avoid confusion when using query params.

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -20,38 +20,48 @@ export class Configurations {
   // 200 - success
   // 404 - not found
   // 500 - internal server error
-  get(configName: string) {
+  get(configName: string, params?: { dir?: string }) {
     const encodedName = encodeURIComponent(configName);
     return this.client.get<Configuration | ConfigurationError>(
       `/configurations/${encodedName}`,
+      { params },
```
sagerb, line 27: Do we need to encode the param dir value, before sending it, in ALL of the API requests? (Asking for this one only, but asking for all API requests).

extensions/vscode/src/api/resources/Packages.ts
```diff
@@ -43,15 +43,23 @@ export class Packages {
...
-  createPythonRequirementsFile(python?: string, saveName?: string) {
-    return this.client.post<void>("packages/python/scan", { python, saveName });
+  createPythonRequirementsFile(
+    python?: string,
+    saveName?: string,
+    params?: { dir?: string },
+  ) {
+    return this.client.post<void>(
+      "packages/python/scan",
+      { python, saveName },
```
sagerb, line 55: Encode the save name as well here?

sagerb (approved):
Changes look good, although I'm wondering if we need to be encoding the directory names and our previously implemented save name params.

extensions/vscode/src/api/resources/Configurations.ts
dotNomad: Yes we do. Fantastic catch. I've been testing, but didn't test with a double nested one. I will do that now, and update this PR.

extensions/vscode/src/api/resources/Packages.ts
dotNomad: This is in the body so should be good. We just need to encode URI elements.

extensions/vscode/src/api/resources/Configurations.ts
dotNomad: Did some digging into this and testing locally. Axios encodes params passed using its `AxiosRequestConfig` for us. It doesn't encode the the URL which is why we need it for names.

https://github.com/axios/axios/issues/463#issuecomment-250937871

PR #1924, opened 2024-07-02
Use projectDir in config and content record views

dotNomad:
This PR does a few things:
- Adds a description to Configurations and Deployments in their respective views based on the `projectDir`
  - If the `projectDir` is not `.` then the `projectDir` is the description
- Adjusts the delete for both to pass the `projectDir` using the `dir` query param so deletion will work for recursively listed Configurations and Deployments

## Intent

Part of #1845

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/src/views/configurations.ts
+++ b/extensions/vscode/src/views/configurations.ts
@@ -24,7 +24,12 @@ import {
 
 import { confirmDelete, confirmReplace } from "src/dialogs";
 import { getSummaryStringFromError } from "src/utils/errors";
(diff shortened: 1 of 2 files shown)


sagerb (approved):
LGTM. I'm interested in seeing how the description works out for the relative path, and I wonder if the forget prompt should mention the relative path, but let's see how it works out.

PR #1925, opened 2024-07-02
Recursive configuration API

mmarchetti:
## Intent

This PR implements a recursive option to the GET /api/configurations endpoint.

Part of #1850 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Walk the tree looking for .posit directories; when one is found, list configurations from the parent directory. The built-in exclusions are applied to avoid excess directory traversals.

## Automated Tests

Added tests for the recursive flag.

## Directions for Reviewers

```
$(just executable-path) ui -vv --listen=localhost:9001 &
curl -s localhost:9001/api/configurations?recursive=true | jq
$ curl -s 'localhost:9001/api/configurations?recursive=true&dir=test' | jq
$ curl -s 'localhost:9001/api/configurations?recursive=true&dir=test&entrypoint=simple.py' | jq
```

PR #1927, opened 2024-07-03
Add recursive param support to config getAll

dotNomad:
Follow up to #1925 which adds `recursive` param support to the `api.configurations.getAll()` method.


sagerb (approved):
LGTM

PR #1929, opened 2024-07-03
Add recursive parameter to inspection API

mmarchetti:
## Intent

This PR implements a recursive option to the POST /api/inspect endpoint.

Part of https://github.com/posit-dev/publisher/issues/1850

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Walk the tree, inspecting every directory we find. The built-in exclusions are applied to avoid excess directory traversals.

Currently, performance is not great. On my machine, running in this source repo:
* Inspecting the entire repo: 24 seconds to return 40 configurations
* Inspecting test/sample-content: 8 seconds to return 14 configurations
* Inspecting the entire repo with `entrypoint=simple.py`: 8 seconds to return 2 configurations
* Inspecting test/sample-content with `entrypoint=simple.py`: 2 seconds to return 2 configurations

## Automated Tests

There are no tests for this endpoint yet; that needs to change.

## Directions for Reviewers

```
$(just executable-path) ui -vv --listen=localhost:9001 &
curl -s -XPOST -d '{}' 'localhost:9001/api/inspect?recursive=true&dir=test/sample-content' | jq
curl -s -XPOST -d '{}' 'localhost:9001/api/inspect?recursive=true' | jq
curl -s -XPOST -d '{}' 'localhost:9001/api/inspect?recursive=true&dir=test/sample-content&entrypoint=simple.py' | jq
```

PR #1931, opened 2024-07-03
Only request renv::paths() if no renv.lock present

mmarchetti:
## Intent

This PR avoids invoking R to request `renv::paths$lockfile()` if there's an `renv.lock` in the project directory.

Follow-on to #1929 

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [x] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Added a test for this behavior.

## Directions for Reviewers
Inspect an R project that contains renv.lock, and one that doesn't.

If there's an renv.lock file, you shouldn't see a log line like:
```
level=DEBUG msg="Running command" cmd=/usr/local/bin/R args="-s -e renv::paths$lockfile()"
```

### DIFF
--- a/internal/inspect/r.go
+++ b/internal/inspect/r.go
@@ -176,6 +176,15 @@ func (i *defaultRInspector) getRVersion(rExecutable string) (string, error) {
 var renvLockRE = regexp.MustCompile(`^\[1\] "(.*)"`)
 
 func (i *defaultRInspector) getRenvLockfile(rExecutable string) (util.AbsolutePath, error) {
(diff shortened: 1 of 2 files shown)


dotNomad (approved):
Looks great. Pulled it down and tested it with and without a `renv.lock` like you suggested

sagerb (approved):
LGTM and Verified (using the UX)

PR #1932, opened 2024-07-03
Cache interpreter versions

mmarchetti:
## Intent

This PR avoids repeatedly invoking R and Python interpreters to query their versions, by caching those versions.

Draft PR because I need to think about whether we need to not cache the `.pyenv/shims/python` interpreter, which runs different interpreters based on pyenv settings and potentially `.python-version` files. 

Follow-on to #1929 

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 
- - [x] Performance

## Approach

Cache interpreter versions in agent memory, using interpreter path as the key.

## Automated Tests

Existing tests have been updated.

## Directions for Reviewers

Inspect projects (especially recursively). We should only request each R or Python interpreter version once.

### DIFF
--- a/internal/inspect/python.go
+++ b/internal/inspect/python.go
@@ -151,7 +151,12 @@ func (i *defaultPythonInspector) getPythonVersion(pythonExecutable string) (stri
 	}
 	version := strings.TrimSpace(string(output))
 	i.log.Info("Detected Python", "version", version)
-	pythonVersionCache[pythonExecutable] = version
+
+	// Cache interpreter version result, unless it's a pyenv shim
+	// (where the real Python interpreter may vary from run to run)
+	if !strings.Contains(pythonExecutable, "shims") {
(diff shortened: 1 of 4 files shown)


tdstein (approved):
just curious; what happens if an executable is deleted on the system after it has been added to the cache?

mmarchetti:
> just curious; what happens if an executable is deleted on the system after it has been added to the cache?

If it was still the selected interpreter in Positron/VSCode, the extension would succeed at creating configurations using it, but fail if dependency scanning was attempted.

It seems like we should have these cache entries expire, to handle that case as well as the case where a user removes a local virtualenv and recreates it with a different Python version (for example, because their Connect server doesn't have the prior version).

PR #1933, opened 2024-07-03
Add inspect recursive param support to API module

dotNomad:
Follow up to #1929 that adds support for the param to the API module.


sagerb (approved):
LGTM

PR #1938, opened 2024-07-03
Fix empty array response for recursive resources

### LINKED ISSUE #1937: Recursive configurations and deployments GET endpoints return null instead of empty array
If there are no Configurations or Deployments in the directory the `GET /api/configurations` and `GET /api/deployments\` endpoints incorrectly respond with `null` if `recursive=true` is passed.

They correctly respond with `[]` otherwise.

dotNomad:
This fixes the case where there are no Deployments or Configurations and the `recursive` param was passed. In that case the `response` array was never created and was being encoded to `null`.

This fixes that by instantiating the `response` immediately as a variable length slice.

If `recursive` isn't passed it will be overwritten by `readLatestDeploymentFiles`

If `recursive` is passed and nothing is found it will still be an empty `[]`

## Intent

Fixes #1937 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

Test with deployments/configs, without, and recursively

### DIFF
--- a/internal/services/api/get_configurations.go
+++ b/internal/services/api/get_configurations.go
@@ -86,7 +86,7 @@ func GetConfigurationsHandlerFunc(base util.AbsolutePath, log logging.Logger) ht
 		}
 		entrypoint := req.URL.Query().Get("entrypoint")
 
-		var response []configDTO
+		response := make([]configDTO, 0)
 		if req.URL.Query().Get("recursive") == "true" {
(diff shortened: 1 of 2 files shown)


sagerb (approved):
LGTM

PR #1945, opened 2024-07-10
Handle duplicate filenames across subdirectories

### LINKED ISSUE #1944: Support duplicate names within subdirectories
Deployment and configuration filenames only need to be unique within the same `.posit` dir, but when there are multiple of these in a file tree, names are not good enough to identify deployments.

Need to use something which is unique - Perhaps the deploymentPath, which is the actual path on disk.

sagerb:
## Intent

Resolves #1944 

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

deploymentPath within a deployment is unique across all deployment files on the disk. So this was used as a key to find the deployment, with then the projectDir used to select the configuration and deployment status.

## Directions for Reviewers

To exercise this change, create deployments across multiple projects within a tree of projects, and then reload the VSCode window to restore the last deployment selected. Note that the selection saved is per workspace, so opening up projects at different levels will initially not select the last one.

## Checklist

### DIFF
--- a/extensions/vscode/src/constants.ts
+++ b/extensions/vscode/src/constants.ts
@@ -1,11 +1,11 @@
 // Copyright (C) 2024 by Posit Software, PBC.
 
-export const POSIT_FOLDER = ".posit";
-export const PUBLISH_FOLDER = ".posit/publish";
-export const PUBLISH_DEPLOYMENTS_FOLDER = ".posit/publish/deployments";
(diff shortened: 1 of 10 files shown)


sagerb:
Going to issue a different PR, child branch with more changes.

Issue #1949, opened 2024-07-11
Remove optionality of dir parameter for some API calls in Client


sagerb:
Once we have implemented the subdirectory support for the VSCode extension, we should remove the optionality of the `dir` parameter within some of the APIs, so we can get compilation errors for those we have missed.

PR #1953, opened 2024-07-11
Force directory param to be required on API calls and refactor config-based watchers to work in subdirectories

### LINKED ISSUE #1855: Use selected Deployment's `projectDir` for packages
On hold for #1844 and #1851 

The `Python Packages` and `R Packages` sections of the Home view need to utilize the selected Deployment's `parentDir` for the API requests.

- fetching the packages with `GET /api/configurations/$name/packages/python and .../r`
- creating the Python or R requirements file with `POST /api/packages/python/scan and .../r/scan`

---

Part of the #1849 epic.

sagerb:
refactored changes to make dir mandatory for API calls. Updated watchers for package files to work with subdirectories, and updated package views to work in subdirectories. Updated Bus callbacks to be typesafe.

## Intent

Refactored changes to make dir mandatory for API calls. 
Updated watchers for package files to work with subdirectories
Updated package views to work in subdirectories. 
Updated Bus callbacks to be typesafe.

Resolves #1855 
Reolves #1949 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

Use VSCode to open project's subdirectory directly as well as from a few parent subdirectories up.
With these projects, verify the functionality related to packages for both R and Python projects.
Be sure to both manually and automatically update the package file, along with verifying that a non-default package file name within the active config file will be watched and update within the view.

### DIFF
--- a/extensions/vscode/src/api/resources/Configurations.ts
+++ b/extensions/vscode/src/api/resources/Configurations.ts
@@ -20,7 +20,7 @@ export class Configurations {
   // 200 - success
   // 404 - not found
   // 500 - internal server error
-  get(configName: string, params?: { dir?: string }) {
+  get(configName: string, params: { dir: string }) {
     const encodedName = encodeURIComponent(configName);
     return this.client.get<Configuration | ConfigurationError>(
       `/configurations/${encodedName}`,
@@ -31,7 +31,7 @@ export class Configurations {
   // Returns:
   // 200 - success
   // 500 - internal server error
-  getAll(params?: { dir?: string; entrypoint?: string; recursive?: boolean }) {
+  getAll(params: { dir: string; entrypoint?: string; recursive?: boolean }) {
     return this.client.get<Array<Configuration | ConfigurationError>>(
       "/configurations",
       { params },
@@ -45,7 +45,7 @@ export class Configurations {
   createOrUpdate(
     configName: string,
     cfg: ConfigurationDetails,
-    params?: { dir?: string },
+    params: { dir: string },
   ) {
     const encodedName = encodeURIComponent(configName);
     return this.client.put<Configuration>(
@@ -59,7 +59,7 @@ export class Configurations {
   // 204 - success (no response)
   // 404 - not found
   // 500 - internal server error
-  delete(configName: string, params?: { dir?: string }) {
+  delete(configName: string, params: { dir: string }) {
     const encodedName = encodeURIComponent(configName);
     return this.client.delete(`configurations/${encodedName}`, { params });
   }
@@ -70,8 +70,8 @@ export class Configurations {
   // 400 - bad request
   // 500 - internal server error
   inspect(
+    params: { dir: string; entrypoint?: string; recursive?: boolean },
     python?: string,
-    params?: { dir?: string; entrypoint?: string; recursive?: boolean },
   ) {
     return this.client.post<ConfigurationInspectionResult[]>(
       "/inspect",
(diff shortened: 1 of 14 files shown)


extensions/vscode/src/watchers.ts
```diff
@@ -89,28 +90,41 @@ export class ConfigWatcherManager implements Disposable {
...
+    if (root === undefined) {
       return;
     }
-
+    const configurationFileUri = Uri.joinPath(
+      root.uri,
+      contentRecord.projectDir,
+      ".posit",
+      "publisher",
+      contentRecord.configurationName,
```
mmarchetti, line 106: Can you clarify why this changed from `cfg.configurationPath`? Also, does it need a `.toml` suffix?

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -20,7 +20,7 @@ export class Configurations {
   // 200 - success
   // 404 - not found
   // 500 - internal server error
-  get(configName: string, params?: { dir?: string }) {
+  get(configName: string, params: { dir: string }) {
```
dotNomad, line 23: I understand this is to avoid us from forgetting the `dir` param when we should include it, but I do think we should discuss the value we are losing here. With this change the `api` module diverges from representing the API.

Personally I would prefer the module be a 1-1 with the actual API functionality. Mainly because I believe it causes confusion for anyone who isn't intimately familiar with the patterns of this code base why a URL param (which is normally optional) is not optional here.

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -70,8 +70,8 @@ export class Configurations {
   // 400 - bad request
   // 500 - internal server error
   inspect(
+    params: { dir: string; entrypoint?: string; recursive?: boolean },
     python?: string,
-    params?: { dir?: string; entrypoint?: string; recursive?: boolean },
```
dotNomad, line 74: I found this switch very confusing. The Axios functions, and all of our others do `data` or parameters to fill the data/URL, and then params. Seeing the function change elsewhere really got me scratching my head before I came back here to discover the movement.

extensions/vscode/src/watchers.ts
dotNomad: Yeah I'm a bit confused by this as well. We have the `configurationPath` why not just use that as before with `new RelativePattern` and without making our own path?

dotNomad (changes requested):
Marking this one to have the discussion around the API module pattern change.

extensions/vscode/src/api/resources/Configurations.ts
sagerb: Let's discuss.

extensions/vscode/src/api/resources/Configurations.ts
sagerb: I did not like it either, but it comes at the language requirement to maintain the order of required then optional parameters. I'm good with refactoring this to avoid this awkwardness, but I do want to maintain the typescript parameter checks as they were invaluable to the refactoring work I was doing and I believe to what we have yet to do in the future.

extensions/vscode/src/watchers.ts
sagerb: I'll check, but I think this was probably just an oversight. I'll mark this and the other as being switched back and will test it from there. I'm sure the both of you are probably right.

Issue #1950, opened 2024-07-11
API Call `/configurations/${encodedName}/files` response should suppress projectDir from id / relative path


sagerb:
The UX is having to remove the projectDir from the `id` field of the response, when a `dir` parameter is passed in. This is occurring within the `src/views/homeView.ts` file.

For example:

if no `dir` is passed in or `dir === "."`, then the files returned look correct (as we need them).
```
fileType = 'REGULAR'
id = 'simple.py'
isDir = false
isEntrypoint = false
isFile = true
modifiedDatetime = '2024-05-20T08:44:04-07:00'
reason = {source: 'file', pattern: '!simple.py', exclude: true, fileName: 'configuration-1.toml', filePath: '/Users/billsager/dev/publishing-client-alt/t…-simple/.posit/publish/configuration-1.toml'}
rel = 'simple.py'
relDir = '.'
size = 369
```

however, if we do pass in a `dir`, such as `sample-content/fastapi-simple`

```
abs = '/Users/billsager/dev/publishing-client-alt/test/sample-content/fastapi-simple/simple.py'
base = 'simple.py'
files = (0) []
fileType = 'REGULAR'
id = 'sample-content/fastapi-simple/simple.py'
isDir = false
isEntrypoint = false
isFile = true
modifiedDatetime = '2024-05-20T08:44:04-07:00'
reason = {source: 'file', pattern: '!simple.py', exclude: true, fileName: 'configuration-1.toml', filePath: '/Users/billsager/dev/publishing-client-alt/t…-simple/.posit/publish/configuration-1.toml'}
rel = 'sample-content/fastapi-simple/simple.py'
relDir = 'sample-content/fastapi-simple'
size = 369
```

We can't simply use the base, because if we have a subdirectory included in the files, it comes out like this, and in order to correctly insert it into the requests to update the config file inclusion line, it has to be relative to the projectDir.

```
{
  "id": "sample-content/fastapi-simple/test",
  "fileType": "DIR",
  "base": "test",
  "reason": {
    "source": "file",
    "pattern": "*",
    "exclude": false,
    "fileName": "configuration-1.toml",
    "filePath": "/Users/billsager/dev/publishing-client-alt/test/sample-content/fastapi-simple/.posit/publish/configuration-1.toml"
  },
  "files": [
    {
      "id": "sample-content/fastapi-simple/test/abc.txt",
      "fileType": "REGULAR",
      "base": "abc.txt",
      "reason": {
        "source": "file",
[...]

PR #1954, opened 2024-07-12
Enable files view to work when project is located in subdirectory

### LINKED ISSUE #1854: Use selected Deployment's `projectDir` to list files and include/exclude
On hold for #1844 and #1851 

The `Project Files` section in the Home View will need to utilize the `parentDir` of the selected Deployment for all of the calls to `GET and POST /api/configurations/$name/files` to correctly list, include, and exclude files.

Blocked by #1950 - Bug

---

Part of the #1849 epic

sagerb:
## Intent

Closes #1854 (Use selected Deployment's `projectDir` to list files and include/exclude)

## Type of Change

- - [x] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

After everything was updated to pass in the projectDir from the activeContentRecord, adding and removing files from the config.Files section would not work correctly when the project was not located at the top root dir that VSCode had opened.

This was because the files API returned an ID that included the projectDir, and this was an incorrect value for the config file. I've created issue #1950, but this PR implements a simple workaround that removes the projectDir from the path represented in the id of a file.

## Directions for Reviewers

Make sure you can add and remove exclusions from the files view, with the selected configuration being updated, and the files view updating to show the result.
Also make sure you can update the config file directly, and the files view is updated correctly.

Need to verify this for both projects that are opened directly as well as projects that exist below the main folder opened within VSCode.

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -1147,6 +1147,17 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           },
         );
 
+        // Need to remove the projectDir from the id, since the files id
+        // which will be used to include or exclude is relative to the
+        // projectDir. We really should have the backend agent do this
+        // for us, but for now, it works.
+        response.data.files.forEach((contentRecordFile) => {
+          contentRecordFile.id = contentRecordFile.id.replace(
+            activeDeployment.projectDir + "/",
+            "",
+          );
+        });
+
         this.webviewConduit.sendMsg({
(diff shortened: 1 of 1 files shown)


dotNomad (approved):
Definitely agree that we should use the backend for this. I'm wondering exactly how this will work or if the API will just need to adjust to take non-relative paths. Approving before figuring this out to keep the PR chain moving - especially since you already have a follow-up task.

mmarchetti (approved):

PR #1955, opened 2024-07-12
Show project dir when selecting entrypoints

sagerb:
## Type of Change

Show the projectDirs in the multit-step selectors for entrypoints

Part of #1845 

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

2024-07-12 at 10 18 AM

## Directions for Reviewers

The three multi-steppers were updated in the same way, adding the projectDir to their detail attribute.

You can invoke them by:
- New Config from config view
- New Deployment when there are multiple entrypoints
- Select Config, when there are multiple entrypoints with the same content type as the current deployment.

You should verify that the projectDir is shown in those multiSteps.

## Checklist


dotNomad (approved):
Lovely. So simple 🎉

mmarchetti (approved):

PR #1956, opened 2024-07-12
Updates to configurations view to work with subdirectories

### LINKED ISSUE #1951: Update Configurations View to support subdirectories
Including operations of 
- new
- refresh
- clone
- delete
- rename

sagerb:
## Intent

Resolves #1951 

- move progress indicator to configurations view
- update clone initial name for projectDir
- add progress indicator for children tree population (relies upon sometimes slow API)

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

- Verify operations work for configurations view
- Verify that when creating a new configuration when VSCode has opened a folder containing a lot of entrypoints, that the progress indicator is shown within the Configurations view (vs. down below in the bottom status line).
- Verify that when opening very large directory tree and switching over to publisher, that configurations window when expanded shows progress indicator until it is able to load.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/configurations.ts
+++ b/extensions/vscode/src/views/configurations.ts
@@ -75,10 +75,22 @@ export class ConfigurationsTreeDataProvider
 
     try {
       const api = await useApi();
-      const response = await api.configurations.getAll({
+      const getAllPromise = api.configurations.getAll({
         dir: ".",
         recursive: true,
       });
+
+      window.withProgress(
+        {
+          title: "Initializing",
+          location: { viewId: Views.Configurations },
+        },
+        async () => {
+          return getAllPromise;
+        },
+      );
+
+      const response = await getAllPromise;
       const configurations = response.data;
 
       return configurations.map((config) => {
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/views/configurations.ts
```diff
@@ -75,10 +75,22 @@ export class ConfigurationsTreeDataProvider
...
       });
+
+      window.withProgress(
+        {
+          title: "Initializing",
+          location: { viewId: Views.Configurations },
+        },
+        async () => {
+          return getAllPromise;
+        },
```
dotNomad, line 90: Can we just pass the `getAllPromise` here rather than wrapping it in another function?

dotNomad (approved):
Fantastic. Just had one syntax question

extensions/vscode/src/views/configurations.ts
sagerb: API is expecting a function returned.

PR #1959, opened 2024-07-12
Add progress indicator for homeview APIs

### LINKED ISSUE #1958: Add HomeView progress indicator while operations known to cause refresh and further processing are outstanding
I have noticed when opening a very large subdirectory, that the UX shows itself out-of-date after an operation that should update the display.

For example, opening up my `dev` folder with 44,909,151,820 bytes (50.36 GB on disk), and:
- activating the publisher, there is a significant delay before the home view is updated with the latest info (23 seconds)
- removing a file from being included in the deployment can take 17 seconds to show an updated UX.

sagerb:
## Intent

Resolves #1958 

Adds a HomeView progress indicator while API calls are outstanding

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Wrapped API calls with VSCode's `withProgress` call, showing on the HomeView.

## Directions for Reviewers

Verify you see a progress indicator within HomeView when it is initializing, after switching to publisher view. This will be very visible if you open a very large folder in VSCode as the base dir.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -306,7 +306,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     }
     try {
       const api = await useApi();
-      await api.files.updateFileList(
+      const apiRequest = api.files.updateFileList(
         activeConfig.configurationName,
         uri,
         action,
@@ -314,6 +314,16 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           dir: activeDeployment.projectDir,
         },
       );
+      window.withProgress(
+        {
+          title: "Updating File List",
+          location: { viewId: Views.HomeView },
+        },
+        async () => {
+          return apiRequest;
+        },
+      );
+      await apiRequest;
     } catch (error: unknown) {
       const summary = getSummaryStringFromError(
         "homeView::updateFileList",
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -306,14 +306,24 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
         },
       );
+      window.withProgress(
+        {
+          title: "Updating File List",
+          location: { viewId: Views.HomeView },
+        },
+        async () => {
+          return apiRequest;
+        },
```
dotNomad, line 324: Okay its 2 PR's late, but I did try unwrapping these and since it isn't a function it fails to typing 😞

extensions/vscode/src/views/homeView.ts
```diff
@@ -1140,12 +1240,22 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
         );
+        window.withProgress(
+          {
+            title: "ReFreshing Files",
+            location: { viewId: Views.HomeView },
+          },
+          async () => {
+            return apiRequest;
+          },
+        );
```
dotNomad, line 1257: I wonder if each of our views should have a helper to show progress like the below:

```
  private showProgress(promise: Promise<unknown>, title: string) {
    return window.withProgress(
      {
        title: title,
        location: { viewId: Views.HomeView },
      },
      async () => promise,
    );
  }
```

Could make it a bit less verbose each time.

dotNomad (approved):
Looks great, just had a helper function suggestion.

extensions/vscode/src/views/homeView.ts
sagerb: :-)

PR #1961, opened 2024-07-12
Disable homeview's Easy Deploy View until ready

### LINKED ISSUE #1960: Disable HomeView when intializing
Buttons are enabled, but user should not be able to perform a deployment or operation.

Should we hide the view and instead show a larger loading indicator?

sagerb:
## Intent

Resolves #1960 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Added a new message which is returned after the extension believes the initialization queries have been completed.
Then I added a variable within the home store that indicates if initialization is complete or not.
Message handlers update the value of the variable within the home store, and then the EvenEasierDeploy vue reacts to the value of that ref.

End Result:

https://github.com/user-attachments/assets/550e19da-5b53-4f6f-9fda-f5f6a922908b

## Directions for Reviewers

Open a very large folder with VSCode, activate the publisher and verify the proper views are visible after initialization.
Do the same with a small folder.

## Checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,6 +2,11 @@ vendor
 
 # VSCode Extension
 extensions/vscode/out
+test/sample-content
 
 # webdriver.io service
 test/vscode-ui/.wdio-vscode-service
+extensions/vscode/dist
+extensions/vscode/node_modules
+extensions/vscode/webviews/homeView/node_modules
+extensions/vscode/webviews/homeView/dist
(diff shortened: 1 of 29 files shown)


dotNomad (approved):
Looking good 😎

PR #1962, opened 2024-07-12
Instrument vscode progress and api calls

sagerb:
## Intent

We need visibility into performance issues when the extension is running within VSCode that has opened a large folder full of files and other folders.

This PR instruments the API calls and the progress windows to expose how long some of these operations are displaying to the users, so that we can analyze and optimize the underlying code.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

I have organized my changes within individual commits, listed below.

- disable noisy bus and event traces
- Implement duration tracing for axios calls
- implement showProgress utility, with duration tracing
- use new showProgress utility rather than window.withProgress

I plan to come back and allow all of this tracing to be configurable through extension settings so that we can turn it all off by default and then back on when needed.

## Directions for Reviewers

I'd recommend you review via the commits. They will make much more sense.

## Checklist

### DIFF
--- a/extensions/vscode/src/api/client.ts
+++ b/extensions/vscode/src/api/client.ts
@@ -22,6 +22,19 @@ class PublishingClientApi {
     this.client = axios.create({
       baseURL: apiBaseUrl,
     });
+    this.client.interceptors.request.use((request) => {
+      (<any>request).ts = performance.now();
+      return request;
+    });
+
+    this.client.interceptors.response.use((response) => {
+      const request = response.request;
+      const duration = Math.round(
+        Number(performance.now() - (<any>response).config.ts),
+      );
+      console.log(`Request: ${request.path} took ${duration}ms`);
+      return response;
+    });
     this.apiServiceIsUp = apiServiceIsUp;
 
     this.configurations = new Configurations(this.client);
--- a/extensions/vscode/src/utils/progress.ts
+++ b/extensions/vscode/src/utils/progress.ts
@@ -1,0 +1,26 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+import { ProgressLocation, window } from "vscode";
+
+export async function showProgress(
+  title: string,
+  viewId: string | undefined,
+  until: Promise<any>,
+  trace = true,
+) {
+  const start = performance.now();
+  window.withProgress(
+    {
+      title,
+      location: viewId ? { viewId } : ProgressLocation.Window,
+    },
+    async () => {
+      return until;
+    },
+  );
+  await until;
+  if (trace) {
+    const duration = Math.round(Number(performance.now() - start));
+    console.log(`Progress for "${title}" was displayed for ${duration}ms`);
+  }
+}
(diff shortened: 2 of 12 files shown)


extensions/vscode/src/api/client.ts
```diff
@@ -22,6 +22,19 @@ class PublishingClientApi {
     this.client = axios.create({
       baseURL: apiBaseUrl,
     });
+    this.client.interceptors.request.use((request) => {
+      (<any>request).ts = performance.now();
```
dotNomad, line 26: Rather than making these `any` we can adjust the `axios.d.ts` module type to include `ts`:

```
declare module "axios" {
  interface AxiosRequestConfig {
    ignoreCamelCase?: string[];
    ts?: DOMHighResTimeStamp;
  }
}
```

Note that I made it optional here otherwise we have to include a default in `axios.create`. We can supply that default to remove the optionality of `ts`.

This would allow us to avoid the `(<any>request)` typing and just use the new attribute directly.

extensions/vscode/src/utils/progress.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+import { ProgressLocation, window } from "vscode";
+
+export async function showProgress(
+  title: string,
+  viewId: string | undefined,
```
dotNomad, line 7: To follow JS norms I'd move the optional `viewId` after `until`

dotNomad (commented):
Looks like all of the other console debugs are commented out here. I think that is smart. We can uncomment them locally until we have a nice debug mode or setting that we can optionally enable that debugging code.

---

We should address the typing here. I'm fine doing that in a follow-up PR to avoid needed to do rebasing. Leaving as a comment to get us to be proactive about that change though.

extensions/vscode/src/utils/progress.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+import { ProgressLocation, window } from "vscode";
+
+export async function showProgress(
```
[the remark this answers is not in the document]
sagerb: I was psychic!

extensions/vscode/src/utils/progress.ts
```diff
@@ -0,0 +1,26 @@
...
+  title: string,
+  viewId: string | undefined,
+  until: Promise<any>,
+  trace = true,
```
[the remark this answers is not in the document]
sagerb: totally agree, this was just a quick start. Needed to understand the performance issues I was seeing, so I could provide a heads up on them for Monday, when y'all returned.

extensions/vscode/src/api/client.ts
sagerb: beautiful!

sagerb:
>We should address the typing here. I'm fine doing that in a follow-up PR to avoid needed to do rebasing. Leaving as a comment to get us to be proactive about that change though.

That sounds great! @dotNomad Are you ok with approving this PR then?

dotNomad (approved):
> That sounds great! @dotNomad Are you ok with approving this PR then?

Yep! Sorry about that meant to approve

PR #1978, opened 2024-07-17
Add timestamp to AxiosRequestConfig type

dotNomad:
PR feedback follow-up. Adds the `ts` attribute as an optional attribute for the `AxiosRequestConfig` so we get type safety.

### DIFF
--- a/extensions/vscode/webviews/homeView/src/axios.d.ts
+++ b/extensions/vscode/webviews/homeView/src/axios.d.ts
@@ -1,0 +1,11 @@
+// Copyright (C) 2023 by Posit Software, PBC.
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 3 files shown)


sagerb (approved):
LGTM

dotNomad:
Going to come back to this in a bit to solve that CI failure.

extensions/vscode/webviews/homeView/src/axios.d.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2023 by Posit Software, PBC.
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { AxiosRequestConfig } from "axios";
+
+declare module "axios" {
```
dotNomad, line 6: I needed the second file so the homeView TypeScript compilation could grab the `.d.ts` file.

PR #1979, opened 2024-07-17
Make showProgress viewId optional

dotNomad:
PR feedback follow-up. Moves the `viewId` param for `showProgress` to make it optional.


sagerb (approved):
I have mixed feelings about this change. I liked the fact that you had your think about where the progress indicator was best shown, vs doing it as an afterthought (which is how I often think of optional parameters). 

We did have it optional in prior art, so I guess I'm ok with this change. The existing calls to this helper function all provide the viewId, so the impact here is really only on new code.

I'll approve it, but when you have a chance, let's discuss a bit.

Thanks!

dotNomad:
> I have mixed feelings about this change. I liked the fact that you had your think about where the progress indicator was best shown, vs doing it as an afterthought (which is how I often think of optional parameters).

It was technically already option since it was typed as `viewId: string | undefined`. I'm totally fine requiring a `viewId` in the function calls, and think that would provide a nice benefit.

sagerb:
> It was technically already option since it was typed as `viewId: string | undefined`. I'm totally fine requiring a `viewId` in the function calls, and think that would provide a nice benefit.

Glad to hear it. Yes, let's require the `viewId` in the function calls then.

sagerb (changes requested):
We agreed to make the change, so let's make the change within this PR.

I can make these changes for you, since I know you're busy with user testing.

sagerb:
I've pushed a commit which removes the optionality of the viewID within Show Progress. https://github.com/posit-dev/publisher/pull/1979/commits/08ef450219aade47983a3964f60905f086e3af30

@dotNomad Perhaps you could review those changes?

sagerb (approved):
@dotNomad code looks good.. Would like to have him review the couple of commits which I contributed before it is merged.

extensions/vscode/src/views/credentials.ts
```diff
@@ -97,7 +97,10 @@ export class CredentialsTreeDataProvider
    * @returns
    */
   public add = async (startingServerUrl?: string) => {
-    const credential = await newCredential(startingServerUrl);
+    const credential = await newCredential(
+      Views.Configurations,
```
dotNomad, line 101: Should this be `Views.Credentials`?

dotNomad (commented):
I think I caught one incorrect reference.

extensions/vscode/src/views/credentials.ts
sagerb: Yes. I've pushed an update.

PR #1985, opened 2024-07-17
Improve deployment quick pick detail string

dotNomad:
This PR improves the strings for the Deployment quick picker by:
- Only showing the path if it isn't `.`
- Always showing the path even if there is a credential error
- Using the `•` separator so it is a bit easier to read

<details>
  <summary>Preview</summary>
  
CleanShot 2024-07-17 at 13 51 44@2x
CleanShot 2024-07-17 at 13 52 47@2x

</details> 

## Intent

Part of #1854


sagerb (approved):
LGTM

Issue #1984, opened 2024-07-17
Configuration no longer updates webview on change


dotNomad:
Actions like:
- deleting the selected deployment's configuration
- updating the selected deployment's configuration title doesn't propagate to the webview

dotNomad:
One idea of why we are seeing the refresh files error on config file deletion is the watchers aren't getting properly disposed of with the logic moved to return early if there isn't an active configuration.

dotNomad:
This can lead to `files.getByConfiguration` `404` errors:

![Image](https://github.com/user-attachments/assets/2af9cc0f-1906-4d5d-9eaa-efadb7497ff9)

PR #1986, opened 2024-07-17
Use `activeConfiguration.projectDir` where appropriate

dotNomad:
This PR reverts some of the refactoring done in the nested project work and utilizes `activeConfiguration.projectDir` directly rather than using only `activeContentRecord.projectDir` since all of these operations rely on the configuration. This makes it a bit easier to understand what some of those functions are relying on, and a bit simpler.

This also makes the `activeConfigChanged` listener function use optionality again rather than terminating early to avoid errors caused by the active content record becoming `undefined` but the file watchers were still firing.

## Intent

Part of [#1984](https://github.com/posit-dev/publisher/issues/1984)

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -120,49 +120,43 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     });
 
     useBus().on("activeConfigChanged", (cfg: Configuration | undefined) => {
-      const activeContentRecord = this.getActiveContentRecord();
-      if (!activeContentRecord) {
-        return;
-      }
-
       this.sendRefreshedFilesLists();
       this.onRefreshPythonPackages();
       this.onRefreshRPackages();
 
       this.configWatchers?.dispose();
-      if (cfg) {
-        this.configWatchers = new ConfigWatcherManager(cfg);
-        this.configWatchers.configFile?.onDidChange(
-          this.sendRefreshedFilesLists,
-          this,
-        );
+      this.configWatchers = new ConfigWatcherManager(cfg);
 
-        this.configWatchers.pythonPackageFile?.onDidCreate(
-          this.onRefreshPythonPackages,
-          this,
-        );
-        this.configWatchers.pythonPackageFile?.onDidChange(
-          this.onRefreshPythonPackages,
-          this,
-        );
-        this.configWatchers.pythonPackageFile?.onDidDelete(
-          this.onRefreshPythonPackages,
-          this,
-        );
+      this.configWatchers.configFile?.onDidChange(
+        this.sendRefreshedFilesLists,
+        this,
+      );
 
-        this.configWatchers.rPackageFile?.onDidCreate(
-          this.onRefreshRPackages,
-          this,
-        );
-        this.configWatchers.rPackageFile?.onDidChange(
-          this.onRefreshRPackages,
-          this,
-        );
-        this.configWatchers.rPackageFile?.onDidDelete(
-          this.onRefreshRPackages,
-          this,
-        );
-      }
+      this.configWatchers.pythonPackageFile?.onDidCreate(
+        this.onRefreshPythonPackages,
+        this,
+      );
+      this.configWatchers.pythonPackageFile?.onDidChange(
+        this.onRefreshPythonPackages,
+        this,
+      );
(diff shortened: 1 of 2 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -120,49 +120,43 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     });
 
     useBus().on("activeConfigChanged", (cfg: Configuration | undefined) => {
-      const activeContentRecord = this.getActiveContentRecord();
-      if (!activeContentRecord) {
-        return;
-      }
```
dotNomad, line 126: Instead of failing early we can make everything below this optional like it was previously. Using `undefined` and adapting to it in the sending, refreshing, and config watcher setup.

extensions/vscode/src/views/homeView.ts
```diff
@@ -286,12 +280,6 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
   }
 
   private async updateFileList(uri: string, action: FileAction) {
-    const activeDeployment = this.getActiveContentRecord();
-    if (!activeDeployment) {
-      console.error("homeView::updateFileList: No active deployment.");
-      return;
-    }
-    // this will only be true if the config really exists
```
dotNomad, line 294: A lot of cases like this will never occur because of the active config being so tied to the active deployment, but we don't need to rely on the early termination since these calls all still work if we have a mistake where we have an active config but no active deployment.

sagerb (approved):
I'm going to approve this because I understand what you are doing with the changes, but it would be helpful if you could outline the sequence in which my earlier logic fails? Granted my logic reached back into the selected contentRecord, but I am thought that had an advantage of making sure we're not acting on a change to a config file which is not active.

I'm sure you can explain it pretty quickly. :-)

dotNomad:
> I'm going to approve this because I understand what you are doing with the changes, but it would be helpful if you could outline the sequence in which my earlier logic fails? Granted my logic reached back into the selected contentRecord, but I am thought that had an advantage of making sure we're not acting on a change to a config file which is not active.
> 
> I'm sure you can explain it pretty quickly. :-)

This is in two parts. I attempted to explain with my comments above, but I can try to reword a bit.

A handful of methods here relied on `getActiveContentRecord` to fetch the `projectDir` and `configurationName` to make API calls. Instead we should be using the `getActiveConfig` result to avoid making API calls that we know are going to fail. This can occur if the Content Record is referencing a `configurationName` that doesn't exist for example. Which would make refreshing the files lists, and refreshing R/Python packages fail.

This could be seen if you delete a Configuration used in the selected Content Record. That error is screenshotted in the issue this links to.

Some of the methods make sure that configuration exists, but that makes the `getActiveContentRecord` calling and attribute referencing redundant so it is removed here.

The next part is the `activeConfigChanged` listener. It checks if there is an active content record and returns early if there isn't one. Logically this makes sense since we don't set a configuration without a deployment, but in the code it doesn't dispose of the previous `configWatchers` which could lead to a situation where the previous watchers are still setup and firing.

To solve that I made the entire listener handle the `undefined` case without failing early. It is a chunk of logic but thanks to `?` optional chaining on the individual config watchers, and the refresh methods handling an `undefined` config as well all of the code gets skipped without missing a `dispose`.

The `dispose()` could have been moved up to always occur, and still had the earlier failure, but the watchers already need to be optionally chained since we cannot guarantee a `root` for setting up file watching, and each of the refresh methods already need to handle `undefined` configurations since `getActiveConfig()` can return `undefined`.

dotNomad:
I think the zoomed-out explanation is rather than bouncing out early at high level function, we can instead have low level functions handle `undefined` themselves which prevents us from needing to:
- re-arrange logic (like `dispose()`) before those checks
- forgetting where those checks happen to prevent issues like `sendRefreshedFilesLists` having a potentially `undefined` config

The change to bail out of `activeConfigChanged` could have worked if `dispose()` was moved up, and we checked if the configuration existed rather than using `activeDeployment.configurationName`. But the philosophy shift is relying on those low level functions to handle the state themselves.

PR #1988, opened 2024-07-18
Avoid showing relative root dir in inputs

dotNomad:
This PR refines the `multiInputs` a bit to avoid showing `./` as the path if the `projectDir` `isRelativePathRoot`.

The path will continue to show if the `projectDir !== '.'`


sagerb (approved):
LGTM

PR #1992, opened 2024-07-18
Avoid race condition when creating new deployments

### LINKED ISSUE #1989: Creating a new Deployment no longer selects it automatically
This is a race condition caused by: https://github.com/posit-dev/publisher/blob/994a6c6663344ab0b198d055a1ee5c04b2831395/extensions/vscode/src/views/homeView.ts#L1147

Previously the `refreshContentRecords` didn't pass the selected record, now it does which causes this problem.

dotNomad:
This PR solves a race condition and other issues found around creating new deployments.

The previous edge-case capturing logic was re-introduced that handled optionally sending the selection to the webview. It needs to be optional to avoid this race condition where the new deployment async code fires before the file watcher listeners. It also now handles `undefined` and `null` again so we can specify if we want to force de-selection or just update data.

Some equality checking was also fixed which was causing some of the errors seen.

## Intent

Resolves #1989

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

- Ensure last selection loads correctly on Extension init
- Ensure that a deleted deployment causes the selection to be unselected if the selection was deleted
- Ensure that when a selected deployment changes that it is updated in the sidebar
- Ensure that when a deployment is created it is selected (regardless of what the previous selection was)

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -1146,7 +1146,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
 
   public refreshContentRecords = async () => {
     await this.refreshContentRecordData();
-    this.updateWebViewViewContentRecords(this.getSelectionState());
+    this.updateWebViewViewContentRecords();
     useBus().trigger(
       "activeContentRecordChanged",
       this.getActiveContentRecord(),
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -839,7 +839,10 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       if (
         !this.contentRecords.find(
           (contentRecord) =>
-            contentRecord.saveName === deploymentObjects.contentRecord.saveName,
+            contentRecord.saveName ===
+              deploymentObjects.contentRecord.saveName &&
+            contentRecord.projectDir ===
+              deploymentObjects.contentRecord.projectDir,
         )
       ) {
         this.contentRecords.push(deploymentObjects.contentRecord);
@@ -849,8 +852,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           (config) =>
             config.configurationName ===
               deploymentObjects.configuration.configurationName &&
-            config.configurationPath ===
-              deploymentObjects.contentRecord.projectDir,
+            config.projectDir === deploymentObjects.configuration.projectDir,
         )
       ) {
         this.configs.push(deploymentObjects.configuration);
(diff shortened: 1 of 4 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -849,8 +852,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           (config) =>
             config.configurationName ===
               deploymentObjects.configuration.configurationName &&
-            config.configurationPath ===
-              deploymentObjects.contentRecord.projectDir,
+            config.projectDir === deploymentObjects.configuration.projectDir,
```
dotNomad, line 855: This appeared to be incorrect and caused issues when debugging where the configuration was never found, stayed `undefined` when doing `getActiveConfig()` and this solved that.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1144,7 +1146,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
 
   public refreshContentRecords = async () => {
     await this.refreshContentRecordData();
-    this.updateWebViewViewContentRecords(this.getSelectionState());
+    this.updateWebViewViewContentRecords();
```
dotNomad, line 1149: We don't want to send the selection here, instead we only want to update the data.

extensions/vscode/webviews/homeView/src/HostConduitService.ts
```diff
@@ -95,11 +95,25 @@ const onRefreshContentRecordDataMsg = (msg: RefreshContentRecordDataMsg) => {
...
+    if (home.selectedContentRecord) {
+      home.updateSelectedContentRecordBySelector({
+        deploymentPath: home.selectedContentRecord.deploymentPath,
+      });
+    }
+    return;
+  }
+
+  // At this point we have a selector, so update the selection
+  home.updateSelectedContentRecordBySelector(selector);
```
dotNomad, line 116: This logic was around previously to make
- `null` clear the selection
- `undefined` update data
- `selector` to choose the new `selectedContentRecord`

sagerb (approved):
Code looks good and behavior is confirmed.

extensions/vscode/src/views/homeView.ts
dotNomad: `refreshContentRecords` happens based on file watchers. We don't want to send over the selection with this we just want the content records' data to be updated.

Part of the issue here was that the create was happening, and in between the logic where we capture the deployment created the file was made by the backend which triggered the file watchers.

This led to the Deployment getting selected at the end of the create logic, and at the same time the new data being sent to the webview with the old `this.getSelectionState()`.

If we limit the watchers to only updating data we avoid race conditions with other logic manipulating files.

PR #1994, opened 2024-07-18
Handle JSON errors in server validation

### LINKED ISSUE #1974: Cannot Add Connect behind Proxy
We got this from slack:

https://positpbc.slack.com/archives/C05K993PVEV/p1721156669455599

It appears when Connect is behind a proxy we cannot add the server.  

When you add the `palm` url (https://pub.palm.ptd.posit.it/) the following error is thrown:

Screenshot 2024-07-16 at 3 34 34 PM

Based on the logs, it looks like we're having trouble parsing the response:

> time=2024-07-16T15:11:34.781-04:00 level=INFO msg="Testing authentication" method="Connect API key" url=https://pub.palm.ptd.posit.it/connect/
> time=2024-07-16T15:11:35.654-04:00 level=DEBUG msg="API request" method=GET path=/__api__/v1/user body="" response="<!DOCTYPE html>\n\n\n\n\n<html lang=\"en\">\n<head>\n    <meta charset=\"UTF-8\">\n\n    <script type=\"text/javascript\" nonce=\"32Zh1d1fB6Icnylb6KyWqA\">if (typeof module === 'object') {window.module = module; module = undefined;}</script><style type=\"text/css\" nonce=\"32Zh1d1fB6Icnylb6KyWqA\">\n 
> ...
> 

When we _can_ successfully add a Server we hit (dogfood for reference):

> time=2024-07-16T15:18:37.180-04:00 level=INFO msg="Testing authentication" method="Connect API key" url=https://rsc.radixu.com
> time=2024-07-16T15:18:39.521-04:00 level=DEBUG msg="API request" method=GET path=/__api__/v1/user body="" response="" error="unexpected response from the server (401)"
> time=2024-07-16T15:18:39.521-04:00 level=INFO msg="Connect server connectivity check failed" url=https://rsc.radixu.com
> time=2024-07-16T15:18:39.521-04:00 level=INFO msg="Access Log" method=POST url=/api/test-credentials elapsed_ms=2340 status=200 req_size=32 resp_size=27 client_addr=127.0.0.1:64776 req.url=https://rsc.radixu.com

I think we expect a 401 response and we're surprised to hit the proxy.

This URL works with both RStudio IDE (`rsconnect`) and `rsconnect-python`

From @colearendt 

> We let certain things through (from history): X-Auth-Token header is present (for IDE access), or Authorization: Key * header for API requests.
> https://github.com/rstudio/ptd/blob/e72d358f58120144db25cf3cf8aa50690a8bd50a/ptd/src/ptd/pulumi_resources/traefik_forward_auth.py#L153-L154
> Relevant Connect docs:
> https://docs.posit.co/connect/admin/authentication/proxied/#ide-deployment

mmarchetti:
## Intent

An authenticating proxy such as the one in front of [palm](https://pub.palm.ptd.posit.it/) may return a login page redirect for an unauthenticated API request. We make such a request when testing a server URL without an API key (during the Add Credential flow in the extension). This PR changes loosens the URL-only validation to consider HTML responses to be successful. The next validation step includes an API key, and remains strict.

Fixes #1974

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Added a unit test for this case.

## Directions for Reviewers

Try various server URLs:
* https://rsc.radixu.com
* https://pub.palm.ptd.posit.it/
* ~https://pub.palm.ptd.posit.it/connect/~
* https://google.com

### DIFF
--- a/extensions/vscode/src/utils/url.ts
+++ b/extensions/vscode/src/utils/url.ts
@@ -5,6 +5,10 @@ export const formatURL = (input: string): string => {
   if (/^[a-zA-Z]+:\/\//.test(input)) {
     return input;
   }
+  if (input.endsWith("/connect/")) {
+    // This is a dashboard URL; trim to get the server URL.
+    input = input.slice(0, -8);
+  }
   return `https://${input}`;
 };
 
--- a/extensions/vscode/src/utils/url.ts
+++ b/extensions/vscode/src/utils/url.ts
@@ -5,10 +5,6 @@ export const formatURL = (input: string): string => {
   if (/^[a-zA-Z]+:\/\//.test(input)) {
     return input;
   }
-  if (input.endsWith("/connect/")) {
-    // This is a dashboard URL; trim to get the server URL.
-    input = input.slice(0, -8);
-  }
   return `https://${input}`;
 };
 
--- a/internal/clients/connect/client_connect.go
+++ b/internal/clients/connect/client_connect.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"net/url"
 	"regexp"
 	"strings"
 	"time"
@@ -92,7 +91,13 @@ var errInvalidServer = errors.New("could not access the server; check the server
 func isConnectAuthError(err error) bool {
 	// You might expect a 401 for a bad API key (and we'll handle that).
 	httpErr, ok := err.(*http_client.HTTPError)
-	return ok && (httpErr.Status == http.StatusUnauthorized)
+	if ok && (httpErr.Status == http.StatusUnauthorized) {
+		return true
+	}
+	// An authenticating proxy may return a redirect to an auth page,
+	// which will return a 200 with an HTML body.
+	var serr *json.SyntaxError
+	return errors.As(err, &serr)
 }
 
 func (c *ConnectClient) TestAuthentication(log logging.Logger) (*User, error) {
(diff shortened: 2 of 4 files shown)


kgartland-rstudio:
If anyone else was curious, using a customer Landing page does not hit this (I thought it would).

kgartland-rstudio:
I can now use `https://rsc.radixu.com/connect` to get passed the URL validation but it fails the API Key validation (Invalid API Key).

Additionally it allowed me to get passed the URL validation when I already had a set of creds for `https://rsc.radixu.com`. I imagine we still want to fail the URL validation if creds using that address were already created.

kgartland-rstudio:
The RStudio IDE doesn't allow you to append `/connect` to your URL either.  Maybe instead of supporting this case, we can just throw a helpful message if we can validate the URL sans `/connect` that says they have to drop the `/connect`.  

I wonder if we have to be careful here as a customer could have Connect hosted on path ending in `/connect/`.

mmarchetti:
> I wonder if we have to be careful here as a customer could have Connect hosted on path ending in `/connect/`.

This is a good point, maybe I should back out that part of the change.

sagerb:
> This is a good point, maybe I should back out that part of the change.

I see you backed it out, which makes sense, although I wonder if we need some part of that.. in which we try the original URL, and if it doesn't work, we try stripping off the /connect? I do agree with folks that it can be very confusing. We'd need to update the client code to take the URL back from the response, to be used. We're already adding https:// if it doesn't exist, just with our client code calling format URL.

But we can do that as a separate PR.

sagerb (approved):
Code looks good to me. See my comment about the backed out code, but that can be a separate PR.

PR #1998, opened 2024-07-18
Activate deployment from entrypoint file

sagerb:
## Intent

Resolves: #1848 

This PR provides the functionality expected when clicking on the publish button for an entrypoint file.

The expected behavior when clicking on the publish button (which appears for an entrypoint file):

1. When there are no existing deployments for the entrypoint (set within the active configuration file) found within the same directory as the entrypoint file, the new deployment multi-stepper is initiated.
2. When there is is only one deployment for that entrypoint found, it is automatically selected as the active deployment.
3. When there are multiple compatible deployments found, and one of those is already selected, it is maintained.
4. When there are multiple compatible deployments found but a different or no deployment is actively selected, then we display a selection list of the compatible deployments.

In all cases, an informational message is displayed within VSCode with status and instructions on what to do next. VSCode provides the ability to hide these status and info messages if the user prefers that.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

There are a few changes worthy of calling out:

1. I discoverd the existence of race conditions with the `showDeploymentQuickPick` and `showNewDeploymentMultiStep`, since it relied upon the data initialized with the initial refresh. There is now a promise that is outstanding until the API calls within the initial refresh have been completed.
2. The error strings displayed for API failures now includes the `statusString` from the axios response.
3. The getAll for contentRecords was not passing in the correctly named parameter, so it's been fixed.

## Directions for Reviewers

To validate, you need to hit all of the scenarios listed above, with projects at different levels of a subdirectory.

What I did was:
1. create a new entrypoint file within an existing project
2. edit that file and click on the publishing button for the file
4. confirm that a new deployment flow is initiated. Go ahead and complete it and it will be selected as active.
5. select a different deployment
6. click on the publishing button for the new entrypoint file again, and the corresponding deployment should be selected.
7. create a new deployment and select the same entrypoint file.
8. select a different deployment for a different project
9. click on the publishing button for the entrypoint file again, and you should be presented with a selection between the two deployments which are available fo that entrypoint. select one.
10. once it is active, select the other one and then click on the publishing button for the entrypoint file and confirm that the active deployment is not changed (since it is compatible with the entrypoint file).

## Checklist


extensions/vscode/src/api/resources/ContentRecords.ts
```diff
@@ -18,7 +18,7 @@ export class ContentRecords {
   // Returns:
   // 200 - success
   // 500 - internal server error
-  getAll(params: { dir: string; entrypoints?: string; recursive?: boolean }) {
+  getAll(params: { dir: string; entrypoint?: string; recursive?: boolean }) {
```
sagerb, line 21: This was a spelling error for the proper param.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -362,8 +364,9 @@ export async function newDeployment(
         const python = await getPythonInterpreterPath();
         const inspectResponse = await api.configurations.inspect(
           {
-            dir: ".",
-            recursive: true,
+            dir: projectDir ? projectDir : ".",
+            entrypoint: entryPoint,
+            recursive: projectDir ? false : true,
```
sagerb, line 369: when we pass in a projectDir, we always disable recursive behavior. This was the way the logic / use worked out. In the future, if we do want something different, obviously we could add a new param to be passed in.

extensions/vscode/src/utils/errors.ts
```diff
@@ -12,6 +12,15 @@ export const getStatusFromError = (error: unknown): number | undefined => {
...
 
+export const getStatusStringFromErrorResponse = (
+  error: unknown,
+): string | undefined => {
+  if (axios.isAxiosError(error)) {
+    return error.response?.statusText;
+  }
+  return undefined;
+};
+
```
sagerb, line 23: The present of this `statusText` value proved to be helpful (when we received some of the more obscure HTTP error codes in a response).

extensions/vscode/src/utils/errors.ts
```diff
@@ -60,14 +72,16 @@ export const getSummaryStringFromError = (location: string, error: unknown) => {
...
   const code = getCodeStringFromError(error);
   const msg = getMessageFromError(error);
   const url = getAPIURLFromError(error);
 
-  if (stat || code || msg || url) {
+  if (status || statusText || code || msg || url) {
     return {
-      status: stat,
+      status,
+      statusText,
```
sagerb, line 84: This didn't sit well for me, so I just fixed it.

extensions/vscode/src/views/homeView.ts
```diff
@@ -90,6 +92,14 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
   private extensionUri: Uri;
   private webviewConduit: WebviewConduit;
+  private initCompleteResolver: (() => void) | undefined = undefined;
+
+  // Promise that methods can wait on, which will be resolved when
+  // the initial queries finish. Then they can use our internal collections of
+  // contentRecords, credentials and configs within this class
+  private initComplete: Promise<void> = new Promise((resolve) => {
+    this.initCompleteResolver = resolve;
+  });
```
sagerb, line 102: This functionality allows for methods to wait on this promise to be resolved. We have to save the the related resolve method to this particular promise, so that when we resolve it, those with a reference to the original promise will be signalled on its completion.

jonkeane:
> 2. When there is is only one deployment for that entrypoint found, it is automatically selected as the active deployment.

Like we talked about in standup: why not have it publish straight away at this point? 

> 3. When there are multiple compatible deployments found, and one of those is already selected, it is maintained.

Same for this one, why not just publish if the intent has already been made (relatively) clear? Or at the very least we should do (4) where we tell them to select (and then publish!)

> 4. When there are multiple compatible deployments found but a different or no deployment is actively selected, then we display a selection list of the compatible deployments.

We should have them select -> publish. We can give them an option at the end that's something softer if we really want "open the publisher" or something. But the vaaaaast majority of people clicking this want to get to publishing ASAP. Let's minimize the clicking needed.

> In all cases, an informational message is displayed within VSCode with status and instructions on what to do next. VSCode provides the ability to hide these status and info messages if the user prefers that.

This won't always be true, but if we're providing information / instructions on what to do next, we should really think if we're doing the right thing. The publisher should be simple enough that someone doesn't need to be told what to do within the UI. There are times where we might not quite get there, but I don't think this is the one place where that is required.

dotNomad:
> Like we talked about in standup: why not have it publish straight away at this point?

I have heard a few opinions and they all point to the expectation that this button publishes. That wasn't my original expectation (I felt like users would want to review before just sending up data to a server), but it is definitely the majority opinion that this should publish straight away.

While I think it should publish, I don't think we should do that in this PR though. This is already doing quite a bit to make the selection, lets get this in and modify to add the deploying after selection is made.

I'm reviewing this now 👀

jonkeane:
> While I think it should publish, I don't think we should do that in this PR though. This is already doing quite a bit to make the selection, lets get this in and modify to add the deploying after selection is made.

💯 fine by me. Iteration is ❤️

extensions/vscode/src/views/homeView.ts
```diff
@@ -771,62 +786,57 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
-        { dir: activeDeployment.projectDir },
-      );
-      showProgress("Updating Deployment", Views.HomeView, apiRequest);
-
-      await apiRequest;
+      // now select the new, updated or existing deployment
+      const deploymentSelector: DeploymentSelector = {
+        deploymentPath: targetContentRecord.deploymentPath,
+      };
+      this.propagateDeploymentSelection(deploymentSelector);
```
dotNomad, line 822: Do we need to propagate the selection here? We weren't before, and if I'm understanding this correctly it should already be picked up by the file watchers. I may be missing something else though.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1189,6 +1209,174 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
 
+  public async handleFileInitiatedDeployment(uri: Uri) {
+    // Guide the user to create a new Deployment with that file as the entrypoint
+    // if one doesn’t exist
+    // Select the Deployment with an active configuration for that entrypoint if there
+    // is only one
+    // With multiple, if a compatible one is already active, then do nothing.
+    // Otherwise, prompt for selection between multiple compatible deployments
+
+    // console.log("'Deploy with this Entrypoint' button hit!", uri);
```
dotNomad, line 1220:
```suggestion
```

extensions/vscode/src/views/homeView.ts
```diff
@@ -1189,6 +1209,174 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+
+    // get all of the deployments for projectDir
+    // get the configs which are filtered for the entrypoint in that projectDir
+    // determine a list of deployments that use those filtered configs
+
+    let contentRecordList: (ContentRecord | PreContentRecord)[] = [];
+    const getContentRecords = new Promise<void>(async (resolve, reject) => {
+      try {
+        const response = await api.contentRecords.getAll({
+          dir,
```
dotNomad, line 1238: Noting this in case I'm misunderstanding - we don't use the entrypoint filtering here since it doesn't capture pre-deployment records atm?

extensions/vscode/src/views/homeView.ts
```diff
@@ -1189,6 +1209,174 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+        window.showInformationMessage(
+          `Unable to continue with API Error: ${summary}`,
+        );
+        return reject();
+      }
+      resolve();
+    });
+
+    const apisComplete = Promise.all([
+      // getConfigurationInspections,
```
dotNomad, line 1287:
```suggestion
```

extensions/vscode/src/views/homeView.ts
```diff
@@ -1189,6 +1209,174 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+    const compatibleContentRecords: (ContentRecord | PreContentRecord)[] = [];
+    contentRecordList.forEach((c) => {
+      if (configMap.get(c.configurationName)) {
+        compatibleContentRecords.push(c);
+      }
+    });
+
+    // console.log(
+    //   `compatibleContentRecords: # found: ${compatibleContentRecords.length}`,
+    // );
```
dotNomad, line 1315:
```suggestion
```

extensions/vscode/src/views/homeView.ts
```diff
@@ -1189,6 +1209,174 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+    // if no deployments, create one
+    if (!compatibleContentRecords.length) {
+      // call new deployment
+      const selected = await this.showNewDeploymentMultiStep(
+        Views.HomeView,
+        dir,
+        entrypoint,
+      );
+      if (selected) {
+        window.showInformationMessage(
```
dotNomad, line 1326: Noted this in conversations, but once we get these deploying on-click I think we can remove these window notifications.

dotNomad (changes requested):
I really liked the additional functionality where if a selection for the entrypoint was already selected we didn't prompt. That was a good call to reduce user input.

---

I did hit the race condition solved in #1992 which I don't believe is on this branch. This occurred when I made a new deployment using the entrypoint button and it wasn't selected when the sidebar came into focus. The new Deployment was creatd successfully so that worked.

I'm going to request changes just to be sure that is merged in and it works with the fix in #1992.

---

After some playing around it looks like Configuration selection is not working as intended for a nested project. I have two configurations in a `.posit` dir and it is always prompting me to create a new one even with valid Configuration options.

extensions/vscode/src/views/homeView.ts
sagerb: Correct. It is found in the configuration file associated with the content record only, within pre-deployment files. After deployment, the last entrypoint is recorded within the content record. This is the field that the backend API is using for filtering.

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: This was incorrect. We collect recursively, but we also store them by subdirectory, so we can find unique names. I've added the recursive back in.

extensions/vscode/src/views/homeView.ts
```diff
@@ -771,62 +786,57 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     this.requestWebviewSaveSelection();
   }
 
-  private async showSelectConfigForDeployment() {
-    const activeDeployment = this.getActiveContentRecord();
-    if (activeDeployment === undefined) {
+  private async showSelectOrCreateConfigForDeployment(
+    targetContentRecord?: ContentRecord | PreContentRecord,
+    entryPoint?: string,
```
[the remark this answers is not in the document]
sagerb: This should be fixed with my next commit.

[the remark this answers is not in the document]
sagerb: > The optimizations here feel separate from the entrypoint button if I'm understanding it correctly. Would it be possible to break this out into a separate PR?

I'm not sure which ones you're referencing.

extensions/vscode/src/views/homeView.ts
sagerb: No, you are correct, they will be set by the state of the files. This is, as you suggest, a much cleaner way to have it happen.

extensions/vscode/src/views/homeView.ts
sagerb: done within my commit.

extensions/vscode/src/views/homeView.ts
sagerb: removed within my commit.

extensions/vscode/src/views/homeView.ts
sagerb: yes, they will be.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1227,12 +1409,12 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
       commands.registerCommand(
         Commands.HomeView.ShowSelectConfigForDeployment,
-        this.showSelectConfigForDeployment,
+        this.showSelectOrCreateConfigForDeployment,
         this,
       ),
       commands.registerCommand(
         Commands.HomeView.CreateConfigForDeployment,
-        this.createConfigForDeployment,
+        this.showSelectOrCreateConfigForDeployment,
```
dotNomad, line 1417: I wonder if we should just combine the `package.json` commands here since these two now do the same thing.

extensions/vscode/src/views/homeView.ts
```diff
@@ -765,62 +780,60 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
     }
-    const config = await selectConfig(activeDeployment, Views.HomeView);
-    if (config) {
-      const api = await useApi();
-      const apiRequest = api.contentRecords.patch(
-        activeDeployment.deploymentName,
-        config.configurationName,
-        activeDeployment.projectDir,
+    const targetConfiguration = this.getActiveConfig();
+    if (targetConfiguration === undefined) {
```
dotNomad, line 792: This if condition change makes it so when you get into the state where there is a missing configuration nothing happens when clicking the `showSelectOrCreateConfigForDeployment`

dotNomad (changes requested):
Still reviewing this, but caught another error.

extensions/vscode/src/views/homeView.ts
sagerb: See code update: https://github.com/posit-dev/publisher/pull/1998/commits/9d9c3696d234cf30deaca5ccd32aa728f1debc03

dotNomad (approved):
I have tested this pretty thoroughly on nested and non-nested workspace folder roots. This is look great! Moving onto the next PR.

PR #2005, opened 2024-07-22
Publish from entrypoint file click

### LINKED ISSUE #2003: Make entrypoint button publish on click
Follow-up to #1846 and #1848 to make the entrypoint button publish when it is clicked.

sagerb:
## Intent

Resolve #2003 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Initiating deployment after selection of existing or new deployment occurs.

I avoided race conditions with new deployments and configs by collecting the necessary API parameters for publishing directly from the multi-steppers that called the backend APIs after collecting the input from the user. Since they have confirmation from the APIs that the associated file updates were successful, then by using these same values, they are guaranteed to exist. This avoids trying to find the matching objects within the local caches, which are being asynchronously updated by the file watchers.

## Directions for Reviewers

Verify functionality within both mutli-folders and single folder projects for:

1. Clear out any configurations or deployments for the project
2. Do not activate the publishing extension
3. open one of the entrypoint files from the file explorer and click the publishing icon
4. verify that the new deployment flow is initiated and that it attempts to publish that entrypoint project automatically. Also confirm that this deployment is now selected.
5. select another deployment from this one created within the webview, then click the entrypoint's publish button again. Confirm it publishes it again (while updating the current deployment back to it).
6. add another deployment for this entrypoint, which should automatically select the new deployment. Click the entrypoint's publish button and it should publish the current one.
7. Switch to another deployment for a different project, then click on the entrypoint publishing button. Confirm you are presented with the ability to select between the two deployments which you created. Pick one of them and confirm that it is pubished.

## Checklist

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -410,7 +410,7 @@ export async function newDeployment(
       const response = await api.contentRecords.getAll(
         projectDir ? projectDir : ".",
         {
-          recursive: true,
+          recursive: projectDir ? false : true,
         },
       );
       const contentRecordList = response.data;
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -406,7 +406,7 @@ export async function newDeployment(
     try {
       const response = await api.contentRecords.getAll({
         dir: projectDir ? projectDir : ".",
-        recursive: false,
+        recursive: true,
       });
       const contentRecordList = response.data;
       // Note.. we want all of the contentRecord filenames regardless if they are valid or not.
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -46,8 +46,12 @@ import { WebviewConduit } from "src/utils/webviewConduit";
 import { fileExists, relativeDir, isRelativePathRoot } from "src/utils/files";
 import { newDeployment } from "src/multiStepInputs/newDeployment";
 import { Utils as uriUtils } from "vscode-uri";
-
-import type { DeploymentSelector, HomeViewState } from "src/types/shared";
+import type {
+  DeploymentSelectionResult,
+  DeploymentSelector,
+  HomeViewState,
+  PublishProcessParams,
+} from "src/types/shared";
 import {
   DeployMsg,
   EditConfigurationMsg,
@@ -786,10 +790,21 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     this.requestWebviewSaveSelection();
   }
 
+  private getCredentialNameForContentRecord(
+    contentRecord: ContentRecord | PreContentRecord,
+  ): string | undefined {
+    const credential = this.credentials.find(
+      (credential) =>
+        normalizeURL(credential.url).toLowerCase() ===
+        normalizeURL(contentRecord.serverUrl).toLowerCase(),
+    );
+    return credential?.name;
+  }
+
   private async showSelectOrCreateConfigForDeployment(
     targetContentRecord?: ContentRecord | PreContentRecord,
     entryPoint?: string,
-  ) {
+  ): Promise<DeploymentSelectionResult | undefined> {
     if (!targetContentRecord) {
       targetContentRecord = this.getActiveContentRecord();
     }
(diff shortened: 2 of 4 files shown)


extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -400,9 +407,12 @@ export async function newDeployment(
...
   const getContentRecords = new Promise<void>(async (resolve, reject) => {
     try {
-      const response = await api.contentRecords.getAll(".", {
-        recursive: true,
-      });
+      const response = await api.contentRecords.getAll(
+        projectDir ? projectDir : ".",
+        {
+          recursive: true,
+        },
```
mmarchetti, line 414: Does this need to run recursively when there's a project dir?

sagerb: Yes, indeed. Good catch. Fix pushed up.

dotNomad: Double checking this since it is still `projectDir ? false : true`

We don't actually want to be recursive when a projectDir, the same as `getConfigurationInspections` since the `projectDir` param was passed. That is done when the `showNewDeploymentMultiStep` gets a `projectDir` only when we initiate from a entrypoint file - so we don't want a recursion we only want the one entrypoint.

However PreDeployment records are not filtered on entrypoints so we have to do it manually here.

I just want to be sure I'm understanding that, and this code should stay the same?

extensions/vscode/src/views/homeView.ts
```diff
@@ -780,13 +784,26 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     this.requestWebviewSaveSelection();
   }
 
-  private async showSelectOrCreateConfigForDeployment() {
+  private getCredentialNameForContentRecord(
```
dotNomad, line 787: Perhaps we should change this to return the entire `Credential | undefined` then do `?.name` on the result just to make it more general purpose.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1199,6 +1271,13 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     }
   };
 
+  public initiatePublish(target: PublishProcessParams) {
+    this.onDeployMsg({
+      kind: WebviewToHostMessageType.DEPLOY,
+      content: target,
+    });
+  }
```
dotNomad, line 1279: This feels a bit strange to me. It isn't sending a message, but is passing it just because `onDeployMsg` is the thing that actually starts the deploy with `deployProject`.

Maybe we should split that logic out of there, and make `onDeployMsg` call the new function, and remove this middle man.

dotNomad (approved):
Only off case I could find was publishing without a Credential setup:

CleanShot 2024-07-23 at 16 30 45@2x

And the error message handled this very well.

The case where I didn't have a valid Configuration prompted to make an entirely new Deployment - that may not be what we want exactly, but it is a good first start, and we can adjust based on feedback.

---

I had some code cleanup comments, but nothing pressing that we cannot do as a follow-up.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1304,32 +1387,41 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+        projectDir: contentRecord.projectDir,
+      };
+      // publish!
+      return this.initiatePublish(target);
     }
     // if there are multiple compatible deployments, then make sure one of these isn't
     // already selected. If it is, do nothing, otherwise pick between the compatible ones.
     const currentContentRecord = this.getActiveContentRecord();
     if (
+      !currentContentRecord ||
```
dotNomad, line 1424: Oh weird not sure why this comment got lost. I was noting that this felt strange to me. Is this not already caught by the `find()` logic? I just want to be sure I'm not missing something in my understanding.

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: Yes, sorry, my bad. When there is no project dir, it needs to be run recursively. When there is a project dir passed in, there is no need for us to run recursively, because we won't have entrypoint selection across multiple dirs, which then selects which deployment names need to be considered.

The code has been updated ahead of you seeing it and it is correct.

PR #2013, opened 2024-07-23
Make dev versions always the latest

mmarchetti:
## Intent

This PR changes the dev version number to 99.99.99, so that dev versions will never be auto-"updated" to a marketplace version.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

Build the VSIX and install locally. Reload the window, check extension updates, etc. The Publisher should not be updated.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -4,7 +4,7 @@
   "author": "Posit, PBC",
   "displayName": "Posit Publisher",
   "description": "Publish and share your Python and R content on Posit Connect",
-  "version": "0.0.0",
+  "version": "99.99.99",
   "license": "MIT",
   "icon": "assets/img/color/posit-publisher.png",
   "repository": {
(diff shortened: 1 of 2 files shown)


extensions/vscode/package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Posit, PBC",
   "displayName": "Posit Publisher",
   "description": "Publish and share your Python and R content on Posit Connect",
-  "version": "0.0.0",
+  "version": "99.99.99",
```
sagerb, line 7:
```suggestion
  "version": "99.0.0",
```

sagerb (approved):
LGTM. I do think just the major version to 99 is enough, per my code suggestion.

dotNomad (approved):
This will make manual install SO much easier.

PR #2014, opened 2024-07-23
Handle configuration errors in selected configuration

dotNomad:
This PR addresses an issue where `ConfigurationError` selected configurations were not handled well:
- the edit button to edit the configuration would not appear
- the selection would not propagate to the webview
- the "Edit the Configuration" error link would not open the configuration for editting

This bug came from a combination of the refactor in #1961 removing updates to the selected configuration when configurations were sent to the frontend, and from the move to make selected configurations computed in the webview.

## Approach

The approach here was to always grab the selected Configuration even if it was in error and handle the typing in the areas that did not support `ConfigurationError` yet.

## Directions for Reviewers

- Test Configuration Errors (invalid schema) in a nested and non-nested workspace root
- Ensure the error and edit button both link to the correct erroring config
- Ensure the multi-selectors are correctly showing the erroring title
- Ensure that normal Configuration workflows function (selecting a new one, adding a new one, etc)

### DIFF
--- a/extensions/vscode/src/types/quickPicks.ts
+++ b/extensions/vscode/src/types/quickPicks.ts
@@ -2,6 +2,7 @@
 
 import {
   Configuration,
+  ConfigurationError,
   ContentRecord,
   PreContentRecordWithConfig,
 } from "src/api";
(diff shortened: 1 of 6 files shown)


sagerb (approved):
LGTM and have verified functionality.

PR #2042, opened 2024-07-26
Return empty array instead of nil

### LINKED ISSUE #2032: Permissions: when opening a folder that includes files/folders without user perms a confusing error is thrown
I hit this originally in Workbench, where this may occur more frequently than a local install but I was able to reproduce locally as well.

Using the Workbench instance here:
https://dev.palm.ptd.posit.it/

If I open the `/home/` directory which includes all Users. I can click Add Deployment but then this error is thrown:

> Unable to continue with project inspection failure. An error has occurred at newDeployment, configurations.inspect, Msg=Cannot read properties of null (reading 'forEach')

Screenshot 2024-07-25 at 1 52 21 PM

I was able to reproduce this locally by opening `/` (root) dir and clicking Add Deployment.

mmarchetti:
## Intent

Fixes #2032

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

There are a number of places in the Go code where we return nil when a result set would be empty. This PR changes those to return an empty slice. We only return nil on error returns, where the slice value won't be used by the caller.

## Automated Tests

Existing tests should pass.

## Directions for Reviewers

Open an empty folder, and try to add a deployment.

### DIFF
--- a/internal/clients/connect/client_connect_test.go
+++ b/internal/clients/connect/client_connect_test.go
@@ -60,7 +60,7 @@ type taskTest struct {
 }
 
 func (s *ConnectClientSuite) TestWaitForTask() {
-	var actualPackages []packageStatusEvent
+	actualPackages := []packageStatusEvent{}
 	emitter := events.NewMockEmitter()
 	emitter.On("Emit", mock.Anything).Run(func(args mock.Arguments) {
 		event := args.Get(0).(*events.Event)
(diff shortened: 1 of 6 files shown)


dotNomad (approved):
Fantastic. I assumed these returned empty arrays initially, but was unaware of the Go gotcha until #1938

Thanks for fixing these

PR #2048, opened 2024-07-29
Add start of Vitest infrastructure for Home webview

dotNomad:
This PR adds the beginnings of test infrastructure, in Vitest, for our Vue Home webview view

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

The approach here was to identify some utilities that didn't require much work at all to begin testing, and grab a core piece of code - in this case the `HostConduitService` and get some more complicated tests in place.

This sets up how we can use the Home store, the HostConduitService, and sending some messages (using Mocks where needed)

### DIFF
--- a/extensions/vscode/webviews/homeView/src/HostConduitService.test.ts
+++ b/extensions/vscode/webviews/homeView/src/HostConduitService.test.ts
@@ -1,0 +1,80 @@
+import { setActivePinia, createPinia } from "pinia";
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import { App, createApp } from "vue";
+
+import { HostToWebviewMessageType } from "../../../src/types/messages/hostToWebviewMessages";
+import { WebviewToHostMessageType } from "../../../src/types/messages/webviewToHostMessages";
+import { useHostConduitService } from "src/HostConduitService";
+import { useHomeStore } from "src/stores/home";
+import { vscodeAPI } from "src/vscode";
+import { WebviewApi } from "vscode-webview";
+
+export function withSetup<T>(composable: () => T): App {
+  const app = createApp({
+    setup() {
+      return { ...composable() };
+    },
+    template: "<div></div>",
+  });
+  app.mount(document.createElement("div"));
+  return app;
+}
+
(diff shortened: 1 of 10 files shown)


extensions/vscode/webviews/homeView/src/vscode.ts
```diff
@@ -0,0 +1,10 @@
+import { WebviewApi } from "vscode-webview";
+
+let apiInstance: WebviewApi<unknown> | undefined = undefined;
+
+export const vscodeAPI = (): WebviewApi<unknown> => {
```
dotNomad, line 5: This was a bit weird. I tried to get `vitest` to play nicely with `acquireVsCodeApi()`, but I just couldn't get anything to work. 

So instead I created this function / singleton to call it so it was easily mockable as a module import function.

extensions/vscode/webviews/homeView/src/HostConduitService.test.ts
```diff
@@ -0,0 +1,80 @@
...
+import { App, createApp } from "vue";
+
+import { HostToWebviewMessageType } from "../../../src/types/messages/hostToWebviewMessages";
+import { WebviewToHostMessageType } from "../../../src/types/messages/webviewToHostMessages";
+import { useHostConduitService } from "src/HostConduitService";
+import { useHomeStore } from "src/stores/home";
+import { vscodeAPI } from "src/vscode";
+import { WebviewApi } from "vscode-webview";
+
+export function withSetup<T>(composable: () => T): App {
```
dotNomad, line 12: We can probably break this out into another util tests file at some point, but this helper created a simple App that used the composable so `onMount()` and similar lifecycle functions worked for composables.

jonkeane:
Totally unrelated to the code but is this /vi:.tεst/ or /vit.εst/ (testing vi or the most vit?)

sagerb (approved):
Looks like a good beginning. Thanks for pulling this together.

I'll probably have some questions once I go to add some unit tests, but for now, this looks good enough to merge.

PR #2038, opened 2024-07-26
Handle empty notebook files

mmarchetti:
## Intent

Fixes #2036

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Detect when the JSON decoder returns io.EOF. Ignore the error, and don't return a configuration for this entrypoint (since an empty notebook file cannot be deployed).

## Automated Tests

Added a unit test.

## Directions for Reviewers

Open the sample-content directory and add a deployment (there's an empty notebook file in there).

PR #2051, opened 2024-07-30
Integrate fast entrypoint selection into new deployment multi-stepper

### LINKED ISSUE #2045: Show open entrypoints first in the list when adding a Deployment
We can create a separation between the open files on the top, and the rest on the bottom.

### LINKED ISSUE #2046: Show entrypoint list with unique files list
Right now we duplicate an entrypoint file in the selection list for each type we have detected. Ideally we would show the file one time so users can select it, then select the type they would like to use to make it a bit easier to make the initial selection.

sagerb:
## Intent

This PR updates the new deployment multi-stepper and breaks up the previous page which showed both entrypoint files as well as content types, into two pages. Switching to the two pages allows us to remove the expensive deep recursion required to inspect all possible entry points. With the change, we perform a quick directory walk looking only at the file extensions, and then once one of these is selected, we inspect the file to determine its content type / initial configuration file.

To support the selection of an entrypoint file, I have separated the list items for the potential entrypoint files that are opened within VSCode vs. the ones which are present within the VSCode workspace.

2024-07-30 at 3 49 PM

When the entrypoint selected could be deployed in multiple ways, we present an additional step to handle the selection.

2024-07-30 at 3 51 PM

When the entrypoint can be only one content type, we suppress this page.

Resolves #2045 
Resolves #2046 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

Changes included here:
1. The new `/entrypoints` API was integrated into the UX codebase
2. The agent code was updated to perform case-insensitive comparisons when identifying possible entrypoint files.
3. The flow associated with the new deployment multi-stepper was updated.

## Directions for Reviewers

Note that I have separated functionality into individual commits to help with a more understandable review.

These changes are only in effect for the newDeployment flow. To verify:
1. Invoke the new Deployment flow 
2. Select an entrypoint which can have multiple content types (like an RMD file), and verify that content-type selection page follows the entrypoint selection page
3. Verify that selection of an entrypoint file that can be only one type does not show the content-type selection page.
4. Verify that moving forward and back, update the current page numbers and total page numbers correctly.
5. Verify an error is displayed when there are no files that qualify as entry points available to be found.
6. Verify this functionality on large project hierarchies and small ones.

## Checklist

### DIFF
--- a/extensions/vscode/src/api/resources/Configurations.ts
+++ b/extensions/vscode/src/api/resources/Configurations.ts
@@ -7,6 +7,7 @@ import {
   ConfigurationDetails,
   ConfigurationError,
   ConfigurationInspectionResult,
+  EntryPointPath,
 } from "../types/configurations";
 
 export class Configurations {
@@ -95,5 +96,17 @@ export class Configurations {
         },
       },
     );
+  }
+
+  // Inspect the project, returning all possible (detected) configurations
+  // Returns:
+  // 200 - success
+  // 500 - internal server error
+  getEntrypoints(projectDir: string) {
+    return this.client.post<EntryPointPath[]>("/entrypoints", {
+      params: {
+        dir: projectDir,
+      },
+    });
   }
 }
--- a/extensions/vscode/src/api/resources/Configurations.ts
+++ b/extensions/vscode/src/api/resources/Configurations.ts
@@ -7,7 +7,6 @@ import {
   ConfigurationDetails,
   ConfigurationError,
   ConfigurationInspectionResult,
-  EntryPointPath,
 } from "../types/configurations";
 
 export class Configurations {
@@ -96,17 +95,5 @@ export class Configurations {
         },
       },
     );
-  }
-
-  // Inspect the project, returning all possible (detected) configurations
-  // Returns:
-  // 200 - success
-  // 500 - internal server error
-  getEntrypoints(projectDir: string) {
-    return this.client.post<EntryPointPath[]>("/entrypoints", {
-      params: {
-        dir: projectDir,
-      },
-    });
   }
 }
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -1,7 +1,6 @@
 // Copyright (C) 2024 by Posit Software, PBC.
 
 import path from "path";
-
 import {
   MultiStepInput,
   MultiStepState,
@@ -13,6 +12,7 @@ import {
 import {
   InputBoxValidationSeverity,
   QuickPickItem,
+  QuickPickItemKind,
   ThemeIcon,
   Uri,
   commands,
@@ -26,6 +26,7 @@ import {
   PreContentRecord,
   contentTypeStrings,
   ConfigurationInspectionResult,
+  EntryPointPath,
 } from "src/api";
 import { getPythonInterpreterPath } from "src/utils/config";
 import {
@@ -40,7 +41,7 @@ import { formatURL, normalizeURL } from "src/utils/url";
 import { checkSyntaxApiKey } from "src/utils/apiKeys";
 import { DeploymentObjects } from "src/types/shared";
 import { showProgress } from "src/utils/progress";
-import { isRelativePathRoot } from "src/utils/files";
+import { vscodeOpenFiles } from "src/utils/files";
 
 type stepInfo = {
   step: number;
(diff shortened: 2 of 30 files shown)


extensions/vscode/src/utils/files.ts
```diff
@@ -267,3 +269,50 @@ export function relativeDir(uri: Uri): string | undefined {
...
+ *
+ * @returns An array of relative paths (to the workspace)
+ *          NOTE: paths are not included if they are outside of the workspace
+ */
+export function vscodeOpenFiles(): string[] {
+  // build up a list of open files, relative to the opened workspace folder
+  const openFileList: string[] = [];
+  window.tabGroups.all.forEach((tabGroup) => {
+    tabGroup.tabs.forEach((tab) => {
+      const input = tab.input as TabInputText;
```
dotNomad, line 310: This initially felt concerning to me, but I tried a host of different types that wouldn't have URIs and since `relativePath` supports `undefined` I couldn't ever get bad behavior.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -490,21 +756,28 @@ export async function newDeployment(
...
     input: MultiStepInput,
     state: MultiStepState,
   ) {
+    // in case we have backed up from the subsequent check, we need to reset
+    // the array that it will update. This will allow steps to be the minimum number
+    // as long as we don't know it will take another one.
+    inspectionResults = [];
+
     // skip if we only have one choice.
     if (entryPointListItems.length > 1) {
```
dotNomad, line 779: This no longer works how we anticipate. I noticed while working in the `fastapi-simple` directory with only the `simple.py` file as an entrypoint I was still getting the file selection prompt.

I debugged it a bit, and it looks like the "Open Files" label is part of the array.

Perhaps we need to keep a separate array of entrypoints rather than looking at the items or something else?

CleanShot 2024-07-31 at 16 28 21

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -302,37 +470,101 @@ export async function newDeployment(
...
+            });
+          }
+        });
+      } catch (error: unknown) {
+        const summary = getSummaryStringFromError(
+          "newDeployment, configurations.inspect",
+          error,
+        );
+        window.showErrorMessage(
+          `Unable to continue with project inspection failure for ${entryPoint}. ${summary}`,
```
dotNomad, line 555: I get this error every time I attempt to use the entrypoint button without a Deployment already created.

CleanShot 2024-07-31 at 16 46 46@2x

dotNomad (changes requested):
I noticed that if I select the `empty.ipynb` notebook we have in the `sample-content` repo that I was not given an error - things just went forward. I think this is because the inspect API will return an unknown configuration when an entrypoint is specified.

But that leads me to a state where I have an error in my config (since `type = 'unknown'` is invalid). This also immediately causes an error in our refresh file lists API.

CleanShot 2024-07-31 at 16 35 42@2x

We should probably handle the files error better, but I wanted to start the discussion to decide on how we want this to behave.

---

I caught a few issues in comments, and the one above. This is very close, but I think we need to tighten it up a bit.

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -96,4 +97,16 @@ export class Configurations {
       },
     );
   }
+
+  // Inspect the project, returning all possible (detected) configurations
+  // Returns:
+  // 200 - success
+  // 500 - internal server error
+  getEntrypoints(projectDir: string) {
```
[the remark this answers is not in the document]
sagerb: I went ahead and made this change, as it is easily verifiable. 
https://github.com/posit-dev/publisher/pull/2051/commits/9578a4e165c327dfe711d6b933ed9899f4f3afdb

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: This has been resolved. See commit: https://github.com/posit-dev/publisher/pull/2051/commits/40b9a0ad82d8887725ca07cc9e701bc3a1946179

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: Looks like this was also resolved with the fix in commit: https://github.com/posit-dev/publisher/commit/40b9a0ad82d8887725ca07cc9e701bc3a1946179

sagerb:
> I noticed that if I select the `empty.ipynb` notebook we have in the `sample-content` repo that I was not given an error - things just went forward. I think this is because the inspect API will return an unknown configuration when an entrypoint is specified.
> 
> But that leads me to a state where I have an error in my config (since `type = 'unknown'` is invalid). This also immediately causes an error in our refresh file lists API.

I recreated the invalid configuration file (due to unknown), but I'm not seeing any refresh error. 

2024-08-02 at 4 24 PM

This feels like an unrelated bug to this PR. Can we separate it?

mmarchetti:
> I noticed that if I select the `empty.ipynb` notebook we have in the `sample-content` repo that I was not given an error - things just went forward. I think this is because the inspect API will return an unknown configuration when an entrypoint is specified.

`empty.ipynb` isn't a valid notebook - it is an empty file instead of valid JSON, so it can't be deployed. #2038 made this return an `unknown` configuration instead of erroring out during inspection.

extensions/vscode/src/multiStepInputs/newDeployment.ts
dotNomad: I'm getting a strange issue now where anytime I click the entrypoint button the homeview gets focus, but nothing else happens. It looks like the `newDeployment` function is bailed out of?

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -302,37 +471,101 @@ export async function newDeployment(
...
+  const getConfigurationInspectionQuickPicks = (
+    relEntryPoint: EntryPointPath,
+  ) => {
+    return new Promise<QuickPickItemWithIndex[]>(async (resolve, reject) => {
+      const inspectionListItems: QuickPickItemWithIndex[] = [];
+
+      try {
+        const python = await getPythonInterpreterPath();
+        const dir = path.dirname(relEntryPoint);
+        const entryPointFile = path.basename(relEntryPoint);
```
dotNomad, line 532: Looks like the issue is coming up here.

In a single folder workspace like `fastapi-simple` `entrypointFile = 'simple.py'` and `dir = '.'` which works. But with `sample-content` open I get those same results which causes the inspect API request to fail since `simple.py` is not in `sample-content` at the `dir=.` level

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -302,37 +471,101 @@ export async function newDeployment(
     return newCredentialForced(state) || newCredentialSelected(state);
   };
 
-  const hasMultipleEntryPoints = () => {
+  const hasMultiplePossibleEntryPointFiles = () => {
     return entryPointListItems.length > 1;
   };
```
dotNomad, line 476: This is still using `entryPointListItems` rather than `discoveredEntryPoints` so it returns `true` since the "Open Files" text is an item in the `entryPointListItems`.

This causes the steps to be incorrect when using the "Add New Deployment" in the Home View. This doesn't occur when using the entrypoint button since it doesn't use the "Open Files" bit.

dotNomad (approved):
Thanks for the additional work on this PR. It is looking fantastic.

I'm very comfortable improving `unknown` type Pre-Deployments in a follow-up. This is a huge step in making larger directories work well in the extension 🎉

PR #2053, opened 2024-07-31
Add new publisher icon and bundle it in an icon-font

### LINKED ISSUE #2054: Use new publisher icon and bundle in icon-font
Use the Publisher icon, and instead of using `png` files we should bundle our `svg` icons in an icon-font to have vector icons in our extension.

Other extensions and VSCode itself (for its Codicons) use `fantasticon` to do this.

dotNomad:
This PR introduces a new script `npm run build-icons` at the extension level which creates and bundles an icon-font and CSS file to utilize vector icons.

The `build-font` script is done on `vscode:prepublish` so shouldn't need to be run manually. It includes only the `posit-publisher-publish` icon which is our new logo.

This PR also updates the logo of the extension itself so the marketplace will update when we next publish a version.

<details>
  <summary>Preview</summary>
  
CleanShot 2024-07-30 at 18 30 36@2x
CleanShot 2024-07-30 at 18 31 23@2x

</details> 

## Intent

Resolves #2054 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

I used [`fantasticon`](https://github.com/tancredi/fantasticon) similar to [GitLens](https://github.com/gitkraken/vscode-gitlens/tree/main), and Microsoft's [vscode-codicons](https://github.com/microsoft/vscode-codicons).

I kept the `fantasticonrc` configuration as simple as possible, and only included our one icon, but this should be very easy to extend if we wish to include anymore.

I didn't use the icon in the home view, but I did include code to include the CSS and bundled the font file so if we wish to we can use it.

The icon SVG was adjusted to be:
- minimized
- edge to edge (similar to most VSCode codicons)

## Directions for Reviewers

Pull down the `vsix` and install it to see if the bundled icon works
Run in development mode and ensure the icon works

The icon should be visible in two spots:
- the sidebar
- the entrypoint button


extensions/vscode/assets/icons/template/mapping.json
```diff
@@ -0,0 +1,3 @@
+{
+  "publish": 61697
```
dotNomad, line 2: This was generated without the template, but figured including the mapping would make it easier to reference if used in the `package.json` for example.

extensions/vscode/assets/img/posit-publisher.png
dotNomad, line 1: This is now the icon that will be displayed in the extension marketplaces (when we publish)

sagerb (approved):
Changes look good and functionality is verified.

Issue #1970, opened 2024-07-16
Optimize recursive scans for large subdirectories


mmarchetti:
We need to examine any recursive scans being done when large folders (with many subdirectories and files) are opened within VSCode. Our operations seem to be taking a long time to update the UX, and this is probably due to the scans. (Only difference between acceptable, near instant response time with a smaller folder hierarchy vs. a large one.)

I would guess that there are some places where the refresh is not taking into account the projectDir, when understanding where to start.

We could also greatly improve this by reducing the number of parallel API calls we are doing since the extension does not have a single store of data.

You can also see that several refreshes are called multiple times very quickly.

mmarchetti:
Recursive APIs to look at:
* Configurations
* Deployments
* Inspection

mmarchetti:
Recursive inspection is being removed via the new `entrypoints` API. We still need to look at recursive fetching of deployments and/or configurations at extension activation time.

mmarchetti:
We've made improvements in 1.1.5 to avoid recursive inspection; the remainder of this work is moving out.

PR #2064, opened 2024-08-01
Split out state logic to `PublisherState` class

dotNomad:
This PR moves out some of the state logic in `homeView.ts` to a new `PublisherState` class to make it a bit easier to reason about our lists of resources (Content Records, Configurations, and Credentials) and the current selections.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

I kept the logic exactly the same, just moved the location of the code. I did it one resource at a time so the commits may be helpful to review this in segments.

### DIFF
--- a/extensions/vscode/src/state.ts
+++ b/extensions/vscode/src/state.ts
@@ -1,4 +1,4 @@
-import { Disposable } from "vscode";
+import { Disposable, ExtensionContext } from "vscode";
 
 import {
   Configuration,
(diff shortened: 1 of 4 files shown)


sagerb (approved):
This looks good as an incremental step towards what I found myself wanting in which the logic which manipulated and managed the state was also moved out of `homeView` into a separate state service.

A migration like this can be a bear if updates are being made in parallel while this PR is outstanding, and we have a few changes immediately targeted  which will build upon some of this code, I'm going to approve and merge this code in, so I can get started on reducing our initial recursive scans for #1970 

Looks like I need to update this code to remove a linting error. I'll take a look.  (Needs a copyright header, I'll fix it.(

PR #2071, opened 2024-08-05
Provide ability to dismiss deployment in progress

### LINKED ISSUE #2057: Feedback: Cancel hung deployment
From: https://positpbc.slack.com/archives/C06B0UZKT34/p1722398469369999

My deployment got stuck "waiting for the environment to be ready". It would be helpful if there was a way to cancel the deployment in VS Code.

Image

sagerb:
## Intent

Resolves #2057 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

As discussed with the team, we'll be implementing this within the UX only, while the agent continues to process the deployment in the background.

A new cancel button is available on the deployment in progress message. Clicking this causes a mock `publish/failure` message to be sent out and then causes all newer messages received within the same localId, to be discarded.

Components will continue to provide status per the file watchers, so a special canceled notification provides some direction as to what to expect.

2024-08-05 at 4 29 PM

2024-08-05 at 4 30 PM

## Directions for Reviewers

Validate full deployments complete.

Validate you can cancel one, seeing the error notification message as indicated above. Open the log and confirm none of the phases are still spinning.

Validate you can start another deployment after cancelling the last one and that all behavior is as expected.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/deployProgress.ts
+++ b/extensions/vscode/src/views/deployProgress.ts
@@ -9,32 +9,54 @@ export function deployProject(localID: string, stream: EventStream) {
     {
       location: ProgressLocation.Notification,
       title: `Deploying your project`,
-      cancellable: false,
+      cancellable: true,
     },
-    (progress) => {
+    (progress, token) => {
       let resolveCB: (reason?: any) => void;
       let rejectCB: (reason?: any) => void;
       const registrations: UnregisterCallback[] = [];
-
-      const unregiserAll = () => {
-        registrations.forEach((cb) => cb.unregister());
-      };
-
       const promise = new Promise<void>((resolve, reject) => {
         resolveCB = resolve;
         rejectCB = reject;
       });
+      let streamID = localID;
+
+      const unregisterAll = () => {
+        registrations.forEach((cb) => cb.unregister());
+      };
+
+      token.onCancellationRequested(() => {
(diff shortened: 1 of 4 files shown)


extensions/vscode/src/views/deployProgress.ts
```diff
@@ -9,32 +9,54 @@ export function deployProject(localID: string, stream: EventStream) {
...
-      const unregiserAll = () => {
+      const unregisterAll = () => {
         registrations.forEach((cb) => cb.unregister());
       };
 
-      const promise = new Promise<void>((resolve, reject) => {
-        resolveCB = resolve;
-        rejectCB = reject;
+      token.onCancellationRequested(() => {
+        streamID = "NEVER_A_VALID_STREAM";
```
dotNomad, line 29: I'm a bit confused why this setting of `streamID` is here.

dotNomad (approved):
Went through the directions for reviews and had no issues ✅

extensions/vscode/src/views/deployProgress.ts
sagerb: Just to make sure it NEVER matches when compared against incoming streamIDs, since there are still events flowing through the pipes.

PR #2089, opened 2024-08-07
Switch from env vars to file credentials

### LINKED ISSUE #2088: Use file in home directory for a persistent credential in Workbench
Environment variable support using `CONNECT_SERVER` and `CONNECT_API_KEY` did not work in VSCode in Workbench as the extension backend is launched prior to `.bashrc` being sourced. No `export` for environment variables can be seen even after reloading the window and restarting the session.

We need something persistent that we can fetch when the API call occurs so we are going to have a file to store credentials when there is no keyring to access.

mmarchetti:
## Intent

This PR implements a read-only single-credential file for use with Workbench until we can get full credential persistence. It replaces to environment variable method, since they are not available when workbench VSCode sessions start up.

Fixes #2088 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Read a file instead of environment variables.

## Automated Tests

Updated the existing tests.

## Directions for Reviewers

Create ~/.connect-credentials with a format like

```
url = 'https://rsc.radixu.com/'
key = 'n4Dm9V7or0bUkyvD6Q4l4dcY1eNDEqZs'
```

## Checklist

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -30,10 +30,10 @@ import (
 	"fmt"
 	"net"
 	"net/url"
-	"os"
 
 	"github.com/google/uuid"
 	"github.com/posit-dev/publisher/internal/util"
+	"github.com/spf13/afero"
 	"github.com/zalando/go-keyring"
 )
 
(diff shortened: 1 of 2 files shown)


dotNomad (approved):
I did notice that if you delete this credential we aren't doing any sort of updating to ensure that it is still around when we hit deploy.

We did this in the Credentials view by telling the bus to update, but we would have to do a file watcher setup here. This is similar for creation where we have to hit the refresh button in the Credentials view to pickup the file if it was created.

I think since this is a more temporary work around until we find better ways to persist when we don't have a keyring that this is acceptable.

---

Tested and was able to deploy with the new credential. I also used the `VSIX` and confirmed this worked on Workbench.

PR #2093, opened 2024-08-08
Sort configurations of the same entrypoint by type

### LINKED ISSUE #2050: When picking content type, prefer Quarto over RMD
We can sort these on the backend so Quarto is listed first responding to an inspect.

mmarchetti:
## Intent

This PR sorts returned configurations by entrypoint and then by type; currently they are only sorted by entrypoint.

* For an Rmd, it will sort the `quarto` configuration before the `rmd` one (or `quarto-shiny` before `rmd-shiny`).
* For a notebook, it will sort the `notebook` configuration before the `quarto` one.

Fixes #2050

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

The sort order isn't tested; this PR doesn't add a test for it.

## Directions for Reviewers

Inspect files the can have multiple deployment types, and see that the sort order of the offered configurations is correct.

## Checklist

### DIFF
--- a/internal/inspect/detectors/all.go
+++ b/internal/inspect/detectors/all.go
@@ -93,7 +93,11 @@ func (t *ContentTypeDetector) InferType(base util.AbsolutePath, entrypoint util.
 		} else if !aIsPreferred && bIsPreferred {
 			return 1
 		} else {
-			return strings.Compare(entrypointA, entrypointB)
+			if entrypointA == entrypointB {
+				return strings.Compare(string(a.Type), string(b.Type))
+			} else {
+				return strings.Compare(entrypointA, entrypointB)
+			}
 		}
 	}
 


internal/inspect/detectors/all.go
```diff
@@ -93,7 +93,11 @@ func (t *ContentTypeDetector) InferType(base util.AbsolutePath, entrypoint util.
 		} else if !aIsPreferred && bIsPreferred {
 			return 1
 		} else {
-			return strings.Compare(entrypointA, entrypointB)
+			if entrypointA == entrypointB {
+				return strings.Compare(string(a.Type), string(b.Type))
```
dotNomad, line 97: This looks like a alphanumeric string comparison so this PR just sorts the configuration types on that axis rather than actually prioritizing Quarto > RMD right?

dotNomad (approved):
Great to have some additional sorting to have consistent groupings of configurations.

internal/inspect/detectors/all.go
mmarchetti: Yes. The two cases we needed covered happened to be alphabetically ordered. I can map them to priority numbers and compare those, if this seems like too much of a shortcut.

dotNomad: Totally fine waiting on an actual map of priority until it is actually needed.

My fear here is that the code here doesn't represent our intention of prioritization so it is a bit "undocumented" other than this issue.

PR #2097, opened 2024-08-08
Add Tree Project Files view

### LINKED ISSUE #1210: Feedback: hierarchical Files view
https://positpbc.slack.com/archives/C06B0UZKT34/p1711545418302709

> Feedback: The Deployment Files >> Included Files listing offers no hierarchy, which means that using the UI to exclude files is tedious for Quarto projects, which render input.html and lots of files beneath input_files. It’s worse for Quarto websites and other, richer projects, which render everything beneath the some output directory like _site.The good thing is that I can watch the .positignore get built, notice that there are files contained within hierarchical paths, and adjust those rules as they’re added. It’s really nice that the include/exclude lists are automatically updated to reflect those edits.

dotNomad:
This PR adds a new view for the Project Files area of the Home view - the Tree view. With this PR it becomes the default view, and the Project Files view is opened by default to help users know which files are included with their deployments.

The views can be toggled via an action next to the "Refresh Files" button allowing the flat file view to still be used. This PR does not include any logic to remember the state of the Project Files view.

The new tree view has the same functionality:
- Shows added files with the `PostDecor` green `A`
  - This only works on files at the moment since directories are not listed in the Content Records, and I did not want to add recursion to detect if any nested files are new - to display the `A` at the folder level. This can be added later.
- Uses checkboxes to show the state of include/exclude and when clicked will submit an include or exclude API call
  - this also works on directories in this view
  - De-emphasized text is also used to help show that those files are excluded
  - Built In exclusions are disabled and have low-contrast text
- There is an action button to open the files
- Tooltips are the same between the two views

The tree view doesn't split up included and excluded files since it is displaying the file tree.

<details>
  <summary>Preview</summary>

https://github.com/user-attachments/assets/ceefc183-956e-469d-8d91-26f54d5b374f

</details> 

## Intent

Resolves #1210 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach here was to have `ProjectFiles` control which view was being displayed and the lower level components control the list or tree view.

Each view has shared logic for tooltips and included/excluded file logic which is now in util files.

Some CSS and component work was done to add the colorization element.

A new component was created call `TreeItemCheckbox` which is very similar to `TreeItem` but needed to be a separate component due to the logical differences. I tried to combine them or have a parent component, but it got a bit too messy and was easier to implement with shared code.

## Directions for Reviewers

Test all of the functionality in both Project Files views.


extensions/vscode/src/utils/files.ts
```diff
@@ -207,34 +205,6 @@ export function standalonePathSorter(sep: string) {
   };
 }
 
-export function splitFilesOnInclusion(
```
dotNomad, line 210: This functionality was moved to the util files TypeScript file at the webview level so the files as well as the `includedFiles` and `excludedFiles` list can be in the store.

extensions/vscode/webviews/homeView/src/components/views/projectFiles/ListProjectFiles.vue
dotNomad, line 1: This file is _extremely_ close to the original `ProjectFiles.vue` functionality. The only notable difference is the shared functions have been moved to util files.

mmarchetti:
This looks really nice. 

I did try to click on the icon (check or X) that indicates include/exclude status when I wanted to change that status; it opened the file instead. I guess I did that because the include/exclude button is hidden until you hover over the file, and the icon is suggestive of a checkbox. Not sure if we should change anything, but I wanted to mention that was my first experience.

sagerb:
This is a great start, but I wonder if we should change up the UX a little.

It would be nice if it more closely matched the file explorer interface:
2024-08-12 at 9 57 AM

in that we don't use the inclusion/exclusion state icon as the first thing you see, instead I think you should see the name. It took me a bit to see the difference of the check vs x.

I'd suggest we explore use of either the check and empty box to the right of the file name or we look at crossing out the file name, as a way of not including it.

dotNomad:
Based on the feedback I have re-done the Tree View to use checkboxes, made it open by default, and made the Tree View the default view.

I have re-written the description of the PR and included a new preview video.

sagerb (approved):
Just a little nit. As discussed, there are some very weird scenarios which the user can get into due to our implementation of ordered inclusion/exclusion, while the API simply adds or removes.

Let's merge this in and then we can figure out what to do about the scenarios.

jonkeane:
> As discussed, there are some very weird scenarios which the user can get into due to our implementation of ordered inclusion/exclusion, while the API simply adds or removes.

Were these discussed in the PR or synchronously? If the second, mind adding a description of those here too for posterity?

kgartland-rstudio:
The `.posit` dir doesn't appear as selected but it's deployed by default. I think we decided we always want to send the `.posit` dir so maybe it should be selected by default and un-deselectable?

kgartland-rstudio:
Usability nit: the entrypoint can be deselected and deployed. It was harder to deselect it in the previous view because you had to hit the little icon, but now since I can just click anywhere in the filename, it's easier to accidentally add/remove a file. I actually did that just now completely on accident.

kgartland-rstudio:
As Bill mentioned above, I was able to get into a weird state with nested directories.

Here I selected `subdir` and all the files/folders beneath were selected as expected but then I unchecked `subdir/subsubdir/newtext.txt` and `text.txt` and ended up with this state:
Screenshot 2024-08-22 093429


This ended up deploying all files as selected in the view which is good, but the Configuration file ended up being confusing. I didn't realize which rule would win here.

kgartland-rstudio:
This occurs in the previous Project Files view as well but something to consider...when you have nested directories, the directories get deployed whether they're selected or not. We get empty directories in Connect if they're unselected.

kgartland-rstudio:
If a file multiple directories down the tree is excluded in the tree view, when I switch to the list view, I cannot include the file (no `+` icon).  

I can't reproduce this now but I did get into a situation where if I had _included_ the file in the tree view, the file had the `-` icon but when I clicked it, nothing happened.  

I can't seem to reproduce it now but it makes me wonder if we really need two different views for the Project Files.  It seems like there may be some state issues when flipping back and forth.

dotNomad:
> I did try to click on the icon (check or X) that indicates include/exclude status when I wanted to change that status; it opened the file instead. I guess I did that because the include/exclude button is hidden until you hover over the file, and the icon is suggestive of a checkbox. Not sure if we should change anything, but I wanted to mention that was my first experience.

This has been fixed up! Now there are a few click targets:
- The "twisty" (the caret to open close directories) clicks to collapse or show children
- The checkbox and label section will check/uncheck the box
- The open action is clickable

extensions/vscode/webviews/homeView/src/components/views/projectFiles/ListProjectFiles.vue
```diff
@@ -0,0 +1,111 @@
...
+  >
+    <template #default="{ indentLevel }">
+      <TreeItem
+        v-for="file in home.flatFiles.includedFiles"
+        @click="
+          sendMsg({
+            kind: WebviewToHostMessageType.VSCODE_OPEN_RELATIVE,
+            content: { relativePath: file.id },
+          })
+        "
```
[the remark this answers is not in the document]
dotNomad: Since we added #2160 I will just remove this all in a follow-up PR

PR #2098, opened 2024-08-09
Ignore local imports during dependency scanning

### LINKED ISSUE #2096: Requirements scanning detects local imports as package dependencies
If you scan for Python dependencies and your code imports from another local Python file, that will be wrongly detected as a package dependency.

For example, in app.py:
```
from data import get_data
```

in data.py:
```
def get_data(): ...
```

When scanning, we need to see that `data` is a project-local import and not from an installed package.

mmarchetti:
## Intent

This PR omits local module/package import from being listed as package dependencies from the scanning process.

Fixes #2096

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Check for the existence of a module or local directory matching the top-level import name.

## Automated Tests

Updated an existing test to check for these cases.

## Directions for Reviewers

Scan for Python dependencies in the `posit-conf-survey` project. The local imports (`data`, `location`, etc) should be excluded from the generated `requirements.txt` file.

## Checklist

### DIFF
--- a/internal/inspect/dependencies/pydeps/project_imports.go
+++ b/internal/inspect/dependencies/pydeps/project_imports.go
@@ -43,7 +43,9 @@ func (s *defaultProjectImportScanner) ScanProjectImports(base util.AbsolutePath)
 			return err
 		}
 		code := ""
-
+		if info.IsDir() {
+			return nil
+		}
 		switch strings.ToLower(path.Ext()) {
 		case ".py":
 			contents, err := path.ReadFile()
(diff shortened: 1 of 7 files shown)


jonkeane:
This would be a much bigger change, so not needed in this PR, but do you want me to make an issue for the thought of extending this slightly to add the excluded file(s) to the file includes? Some of that depends on the outcome of questions about what minimal files to send are (e.g. if we include all `*.py` for apps, this is less useful)

dotNomad (approved):
Confirmed that local imports are no longer picked up with this change.

PR #2101, opened 2024-08-10
Skip recursive query of deployments and configurations if selection was saved and is present

sagerb:
## Intent

Part of addressing #1970 

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

If the extension is able to retrieve the previous selection from VSCode workspace storage, and the content record is valid and configuration can be found (valid or not), then skip the recursive API query for all deployments and configurations as we don't need the list yet.

This PR does also change the select deployment list to perform a full query first, matching the behavior of the multi-step flows. I was finding circumstances where the population of the content records within the homeView was not correct, so this change, while it does slow down the selection, does allow it to be correct.

We are balancing on the edge of needing to implement the data state within the homeview, and keep it up-to-date and accurate via the file-watchers, so that all of our views can just worry about data.

The result of these changes can be seen in the difference of these two videos, before and after:

https://github.com/user-attachments/assets/e0cad970-4a3f-4633-8f77-da498c05d8a6

https://github.com/user-attachments/assets/2e789e3f-8fef-48f1-b510-d782e91d5fff

## Directions for Reviewers

This PR depends upon another, which introduced versioning for the object we are storing within VSCode workspace storage for the user's last selection. If a valid selection is not found (as in the newest version), then you'll see the initial recursive retrieval, since we need to know if there are deployments or not.  

## Checklist

### DIFF
--- a/extensions/vscode/src/state.ts
+++ b/extensions/vscode/src/state.ts
@@ -176,7 +176,7 @@ export class PublisherState implements Disposable {
       return cfg;
     } catch (error: unknown) {
       const code = getStatusFromError(error);
-      if (code !== 400) {
+      if (code !== 404) {
         // 400 is expected when doesn't exist on disk
         const summary = getSummaryStringFromError(
           "getSelectedConfiguration, contentRecords.get",
--- a/extensions/vscode/src/state.ts
+++ b/extensions/vscode/src/state.ts
@@ -38,7 +38,7 @@ function findConfiguration<T extends Configuration | ConfigurationError>(
   configs: Array<T>,
 ): T | undefined {
   return configs.find(
-    (c) => c.configurationName === name && c.projectDir === projectDir,
+    (cfg) => cfg.configurationName === name && cfg.projectDir === projectDir,
   );
 }
 
@@ -46,7 +46,7 @@ function findCredential(
   name: string,
   creds: Credential[],
 ): Credential | undefined {
-  return creds.find((c) => c.name === name);
+  return creds.find((cfg) => cfg.name === name);
 }
 
 function findCredentialForContentRecord(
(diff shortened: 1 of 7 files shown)


mmarchetti:
> We are balancing on the edge of needing to implement the data state within the homeview, and keep it up-to-date and accurate via the file-watchers, so that all of our views can just worry about data.

Yes... the need is clear, we just need to plan when.

extensions/vscode/src/state.ts
```diff
@@ -105,23 +106,88 @@ export class PublisherState implements Disposable {
...
 
-  getSelectedConfiguration() {
-    const contentRecord = this.getSelectedContentRecord();
+  async getSelectedConfiguration() {
+    const contentRecord = await this.getSelectedContentRecord();
     if (!contentRecord) {
       return undefined;
     }
-    return this.findValidConfig(
+    const c = this.findValidConfig(
```
mmarchetti, line 157: nit: we've mostly used `cfg` as the generic name for Configurations.

extensions/vscode/src/state.ts
```diff
@@ -105,23 +106,88 @@ export class PublisherState implements Disposable {
...
+      );
+      const c = response.data;
+      // its not foolproof, but it may help
+      if (!this.findConfig(c.configurationName, c.projectDir)) {
+        this.configurations.push(c);
+      }
+      return c;
+    } catch (error: unknown) {
+      const code = getStatusFromError(error);
+      if (code !== 400) {
```
mmarchetti, line 179: Should this be 404? I think the get_configuration API only returns a 400 if the requested directory is outside the workspace.

mmarchetti (approved):

extensions/vscode/src/state.ts
sagerb: Yes, good catch!

Commit: https://github.com/posit-dev/publisher/pull/2101/commits/417a42a70bd9b88d5b5666a1864b87249a60a859

extensions/vscode/src/state.ts
sagerb: Yes, good catch!  Commit: https://github.com/posit-dev/publisher/pull/2101/commits/417a42a70bd9b88d5b5666a1864b87249a60a859

Issue #2103, opened 2024-08-12
Issue Reported: entrypoint API error when creating a deployment with Welcome Pages open


sagerb:
https://positpbc.slack.com/archives/C05D7NZD52S/p1723487355678719

```
Unable to continue with project entrypoint detection failure. An error has occurred at newDeployment, entrypoints.get, Msg=Cannot read properties of undefined (reading 'uri')
```

To reproduce, simply have an extension welcome page open and attempt a new deployment from the Publisher interface.

Issue #2069, opened 2024-08-05
Fatal Failures (where window is shown) are not leading the user to any resolution


mmarchetti:
It is akin to us throwing up our hands and declaring "oh well! something went wrong!"

We need to guide the user somewhere...

dotNomad:
We are now including a link to Troubleshooting documentation for some of the `window.showErrorMessage` usages to help users when the error isn't as actionable.

I'm going to close this one up, but if we have examples of specific errors we can open an issue for them - ideally individually.

PR #2105, opened 2024-08-12
Bug fix: where tab is non-text (with no input attribute)

sagerb:
## Intent

Resolves: #2103 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

1. Activate the extensions interface, and click on one of them so that their description page is shown
2. Activate the publisher extension, then hit the plus sign

Validate that the New Deployment entry point pick list displays, without error.

### DIFF
--- a/extensions/vscode/src/utils/files.ts
+++ b/extensions/vscode/src/utils/files.ts
@@ -314,9 +314,11 @@ export function vscodeOpenFiles(): string[] {
   window.tabGroups.all.forEach((tabGroup) => {
     tabGroup.tabs.forEach((tab) => {
       const input = tab.input as TabInputText;
-      const filePath = relativePath(input.uri);
-      if (filePath) {
-        openFileList.push(filePath);
+      if (input) {
+        const filePath = relativePath(input.uri);
+        if (filePath) {
+          openFileList.push(filePath);
+        }
       }
     });
   });


extensions/vscode/src/utils/files.ts
```diff
@@ -314,9 +314,11 @@ export function vscodeOpenFiles(): string[] {
...
-      const filePath = relativePath(input.uri);
-      if (filePath) {
-        openFileList.push(filePath);
+      if (input) {
+        const filePath = relativePath(input.uri);
+        if (filePath) {
+          openFileList.push(filePath);
+        }
       }
     });
```
jonkeane, line 323: This isn't directly related to the specific change here, which is good, but when hitting this error in person it made me think: shouldn't the inside of this for(each) loop be in a try/catch such that a failure on any individual tabs doesn't mean that other, just tabs which are just fine will still be presented?

mmarchetti (approved):

extensions/vscode/src/utils/files.ts
sagerb: With this change, we have achieved the same effect, since none of the inner function calls are expected to throw exceptions. Of course, that is only known by code browsing, vs. having language support for it like Java does.

I'm not currently confident that we have always taken the approach of degrading functionality upon error, vs. reporting a fatal error. I'm hopeful that we can look towards improving this in issue #2069.

PR #2106, opened 2024-08-12
Remove need for recursive deployment query from initial evenEasierDeploy

### LINKED ISSUE #2102: Eliminate the recursive query required to determine if we have any deployments configured
When activating the publisher and the workspace storage does not yet contain a past selection, we kick off recursive queries for all deployments and all configurations to determine if we are going to show the Home View button in

The Add Deployment State:
Image

or the Select Deployment State:
Image

This is a pretty high cost to pay when the user has opened a large project directory, just for the difference.

Let's discuss switching the interface to simply showing something like `Add or Select Deployment`, and updating the deployment selector to also allow a new deployment choice.

sagerb:
## Intent

Resolves #2102 

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

Updated EvenEasierDeploy UX so as to not require a full recursive query for deployments and configurations before showing the UX, when the workspace opened within VSCode does not have a previous selection recorded.

To do this, I've renamed the button to a more generic `Select Deployment...` and then updated the Select Deployment code to also include a list entry which allows creating a new deployment. I've also added two separators within the list, one for existing deployments and a second for new.

The following videos were recorded with an abnormally large project dir (my entire dev directory), but it nicely shows one of the worse cases we might see.

Before: Note the initial delay when we display "Scanning directories..." before the UX is enabled. (Roughly 20 seconds)
https://github.com/user-attachments/assets/c2e4dc89-b97c-4143-b815-785569e12b5f

After: Delay goes away! (Do not be confused with untouched delay in recursive queries when click on select deployment. This is same as before, but with this change, we've cut time for user to interact with Publisher down 50%.
https://github.com/user-attachments/assets/e7db3511-1445-4542-9655-75be74abe1a0

## Directions for Reviewers

Validate functionality when opening project folders which you've previously deployed (and VSCode has kept track of your past selection) and new ones. 
Validate deployment selection and new deployment creation, using very large project directories.

## Checklist

### DIFF
--- a/extensions/vscode/src/types/quickPicks.ts
+++ b/extensions/vscode/src/types/quickPicks.ts
@@ -9,7 +9,7 @@ import {
 import { QuickPickItem } from "vscode";
 
 export interface DeploymentQuickPick extends QuickPickItem {
-  contentRecord: ContentRecord | PreContentRecordWithConfig;
+  contentRecord?: ContentRecord | PreContentRecordWithConfig;
   config?: Configuration | ConfigurationError;
   credentialName?: string;
   lastMatch: boolean;
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -6,6 +6,7 @@ import debounce from "debounce";
 import {
   Disposable,
   ExtensionContext,
+  QuickPickItemKind,
   ThemeIcon,
   Uri,
   Webview,
@@ -913,7 +914,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
 
       // We need the latest lists. No choice but to refresh everything
       // once again.
-      await this.refreshAll(true);
+      await this.refreshAll(true, true);
 
       // Create quick pick list from current contentRecords, credentials and configs
       const deployments: DeploymentQuickPick[] = [];
@@ -923,11 +924,19 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
         ? projectDir
         : lastContentRecord?.projectDir;
       const lastConfigName = lastContentRecord?.configurationName;
+      const createNewDeploymentLabel = "Create a New Deployment";
 
       const includedContentRecords = contentRecordsSubset
         ? contentRecordsSubset
         : this.state.contentRecords;
 
+      if (includedContentRecords.length) {
+        deployments.push({
+          label: "Existing",
+          kind: QuickPickItemKind.Separator,
+          lastMatch: false,
+        });
+      }
       includedContentRecords.forEach((contentRecord) => {
         if (
           isContentRecordError(contentRecord) ||
@@ -999,6 +1008,16 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
         // Should we not push deployments with no config or matching credentials?
         deployments.push(deployment);
       });
+      deployments.push({
+        label: "New",
+        kind: QuickPickItemKind.Separator,
+        lastMatch: false,
+      });
+      deployments.push({
+        iconPath: new ThemeIcon("plus"),
+        label: createNewDeploymentLabel,
+        lastMatch: includedContentRecords.length ? false : true,
+      });
 
       const toDispose: Disposable[] = [];
       const deployment = await new Promise<DeploymentQuickPick | undefined>(
@@ -1034,8 +1053,13 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
         },
       ).finally(() => Disposable.from(...toDispose).dispose());
 
+      // If user selected create new, then switch over to that flow
+      if (deployment?.label === createNewDeploymentLabel) {
+        return this.showNewDeploymentMultiStep(Views.HomeView);
+      }
+
       let deploymentSelector: DeploymentSelector | undefined;
-      if (deployment) {
+      if (deployment && deployment.contentRecord) {
         deploymentSelector = {
           deploymentPath: deployment.contentRecord.deploymentPath,
           deploymentName: deployment.contentRecord.saveName,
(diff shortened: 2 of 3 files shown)


extensions/vscode/src/types/quickPicks.ts
```diff
@@ -9,7 +9,7 @@ import {
 import { QuickPickItem } from "vscode";
 
 export interface DeploymentQuickPick extends QuickPickItem {
-  contentRecord: ContentRecord | PreContentRecordWithConfig;
+  contentRecord?: ContentRecord | PreContentRecordWithConfig;
```
sagerb, line 12: Needed to make this optional to allow for Separator to be added

extensions/vscode/src/views/homeView.ts
```diff
@@ -913,7 +914,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
 
       // We need the latest lists. No choice but to refresh everything
       // once again.
-      await this.refreshAll(true);
+      await this.refreshAll(true, true);
```
sagerb, line 917: Need a complete refresh as well as an update to the selected state

extensions/vscode/src/views/homeView.ts
```diff
@@ -923,11 +924,19 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
         ? contentRecordsSubset
         : this.state.contentRecords;
 
+      if (includedContentRecords.length) {
+        deployments.push({
+          label: "Existing",
+          kind: QuickPickItemKind.Separator,
+          lastMatch: false,
+        });
+      }
```
sagerb, line 939: New seperators

extensions/vscode/src/views/homeView.ts
```diff
@@ -999,6 +1008,16 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+      deployments.push({
+        label: "New",
+        kind: QuickPickItemKind.Separator,
+        lastMatch: false,
+      });
+      deployments.push({
+        iconPath: new ThemeIcon("plus"),
+        label: createNewDeploymentLabel,
+        lastMatch: includedContentRecords.length ? false : true,
+      });
```
sagerb, line 1020: Creation entry and a separator to boot!

extensions/vscode/src/views/homeView.ts
```diff
@@ -1034,8 +1053,13 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
         },
       ).finally(() => Disposable.from(...toDispose).dispose());
 
+      // If user selected create new, then switch over to that flow
+      if (deployment?.label === createNewDeploymentLabel) {
+        return this.showNewDeploymentMultiStep(Views.HomeView);
+      }
+
```
sagerb, line 1060: Easiest way to handle creating a new deployment from within the selection quickpick.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1179,15 +1204,14 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
-        // search recursively for deployments
+      if (forceAll) {
+        // we have been told to refresh everything
         apis.push(this.refreshContentRecordData());
         apis.push(this.refreshConfigurationData());
       }
-      await Promise.all(apis);
+      const allApis = Promise.all(apis);
+      showProgress("Refreshing Data", allApis, Views.HomeView);
+      await allApis;
```
sagerb, line 1214: Progress indicator is new... somehow we missed showing one here previously. It sure does help when the subdirectory is huge.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -95,10 +82,10 @@
     <vscode-button
       v-else
       class="w-full add-deployment-btn"
-      @click="onAddDeployment"
+      @click="onSelectDeployment"
       data-automation="add-deployment-button"
     >
-      Add Deployment
+      Select Deployment...
```
sagerb, line 88: Simplified ux. Now we either show the currently selected deployment (in the way we have done before) or we show this new simple button. This is the biggest change which allows us to eliminate the upfront recursive query before the user can interact with the UX.

extensions/vscode/src/views/homeView.ts
mmarchetti, line 1020: I had a list of deployments in my connect-content directory, and with this at the bottom I had to scroll to see it. Maybe it can be at the top of the list?

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
mmarchetti: The fact that this is sometimes a button and sometimes a dropdown menu is a little jarring. I see why that's helpful (the button doesn't have to present any context) but I wonder if we can unify these somehow, maybe we could show the dropdown with the "Get deploying" message here?

extensions/vscode/src/views/homeView.ts
sagerb: I have implemented this, but I have split it off to a separate issue (https://github.com/posit-dev/publisher/issues/2110) and PR.

sagerb:
>maybe we could show the dropdown with the "Get deploying" message here?

I've gone ahead and implemented this. As discussed, we'll come back with the team to discuss the wording, but the concept is clean (to have both states similar within the UX).

See Commit: https://github.com/posit-dev/publisher/pull/2106/commits/0d6a6157ff762780a2caa82607d06cc719afbc65

PR #2109, opened 2024-08-13
Default to only including specific files

### LINKED ISSUE #2068: Feedback: initial deployment configuration should include only the single file rather than all files in project dir
From working with Garrett: 

Would like to select the files to be sent before deployment, if we stick with wanting to send everything. Otherwise, he would very much get behind only sending the entrypoint file itself by default, and then, if additional files were needed, the user would need to manually add them. He was very much on the side of not sending up every file by default.

mmarchetti:
## Intent

This PR eliminates the default `*` file pattern, in favor of only listing the entrypoint in most cases.

HTML and Quarto have some special handling to include other files.

Fixes #2068

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [x] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Remove the default `*` file pattern. Allow detectors to fill in the file list; if they don't, only include the entrypoint.

## Automated Tests

Some tests need added assertions to cover new functionality.

## Directions for Reviewers

Inspect projects of various types.

## Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -19,7 +19,7 @@ func New() *Config {
 		Schema:   schema.ConfigSchemaURL,
 		Type:     ContentTypeUnknown,
 		Validate: true,
-		Files:    []string{"*"},
+		Files:    []string{},
 	}
 }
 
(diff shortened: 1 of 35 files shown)


sagerb (approved):
Code changes look good to me.

Verified functionality, I think this makes much more sense and am happy to see the config files with the minimum files included.

PR #2111, opened 2024-08-13
Update quick pick items to be displayed in consistent manner when we offer selection of "new"

### LINKED ISSUE #2110: Feedback: Move "New" action to top of quick input list choices
We've had internal feedback that having the entry for a "new" config or deployment can be lost at the end, so it should be shown first.

If we do this, we should probably also fix up the consistency with the separators, to include the new / existing within the select configurations quick input list.

PR is already available for this change, creating this issue to collect team feedback. (https://github.com/posit-dev/publisher/pull/2111)

sagerb:
## Intent

Move items allowing "new" to top of list and make sure "existing" are sorted by title. Also making use of separators consistent.

Resolves #2110 

This PR updates the two quick input selectors that we use, which can allow a "new" operation rather than selecting an existing item, to have:
-  the "new" item is listed first and use two lines to match any existing entries.
- separators of "new" and "existing" added
- existing items sorted by the title

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

We have two operations which display the updated lists:
- The "select" deployment list:
deployment picker with new item shown first
- The "select" configuration list:
config picker with new item shown first

## Directions for Reviewers

Invoke operations for selecting a deployment and selecting a configuration for the deployment:
- making sure that "new" operation is spawned correctly when selected
- make sure that selection of an "existing" item selects the correct one and applies it.

### DIFF
--- a/extensions/vscode/src/multiStepInputs/selectNewOrExistingConfig.ts
+++ b/extensions/vscode/src/multiStepInputs/selectNewOrExistingConfig.ts
@@ -5,6 +5,7 @@ import path from "path";
 import {
   InputBoxValidationSeverity,
   QuickPickItem,
+  QuickPickItemKind,
   ThemeIcon,
   Uri,
   commands,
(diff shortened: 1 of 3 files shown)


mmarchetti (approved):
Love these changes!

Issue #2107, opened 2024-08-12
Bug: API failures when deleting .venv directory - So many API Calls!


sagerb:
See attached video for full steps to reproduce, but basically, just needed to create & activate a virtual environment for one of the python projects, install the dependencies into it with `pip install -r requirements.txt` and then deactivate the venv. Then while the publisher is active, delete the .venv directory with a `rm -rd .venv` command at the terminal.

The video is here:
https://github.com/user-attachments/assets/11921b24-31c8-4bf5-a541-327be2951eab

The agent log is interesting, so many queries!
```
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="Client version" version=1.1.5-10-gbd159a5d
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="Development mode" mode=dev
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="Development build" DevelopmentBuild=true
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="Starting PublishUICmd.Run"
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="created event server"
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="created event stream"
time=2024-08-12T16:27:36.311-07:00 level=INFO msg="created SSE logger"
time=2024-08-12T16:27:36.312-07:00 level=INFO msg="created UI service"
time=2024-08-12T16:27:36.315-07:00 level=INFO msg="UI server running" url=http://127.0.0.1:53809/
time=2024-08-12T16:27:37.347-07:00 level=INFO msg="Detecting deployment type and entrypoint..." path=/Users/billsager/dev/connect-content/bundles/stock-dashboard-python
time=2024-08-12T16:27:37.347-07:00 level=INFO msg="Possible deployment type" Entrypoint="" Type=unknown
time=2024-08-12T16:27:37.347-07:00 level=INFO msg="Getting Python version" python=/Users/billsager/.pyenv/versions/3.8.2/bin/python
time=2024-08-12T16:27:37.347-07:00 level=DEBUG msg="Running command" cmd=/Users/billsager/.pyenv/versions/3.8.2/bin/python args="-E -c import sys; v = sys.version_info; print(\"%d.%d.%d\" % (v[0], v[1], v[2]))"
time=2024-08-12T16:27:37.394-07:00 level=INFO msg="Detected Python" version=3.8.2
[...]

PR #2114, opened 2024-08-14
Handle files being deleted

mmarchetti:
## Intent

This PR introduces error handling in the file walkers for files not being found. This can happen if files are being removed while the walker is iterating.

Part of #2107

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Unit tests don't exercise this case currently.

## Directions for Reviewers

I ran this:
```
$(just executable-path) ui -v --listen=localhost:9001 &
while true; do curl -s -w '%{stderr}%{http_code}\n'  localhost:9001/api/configurations/default/files >/dev/null; done
```
in another shell:
```
uv venv .venv
us pip install notebook
cp -r .venv .venv2
rm -rf .venv2
```

Curl would always return a 500 at some point during the deletion; this should no longer happen.

## Checklist

### DIFF
--- a/internal/services/api/files/files.go
+++ b/internal/services/api/files/files.go
@@ -3,6 +3,8 @@ package files
 // Copyright (C) 2023 by Posit Software, PBC.
 
 import (
+	"errors"
+	"os"
 	"time"
 
 	"github.com/posit-dev/publisher/internal/bundles/matcher"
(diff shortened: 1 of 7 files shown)


sagerb (approved):
This change looks good and in testing it, I was unable to recreate the errors which I had seen in the original issue.

I did, however, get a TIMEOUT error the first time I was deleting the `.venv` directory, but was unable to reproduce that again.

I think we're good with this change!

PR #2115, opened 2024-08-14
Map common toplevel import names

### LINKED ISSUE #2113: Dependency scanning cannot reliably choose packages with different module names
For example:

[`python-dotenv`](https://pypi.org/project/python-dotenv/) has an import like:

```python
from dotenv import load_dotenv
```

and our scan will conclude that the package that should be included in the `requirements.txt` file is `dotenv`.

However the package name itself is `python-dotenv` and [`dotenv`](https://pypi.org/project/dotenv/) is a totally different package.

mmarchetti:
## Intent

This PR adds a fallback import -> package name mapping, to cover cases where the package has a non-matching import name and also isn't installed locally.

Fixes #2113

## Type of Change

- - [] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Define a default mapping of common packages that have mismatched import names. Unfortunately, this information isn't available by querying PyPI (see [discussion](https://discuss.python.org/t/record-the-top-level-names-of-a-wheel-in-metadata/29494/44)).

## Automated Tests

Updated an existing test to test this case.

## Directions for Reviewers

Create an app.py that imports a package with a mismatched name that you don't have installed locally yet, for example
```
import yaml
import dotenv
```

Scan for requirements. The resulting `requirements.txt` should contain
```
pyyaml
python-dotenv
```

## Checklist


jonkeane:
I wonder if a hard coded list like this is what we should do here. The list itself is fragile and susceptible to being not complete. For example `ibis` / `ibis-framework` isn't on there, it's almost certainly the case that people publishing to connect are using `ibis-framework` but in principle it's possible that someone is using [`ibis`](https://pypi.org/project/ibis/) where the rewrite would be very wrong. We also don't have insight into internal package names that are used within orginizations.

For the case where we have the package installed so can inspect the library, we're good. What if instead of rewritting these packages we warn in some way that there might be issues and that installing the libraries and rerunning will provide a better experience?

mmarchetti:
I agree that the list is likely to be fragile and incomplete; that's why I didn't use this approach originally. I like the suggestion of warning the user and giving them an opportunity to either correct the file or set things up better for a rescan (and for local testing). We have the same workflow when creating configuration files, where we are providing a solid starting point but can't guarantee everything is 100%.

PR #2118, opened 2024-08-15
Warn about packages not found during scanning

### LINKED ISSUE #2113: Dependency scanning cannot reliably choose packages with different module names

mmarchetti:
## Intent

This PR warns users when Python package scanning can't find one or more packages in the local library. 

Fixes #2113
A better alternative to #2115

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The package scanner returns a list of package specs, as well as a list of packages that were not found. The frontend warns the user if there are packages that couldn't be found.

## Automated Tests

Updated existing unit test in the backend.

## Directions for Reviewers

Add an import that of a package that isn't installed locally, then run the package scanner. You should see a warning message with instructions on how to resolve it.

## Checklist

### DIFF
--- a/cmd/publisher/commands/requirements_create.go
+++ b/cmd/publisher/commands/requirements_create.go
@@ -35,7 +35,7 @@ func (cmd *CreateRequirementsCommand) Run(args *cli_types.CommonArgs, ctx *cli_t
 		return errRequirementsFileExists
 	}
 	inspector := inspect.NewPythonInspector(absPath, cmd.Python, ctx.Logger)
-	reqs, _, err := inspector.ScanRequirements(absPath)
+	reqs, incomplete, pythonExecutable, err := inspector.ScanRequirements(absPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


sagerb (approved):
Code changes look good. I was a little thrown in what `requirements_show` was used for, but I figured out it is for the CLI. 

Verified import with and without locally installed package.

When missing, this is the message:
```
Could not find installed packages for some imports using /Users/billsager/.pyenv/versions/3.8.2/bin/python. Install the required packages, or select a different interpreter, and try scanning again. Imports: stuff
```

Issue #2119, opened 2024-08-15
When redeployment fails we can no longer update existing deployment


kgartland-rstudio:
I have an existing successfully deployment. If I change my Python version to an invalid one and re-deploy, all Deployment history is lost and I can no longer update my deployment on the server.

STR:
* Create a deployment with a Python app
* Once it succeeds, update the config with an invalid Python version
* Redeploy so a failure occurs

The deployment history is cleared and I now have no way to update the existing deployment.  This seems to occur with any failure.

This instead should keep the Deployment history and allow me to update the existing deployment.

Example:

https://github.com/user-attachments/assets/0b7e97d8-7a6c-401e-ac15-fdfa549848d8

dotNomad:
I believe this is fixed with the merge of #2133. @kgartland-rstudio can you confirm that you are no longer seeing this behavior?

kgartland-rstudio:
Closing, can no longer reproduce.

PR #2134, opened 2024-08-16
Show anchor for editing configuration within error text

### LINKED ISSUE #2047: Provide configuration name and link to error message when we suggest editing the configuration file.
Proper action requires user to "know" that editing the configuration file is accomplished by clicking the pencil icon half a screen above the error message.

Image

sagerb:
## Intent

Resolves #2047 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Implementation looks for a very specific string and if found within the error text, splits the string and inserts an anchor into the HTML. It is only implemented for this single error, but we can move forward with a more generalized approach using the same techniques when we determine other messages that should share this behavior.

anchor

## Directions for Reviewers

Recreate two errors, one of which is the Python mismatch for which this PR is targeting. Confirm this behavior only applies to the Python version mismatch and does not have any impact on other messages.

## Checklist


dotNomad (commented):
[#2119](https://github.com/posit-dev/publisher/issues/2119) is keeping from me from being able to test this without fixes.

When I deploy with an error the deployment is removed from the UI entirely.

Once that is fixed we can come back to this.

sagerb:
@dotNomad I think you're good to re-review this. I've merged in Mike's fix and I don't have any problem seeing/viewing deployments with errors now. Can you verify that you are also not blocked?

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -190,6 +207,9 @@ import { filterConfigurationsToValidAndType } from "../../../../src/utils/filter
 const home = useHomeStore();
 const hostConduit = useHostConduitService();
 
+const editingSplitStr = "editing";
+const editConfigFileDetectionStr = "editing your configuration file";
```
dotNomad, line 211: Honestly I would combine these. I think it would be nice if "editing your configuration file" was the link, and it would consolidate a bit of the logic.

dotNomad (approved):
I was able to see it this time! Works great. Small note about combining some logic.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: I did start with that, but didn't like it (see below). The logic consolidation is pretty minor, so I would like to stick with what I've got (the second option below):

2024-08-22 at 8 57 AM

vs.

2024-08-22 at 9 08 AM

jonkeane:
Sorry I'm a bit late on this: does it look like there's an extra space at the front of the first line?

Also what about "...consider editing your configuration to have one of the versions: 3.11.3"

Slightly more concise + direct. Also would be super slick to condition on number of versions and do "To be version: 3.11.3" if there's one and then "one of versions: ..." if there's more than one

sagerb:
> Sorry I'm a bit late on this: does it look like there's an extra space at the front of the first line?
> 
> Also what about "...consider editing your configuration to have one of the versions: 3.11.3"
> 
> Slightly more concise + direct. Also would be super slick to condition on number of versions and do "To be version: 3.11.3" if there's one and then "one of versions: ..." if there's more than one

Space has been eliminated from DOM with the refactoring I just checked in.

We're receiving the error messages from the agent, but I'll go ahead and include the change per your request.

sagerb:
@dotNomad Do you mind taking a look at the commit I added to create the new Vue component? https://github.com/posit-dev/publisher/pull/2134/commits/439446684df1887c162f832b219d130022758181

I'll also let you know when I've got the agent code change pushed up.

extensions/vscode/webviews/homeView/src/components/TextStringWithAnchor.vue
```diff
@@ -0,0 +1,63 @@
...
+defineOptions({
+  inheritAttrs: false,
+});
+
+const emit = defineEmits<{
+  // <eventName>: <expected arguments>
+  click: [splitOptionId: number]; // named tuple syntax
+}>();
+
+interface SplitOption {
```
dotNomad, line 26: May want to export this since it is used as prop, will make it easier to construct with type safety in other components.

extensions/vscode/webviews/homeView/src/components/TextStringWithAnchor.vue
dotNomad, line 1: Splitting this off into a separate component was the right move 🎉

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -349,6 +358,36 @@ const toolTipText = computed(() => {
 - Server URL: ${home.serverCredential?.url || "<undefined>"}`;
 });
 
+const errorMessage = computed(() => {
+  return home.selectedContentRecord?.deploymentError?.msg || "";
+});
```
dotNomad, line 363: I would just use `home.selectedContentRecord.deploymentError.msg` directly in the template to get the type narrowing. That way we won't need to have the `|| ""`

dotNomad (commented):
Had some comments with the component work.

sagerb:
Updated images for latest changes:
2024-08-23 at 10 34 AM

2024-08-23 at 10 35 AM

With these changes, we now display a button within the error message display (at time of publishing) as well as the already implemented link within the last deployment display.

These latest changes also refactored the logic into sharable component and files so that I could implement this new button within the logs view.

@dotNomad Do you mind taking a new look at this PR? Things have dramatically changed since your earlier review.

sagerb:
@jonkeane How do the screenshots look to you above? They've got the changes you requested, and I think it was a great improvement.

extensions/vscode/webviews/homeView/src/components/TextStringWithAnchor.vue
```diff
@@ -0,0 +1,63 @@
...
+  detectionStr: string;
+  anchorStr: string;
+  id: number;
+}
+
+interface Props {
+  message: string;
+  splitOptions: SplitOption[];
+  class?: string;
```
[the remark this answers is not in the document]
sagerb: Since I have multiple root elements within the component, I was doing this explicitly... but looks like I can use `$attrs.class` to go ahead and address that issue in a better way. OK, I will remove that prop.

extensions/vscode/webviews/homeView/src/components/TextStringWithAnchor.vue
```diff
@@ -0,0 +1,63 @@
...
+    {{ message }}
+  </span>
+</template>
+
+<script setup lang="ts">
+import { computed } from "vue";
+
+defineOptions({
+  inheritAttrs: false,
+});
```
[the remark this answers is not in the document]
sagerb: Good catch, that was left over from a debug session. Sorry about that!

extensions/vscode/webviews/homeView/src/components/TextStringWithAnchor.vue
sagerb: That has now been refactored into a common utility (and exported), since my next to last commit (counting this one which I'm making your requested changes on) needed the functionality within the log view.

extensions/vscode/webviews/homeView/src/components/TextStringWithAnchor.vue
sagerb: I think so too! :-)

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: great! Although I'm not sure why that matches the prop type of string... But it does work. I suppose it is because it still resolves to "" when that message is somehow undefined?

extensions/vscode/src/constants.ts
```diff
@@ -74,6 +74,7 @@ const homeViewCommands = {
   NavigateToDeploymentContent:
     "posit.publisher.homeView.navigateToDeployment.Content",
   ShowContentLogs: "posit.publisher.homeView.navigateToDeployment.ContentLog",
+  EditCurrentConfiguration: "posit.publisher.homeView.edit.Configuration",
```
dotNomad, line 77: It looks like the command added was `posit.publisher.edit.Configuration` and did not include the `homeView`. We can change it here or in the `package.json`

dotNomad (changes requested):
Just caught one thing we have to fix before merging.

dotNomad (approved):
Thanks for making those strings match :shipit:

Issue #2142, opened 2024-08-20
add additional files for deploy tests 


kgartland-rstudio:
Since we now only deploy the entrypoint file by default, the deploy tests need to add additional files to the config prior to deployment.

PR #2145, opened 2024-08-20
add additional files to deploy tests

kgartland-rstudio:
Since we now only deploy the entrypoint file by default, the deploy tests had to be adjusted to allow for additional files to be added to deployments.

fixes: https://github.com/posit-dev/publisher/issues/2142


dotNomad (approved):
This is looking good to me. As long as CI passes should be good to merge 👍

PR #2163, opened 2024-08-22
Files api fields for recursive inclusion/exclusion

### LINKED ISSUE #2147: Include attributes for directories in files API if all nested files are included or excluded
For the hierarchical tree view with checkboxes we want to have the folders in the tree view to have an [indeterminate state](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes) so the three states are:

- Checked if all nested files are included
- Unchecked if all nested files are excluded
- Indeterminate if some nested files are included and some are excluded

To do this we can include attributes if all files are included and another if all files are excluded.

We could collapse these into a single attribute with different enum states as well. As long as we can determine the three states above.

### LINKED ISSUE #2158: Print files list in multiline format in configurations

mmarchetti:
## Intent

This PR adds two new fields to the Files API response: `allIncluded` and `allExcluded`.

* For a file, these trivially indicate whether the file was included or excluded.
* For a directory, these indicate whether everything in the directory tree is included, or all excluded, It's possible/likely for both to be false, and impossible for both to be true.

Fixes #2147 
Fixes #2158

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Postorder traversal of the constructed file tree response, rolling up results from lower levels.

## Automated Tests

Added a couple unit tests to cover this.

## Directions for Reviewers

```
just build
$(just executable-path) ui -vv --listen=localhost:9001 &
# ensure there is a configuration ($NAME) with desired patterns listed in `files`
$ curl localhost:9001/api/configurations/$NAME/files?dir=. | jq
```


internal/services/api/files/services.go
```diff
@@ -62,5 +62,23 @@ func (s filesService) GetFile(p util.AbsolutePath, matchList matcher.MatchList)
...
+		file.AllExcluded = !included
+		return
+	}
+	file.AllIncluded = true
+	file.AllExcluded = true
+
+	for _, child := range file.Files {
+		updateInclusions(child)
+		file.AllIncluded = file.AllIncluded && child.AllIncluded
+		file.AllExcluded = file.AllExcluded && child.AllExcluded
```
dotNomad, line 82: Cool way of accomplishing this 👍

dotNomad (approved):
Pulled this down and immediately tested with `sample-content` and a call to the configurations files api. Everything I got back was as expected. Looks great!

PR #2164, opened 2024-08-22
Remove unused ContentRecords view const

dotNomad:
Just cleaning up an unused const.


sagerb (approved):
Too difficult to understand this change... LOL.

Looks great!

PR #2167, opened 2024-08-22
Remove some throws which were left in the web view

sagerb:
## Intent

@dotNomad is looking into why we aren't applying the linting rules to the webview, but these came up because VSCode does (and we want it to).

This PR contains quick fixes to remove those throws.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

In these cases, I simply output a console error and abort processing the function call. These conditions are all "assert-like" catches for us, so this approach should work fine.

## Directions for Reviewers

Just a quick regression should be fine.

## Checklist

### DIFF
--- a/extensions/vscode/webviews/homeView/src/HostConduitService.ts
+++ b/extensions/vscode/webviews/homeView/src/HostConduitService.ts
@@ -38,9 +38,10 @@ export function useHostConduitService() {
 
   const sendMsg = (msg: WebviewToHostMessage) => {
     if (!hostConduit) {
-      throw new Error(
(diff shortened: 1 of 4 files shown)


dotNomad:
Looks like this has duplicate commits to what is on #2134?

sagerb:
> Looks like this has duplicate commits to what is on #2134?

Yes, good catch. I've now changed the #2134 PR to have a base of this PR (as I had merged this branch into it, to get rid of the linting errors).

dotNomad (approved):
Looks great 😎

PR #2169, opened 2024-08-22
Enable command URIs and remove view log message

dotNomad:
This is a change to use `enableCommandUris` in our webview to use `href` commands to run registered commands in VSCode.

This allows us to remove messages when the messages were only running VSCode commands.

Some helpful documentation and references:
- https://code.visualstudio.com/api/extension-guides/command
- https://code.visualstudio.com/api/extension-guides/command
- https://www.eliostruyf.com/command-uri-vscode-webview-open-files-links/

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling


dotNomad:
This does not work in Workbench. It seems like the VSCode running (at least in `dev.demo.posit.team`) does not work with the command Uri `href` links.

Looking into opening up an issue, and I'll close this up since we cannot implement it without it working in Workbench.

PR #2173, opened 2024-08-23
Add nested directory UI tests

kgartland-rstudio:
This PR adds some nested directory testing. It finds all the detected entrypoints and selects one then verifies the config created.

Webdriver.io doesn't allow you to change directories when it's running so I had to introduce some environment variables to declare the spec to be run and what directory we should open.  

I also snuck in a change to wait for any package locks when installing Connect on fuzzbucket, that was causing flakey failures in the nightly tests.


dotNomad (approved):
Very awesome to have some tests covering this.

PR #2174, opened 2024-08-23
Default included files to the relative root dir

### LINKED ISSUE #2159: When including or excluding files include the file path rather than just the file name

dotNomad:
This PR changes two things:
- when include or excluding a file in the extension UI a `/` is prepended to the file URI when it is sent to the API
- when a file is included in the created Configuration a `/` is prepended to them

This restricts the include to only the directory and _not_ its subdirectories avoiding some confusion when using the Project Files view.

## Intent

Resolves #2159

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

Ensure newly created Configurations work as expected
Ensure that the include and exclude actions in the Project Files view work as expected

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -324,7 +324,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       const api = await useApi();
       const apiRequest = api.files.updateFileList(
         activeConfig.configurationName,
-        uri,
+        `/${uri}`,
         action,
         activeConfig.projectDir,
       );
--- a/internal/initialize/initialize.go
+++ b/internal/initialize/initialize.go
@@ -133,7 +133,7 @@ func normalizeConfig(
 	// The inspector may populate the file list.
 	// If it doesn't, default to just the entrypoint file.
 	if len(cfg.Files) == 0 {
-		cfg.Files = []string{cfg.Entrypoint}
+		cfg.Files = []string{fmt.Sprint("/", cfg.Entrypoint)}
 	}
 	needPython, err := requiresPython(cfg, base)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -324,7 +324,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       const api = await useApi();
       const apiRequest = api.files.updateFileList(
         activeConfig.configurationName,
-        uri,
+        `/${uri}`,
```
sagerb, line 327: Do you think we should be defaulting to the proper platform separator here? So we get forward or backward slashes prepended?

sagerb (commented):
Let's discuss which way we want the slashes within the configuration file, especially here where users may type them themselves. 

With collaboration, we need to be able to be compatible across all platforms. I wonder if we need to insert them, but then convert them internally? 

Seems like we need to nail this down. Perhaps it is only my confusion, as I understand we want to use forward slashes primarily, but that is harsh to Windows users.

extensions/vscode/src/views/homeView.ts
mmarchetti: The config file is expecting gitignore patterns, which always use forward slashes. We can change that if needed, but will need to convert somewhere (likely the backend).

mmarchetti: I think it's better to stick with slashes, because there will be cases where teams collaborate across Windows and non-Windows OSes.

sagerb: OK, I'm good with this approach. We just need to make it consistent and convert from back slashes to forward slashes when the user provides them (by manually editing the file)?

sagerb (approved):
Changes look good and question was answered.

PR #2182, opened 2024-08-26
Use titles to name configuration files in addition to short hash

### LINKED ISSUE #2067: Feedback: name the configuration files based on the project title initially provided.
Feedback from Garrett: https://positpbc.slack.com/archives/C06B0UZKT34/p1722620392603039

dotNomad:
This PR introduces a new naming setup for configuration files where they are named after the provided title of the content.

The file names also include a short unique string of base 32 characters for example: `fastapi-simple-AU9Q` to avoid collisions with collaborations via Git.

## Intent

Resolves #2067

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

The approach here was to avoid enumerating configurations and start using the title. To achieve file name safety a new dependency was brought in - [filenamify](https://github.com/sindresorhus/filenamify).

The random characters are created with `Math.random()` since we do not need cryptographic security on these. They then are converted from numbers to strings with the radix `32`.

## Directions for Reviewers

Test out the two ways we have to create configurations:
- the add Deployment flow
- the Select Active Configuration for Deployment flow, Create a New Configuration option

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -497,6 +497,7 @@
     "test": "npm run compile && node ./out/test/runTest.js"
   },
   "devDependencies": {
+    "@twbs/fantasticon": "^3.0.0",
     "@types/eventsource": "^1.1.15",
     "@types/mocha": "^10.0.2",
     "@types/mutexify": "^1.2.3",
@@ -510,7 +511,6 @@
     "esbuild": "^0.21.4",
     "eslint": "^8.50.0",
     "eslint-config-prettier": "^9.1.0",
-    "@twbs/fantasticon": "^3.0.0",
     "glob": "^10.3.3",
     "mocha": "^10.2.0",
     "typescript": "^5.2.2"
@@ -522,6 +522,7 @@
     "axios": "^1.7.4",
     "debounce": "^2.1.0",
     "eventsource": "^2.0.2",
+    "filenamify": "^6.0.0",
     "get-port": "5.1.1",
     "mutexify": "^1.4.0",
     "retry": "^0.13.1",
--- a/extensions/vscode/src/utils/names.ts
+++ b/extensions/vscode/src/utils/names.ts
@@ -2,38 +2,8 @@
 
 import { InputBoxValidationSeverity } from "vscode";
 
-import { useApi } from "src/api";
 import { isValidFilename } from "src/utils/files";
-
-export async function untitledConfigurationName(
-  projectDir: string,
-): Promise<string> {
-  const api = await useApi();
-  const existingConfigurations = (await api.configurations.getAll(projectDir))
-    .data;
-
-  if (existingConfigurations.length === 0) {
-    return "configuration-1";
-  }
-
-  let id = 0;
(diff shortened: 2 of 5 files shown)


extensions/vscode/package.json
```diff
@@ -510,7 +511,6 @@
     "esbuild": "^0.21.4",
     "eslint": "^8.50.0",
     "eslint-config-prettier": "^9.1.0",
-    "@twbs/fantasticon": "^3.0.0",
```
dotNomad, line 513: This got re-ordered when using `npm install`

extensions/vscode/src/utils/names.ts
```diff
@@ -82,3 +52,45 @@ export function contentRecordNameValidator(
...
+ *   uniqueness
+ * @returns A filename that is safe to use in the filesystem with a unique 4
+ * character ending to avoid Git conflicts.
+ */
+export function newConfigFileNameFromTitle(
+  title: string,
+  existingNames: string[] = [],
+): string {
+  const filename = filenamify(title, {
+    replacement: "-",
```
sagerb, line 70:
```suggestion
    replacement: "_",
```
I tend to think of an underscore replacing unwanted characters rather than dash...

Here's my example:
- title: My title for \\this && that / or such
- w/ dashes: `My title for -this && that - or such-FQDE`
- w/ underscores: `My title for _this && that _ or such-FQDE`

extensions/vscode/src/utils/names.ts
```diff
@@ -82,3 +52,45 @@ export function contentRecordNameValidator(
...
+    maxLength: 95,
+  });
+
+  // Generate unique name endings until we find a unique one
+  let result;
+  do {
+    const uniqueEnding = randomNameEnding(4);
+    result = `${filename}-${uniqueEnding}`;
```
sagerb, line 78:
```suggestion
    result = `${filename}:${uniqueEnding}`;
```
Another nit... a colon rather than a dash

sagerb (approved):
Code looks great, just had a few minor nits regarding the replacement character and the separation character. I'm not sure __why__ I have the preference, but I wanted to float the idea... I'll leave it up to you for your consideration.

extensions/vscode/src/utils/names.ts
dotNomad: We used `-` previously before the numbers which I why I used the hyphen here. I could go either way it all depends on if there is a standard for me - like with Python module naming using underscores.

I'll bring it up in standup for a quick check how others are feeling about it.

extensions/vscode/src/utils/names.ts
dotNomad: This one feels weird to me. I almost never use colons in my file names personally. Not saying I'm against it just voicing my immediate reaction.

dotNomad:
After a short discussion with the team we decided on kebab case, and lowering the max length of the file.

PR #2183, opened 2024-08-26
Draft - catch exceptions for R and Python Package Retrieval

sagerb:
## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

To recreate exceptions which need to be eliminated (which are shown within the debug window of VSCode):
- open `sample-content` directory
- select a deployment from `shinyapp`
- if you don't have a renv.lock file, you should probably see the unhandled exceptions when you activate the extension
- if you do have one, then delete it and then activate the extension. 
You'll see the original exceptions or handled exceptions within the Debug Output window of your development VSCode window.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -19,7 +19,7 @@ import {
   window,
   workspace,
 } from "vscode";
-import { isAxiosError } from "axios";
+import { AxiosError } from "axios";
 import { Mutex } from "async-mutex";
 
 import {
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -570,36 +581,51 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
-            activeConfiguration.configurationName,
-            activeConfiguration.projectDir,
-          );
+          const apiRequest = api.packages
+            .getRPackages(
+              activeConfiguration.configurationName,
+              activeConfiguration.projectDir,
+            )
+            // proposed fix...
+            .catch((error: AxiosError) => {
```
sagerb, line 590: Comment this `.catch` out and you'll see the issue when deleting the `renv.lock` file.

dotNomad:
I did some digging into this and I isolated it to the `showProgress` call.

If I removed the `showProgress` call the errors never showed up. I think we may be able to avoid these by doing something with that promise we are passing.

I will look more into this.

Issue #2175, opened 2024-08-26
Windows: When a failure occurs, clicking "Edit Configuration" throws an error


kgartland-rstudio:
In Windows, if an error occurs I get a button to `Edit Configuration` but when I click it the following error is thrown and the configuration does not open:

> The editor could not be opened due to an unexpected error: Unable to resolve resource c:%5CUsers%5Ckevin%5Cpublisher-git%5Ctest%5Csample-content%5Caddfiles%5C.posit%5Cpublish%5Cconfiguration-1.toml

dotNomad:
Which "Edit Configuration" button exactly has this issue?

Does the Edit Configuration button above the selector (the pencil / pen icon) have the same issue?

kgartland-rstudio:
Oh sorry, the Edit Configuration button that appears in the error message in the lower-right corner. The edit configuration link in the webview does _NOT_ have the same issue.

It's this button:
![Screenshot 2024-08-26 at 1 39 21 PM](https://github.com/user-attachments/assets/84367461-8364-449b-a46c-29093a4861e3)

dotNomad:
@sagerb we may want to take a look at this one for Beta 6.

dotNomad:
`EvenEasierDeploy` does:

```
onEditConfiguration(home.selectedConfiguration!.configurationPath)
```

Which sends the `WebviewToHostMessageType.EDIT_CONFIGURATION` message that does

```
    await commands.executeCommand(
      "vscode.open",
      Uri.file(msg.content.configurationPath),
    );
```

So ultimately we are opening `Uri.file(home.selectedConfiguration.configurationPath)`

---

In `posit.publisher.homeView.edit.Configuration` we do

```
commands.executeCommand(
  "vscode.open",
  Uri.parse(config.configurationPath),
);
```

So maybe the answer is to switch from `Uri.parse` to `Uri.file`?

sagerb:
thanks @dotNomad for looking into this, I totally agree and will implement that specific change.

PR #2185, opened 2024-08-26
Switch to properly use `Uri.file()` rather than `Uri.parse()`.

sagerb:
## Intent

Corrects usage to proper function for file paths. 

Fixes: #2175 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

From the VSCode documentation: 
```
		 * The *difference* between {@link Uri.parse} and {@link Uri.file} is that the latter treats the argument
		 * as path, not as stringified-uri. E.g. `Uri.file(path)` is *not* the same as
		 * `Uri.parse('file://' + path)` because the path might contain characters that are
		 * interpreted (# and ?). See the following sample:
		 * ```ts
		 * const good = URI.file('/coding/c#/project1');
		 * good.scheme === 'file';
		 * good.path === '/coding/c#/project1';
		 * good.fragment === '';
		 *
		 * const bad = URI.parse('file://' + '/coding/c#/project1');
		 * bad.scheme === 'file';
		 * bad.path === '/coding/c'; // path is now broken
		 * bad.fragment === '/project1';
```

Based on this new learning, I searched the codebase and determined that our file utility was making the same mistake, so I've fixed it in here.

## Directions for Reviewers

The conditions around the referenced bug should be verified.
The other two methods that were updated with this change are currently unused, but it is desirable to keep them in the codebase.

## Checklist

### DIFF
--- a/extensions/vscode/src/utils/files.ts
+++ b/extensions/vscode/src/utils/files.ts
@@ -79,7 +79,7 @@ export async function openNewOrExistingFileInPreview(
   appendedContents?: string,
 ) {
   let fileExist = true;
-  const existingUri = Uri.parse(filePath);
+  const existingUri = Uri.file(filePath);
   const newUri = Uri.file(filePath).with({ scheme: "untitled" });
 
   try {
@@ -123,7 +123,7 @@ export async function updateNewOrExistingFile(
   openEditor = false,
 ) {
   let fileExist = true;
-  const uri = Uri.parse(filePath);
+  const uri = Uri.file(filePath);
 
   try {
     await workspace.fs.stat(uri);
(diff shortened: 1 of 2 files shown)


dotNomad (approved):
This looks great! Tested all of the functionality to ensure it still worked:

- Deploying with an error (like using Python `version = '3.1200.5`) and clicking "Edit Configuration"
- Selection/list of "Open Files" when adding a New Deployment works

@kgartland-rstudio would you mind taking a look to see if this fixes the issue in Windows?

PR #2152, opened 2024-08-22
Escape errors rendered by InternalError

mmarchetti:
## Intent

PR #2095 added the error text to 500 responses from the backend to provide more informative error messaging. Error messages may contain contents from the request, such as the request URL. This introduces a potential XSS vulnerability, as reported by Snyk.

This PR sanitizes the error messages.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

PR #2189, opened 2024-08-27
Decode 500 error's data string

### LINKED ISSUE #2176: HTML encoded quotes are displayed in the error text when Config file is invalid
When creating an invalid config, the quotes are displaying as HTML encoded in the error messages:

config-type-error

unknown-key

sagerb:
## Intent

Resolves #2176 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Brought in an HTML decoder which would decode the string that the agent encoded using the go package `html` (escapeString).

Performed this decoding for all responses that fail with a 500 error code, by updating the Axios client response interceptor. Only doing this for the response data field, as that is the only place it was applied on the agent within #2152.

ALSO determined that we were not outputting durations for error responses. I've gone ahead and included that change within this PR as well.

## Directions for Reviewers

Easiest way to reproduce a 500 error is to invalidate a configuration file. For my testing, I changed the `[python]` section to be `[python2]`. Once you have done this, the error displayed in the lower-right corner error window should show the error without the quotes and other special characters being escaped out.

## Checklist

### DIFF
--- a/extensions/vscode/package-lock.json
+++ b/extensions/vscode/package-lock.json
@@ -14,6 +14,7 @@
         "async-mutex": "^0.5.0",
         "axios": "^1.7.4",
         "debounce": "^2.1.0",
+        "entities": "^5.0.0",
         "eventsource": "^2.0.2",
         "get-port": "5.1.1",
         "mutexify": "^1.4.0",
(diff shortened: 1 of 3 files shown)


dotNomad (approved):
This looks very clean! 😎

PR #2194, opened 2024-08-27
Passthrough callbacks with new `showProgressPassthrough`

dotNomad:
This PR ensures that errors generated from fetching R or Python packages are caught to avoid creating uncaught exception logs in the debug console.

I plan to do a follow-up PR to merge `showProgressPassthrough` and `showProgress` and refactor other usages.

This should supersede #2183

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

The approach was to pass the awaited function as the callback for `window.withProgress`, and use the fact that `window.withProgress` returns the "thenable the task-callback returned.".

This avoids creating a new callback and thus the uncaught exception error.

## Directions for Reviewers

To recreate the errors checkout `main` and open up the "Debug Console" while running VS Code in Extension Development Host mode.

- Create a Deployment for `shinyapp` in our `sample-content`
- Remove the `renv.lock` file if you have one,
- Once you don't have a `renv.lock` refresh the home view and you will see the unhandled exceptions

This PR should remove those errors.

### DIFF
--- a/extensions/vscode/src/utils/progress.ts
+++ b/extensions/vscode/src/utils/progress.ts
@@ -24,3 +24,27 @@ export async function showProgress(
     console.log(`Progress for "${title}" was displayed for ${duration}ms`);
   }
 }
+
+export async function showProgressPassthrough<T>(
+  title: string,
+  viewId: string,
+  until: () => Promise<T>,
+  trace = true,
+): Promise<T> {
+  const start = performance.now();
+
+  try {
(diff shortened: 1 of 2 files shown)


sagerb (approved):
Changes look good. Thanks!

PR #2195, opened 2024-08-28
Refactor `showProgress` to passthrough callback return

dotNomad:
Follow-up to #2194 that refactors all `showProgress` usage to use the new passthrough callback return method to ensure that errors are caught.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling


sagerb (approved):
I'm happy we arrived at a straightforward resolution for this. Thanks for accepting the handoff and figuring out what was going on.

The code changes look really good. 

General functionality is confirmed.

I am sorry that we lost the tracing event here for the show progress display. I found that the logging was extremely helpful in determining if we were properly wrapping our API calls or not. I'd like to figure out a way to bring that functionality back in, if at all possible (but this can certainly be done within a different PR).

PR #2198, opened 2024-08-28
update tests to get the new config name since they are no longer static

kgartland-rstudio:
Tests started failing last night because they look for the config name and check its contents. This PR uses a regex to find the filename opened in the editor and then checks the contents of that file.  

Tests are passing against this branch:

https://github.com/posit-dev/publisher/actions/runs/10599971513/job/29376369061

fixes: https://github.com/posit-dev/publisher/issues/2197

### DIFF
--- a/test/vscode-ui/test/specs/nested-fastapi.spec.ts
+++ b/test/vscode-ui/test/specs/nested-fastapi.spec.ts
@@ -5,11 +5,7 @@ import * as fs from "fs";
 import * as path from "path";
 import { fileURLToPath } from "url";
 import { dirname } from "path";
-import {
-  runShellScript,
-  switchToSubframe,
-  waitForInputFields,
-} from "../helpers.ts";
+import * as helper from "../helpers.ts";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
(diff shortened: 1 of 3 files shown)


test/vscode-ui/test/specs/nested-fastapi.spec.ts
```diff
@@ -157,16 +159,21 @@ describe("Nested Fast API Deployment", () => {
...
-    await expect(
-      await workbench.getEditorView().getOpenEditorTitles(),
-    ).toContain("configuration-1.toml");
+    const openEditorTitles = await workbench
+      .getEditorView()
+      .getOpenEditorTitles();
+    const fileNamePattern = /^my fastapi app-.*$/;
+    const realFilename = openEditorTitles.find((title) =>
+      fileNamePattern.test(title),
+    );
```
dotNomad, line 168: Should we pull this out into a testing utility function to avoid the duplicate code?

dotNomad (approved):
Looks great - glad tests are passing.

Mentioned one improvement that could be made (optional), but can we update the title to make it a bit easier to understand what happened in `git log`

kgartland-rstudio:
> Looks great - glad tests are passing.
> 
> Mentioned one improvement that could be made (optional), but can we update the title to make it a bit easier to understand what happened in `git log`

Good call, created a function for it in my helper.

PR #2210, opened 2024-08-28
Suppress permission errors

### LINKED ISSUE #2135: Opening a directory with no detectable entrypoints throws confusing errors
When I open a directory with no detectable entrypoints, I'm able to click the `Select...` dropdown.  Three errors are thrown and the Create New Deployment dropdown still displays.  If I then click `Create New Deployment` then I get a fourth error saying an entrypoint cannot be detected.

Error 1:
 ```
An error has occurred at refreshContentRecordData::contentRecords.getAll, StatusText=Internal Server Error, Code=ERR_BAD_RESPONSE, Msg=open /Users/kgartland/.Trash: operation not permitted, URL=http://localhost:57521/api/get//deployments
```

Error 2:
```
An error has occurred at refreshAll::Promise.all, StatusText=Internal Server Error, Code=ERR_BAD_RESPONSE, Msg=open /Users/kgartland/.Trash: operation not permitted, URL=http://localhost:57521/api/get//deployments
```

Error 3: 
```
An error has occurred at refreshConfigurationData::configurations.getAll, StatusText=Internal Server Error, Code=ERR_BAD_RESPONSE, Msg=open /Users/kgartland/.Trash: operation not permitted, URL=http://localhost:57521/api/get//configurations
```

Error 4 (after clicking Create New Deployment):
```
Unable to continue with project entrypoint detection failure. An error has occurred at newDeployment, entrypoints.get, StatusText=Internal Server Error, Code=ERR_BAD_RESPONSE, Msg=open /Users/kgartland/.Trash: operation not permitted, URL=http://localhost:57521/api/post//entrypoints
```

Instead can we catch this earlier when the Publisher is displayed, throw Error 4 and make the `Select...` dropdown inactive?

### PR DESCRIPTION
**Pull request by mmarchetti**
## Intent

This PR adds handling for permission errors during directory traversal.

Fixes #2135 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The `MatchingWalker` and `filesService` suppress permission errors.

## Directions for Reviewers

Deny read permissions to files or directories in your project (`chmod 000 $PATH`).

* Directories that can't be read will show up as empty in the Project Files view. 
* Files that can't be read will show up in the Project Files view. If you exclude them, you can deploy without error. If you include them, an error will be raised during deployment.

This is similar to the VSCode file tree behavior.

## Checklist

### DIFF
--- a/extensions/vscode/src/api/types/files.ts
+++ b/extensions/vscode/src/api/types/files.ts
@@ -25,6 +25,7 @@ export type ContentRecordFile = {
 export enum FileMatchSource {
   FILE = "file",
   BUILT_IN = "built-in",
+  PERMISSIONS_ERROR = "permissions",
 }
 
 export type FileMatch = {
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
@@ -4,7 +4,10 @@
     :key="file.id"
     :title="file.base"
     :checked="isFileIncluded(file)"
-    :disabled="file.reason?.source === 'built-in'"
+    :disabled="
+      file.reason?.source === 'built-in' ||
+      file.reason?.source === 'permissions'
+    "
     :list-style="isFileIncluded(file) ? 'default' : 'deemphasized'"
     :tooltip="
       isFileIncluded(file)
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
@@ -16,7 +16,7 @@
     @check="includeFile(file)"
     @uncheck="excludeFile(file)"
   >
-    <template v-if="file.files.length" #default="{ indentLevel }">
+    <template v-if="file.isDir" #default="{ indentLevel }">
       <TreeProjectFiles :files="file.files" :indentLevel="indentLevel" />
     </template>
 
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/tooltips.ts
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/tooltips.ts
@@ -13,6 +13,8 @@ export function excludedFileTooltip(file: ContentRecordFile) {
   if (file.reason) {
     if (file.reason.source === FileMatchSource.BUILT_IN) {
       tooltip += `\nThis is a built-in exclusion for the pattern: '${file.reason.pattern}' and cannot be overridden.`;
+    } else if (file.reason.source === FileMatchSource.PERMISSIONS_ERROR) {
+      tooltip += "\nYou don't have permission to access this directory.";
     } else {
       tooltip += `\nThe configuration file ${file.reason?.fileName} is excluding it with the pattern '${file.reason?.pattern}'`;
     }
--- a/internal/bundles/matcher/match_list.go
+++ b/internal/bundles/matcher/match_list.go
@@ -1,9 +1,6 @@
 package matcher
 
 import (
-	"os"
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

**Comment by sagerb** [comment]
I can confirm that deployment does stop with the error as expected. 

2024-08-28 at 3 24 PM

But the files view is not showing a directory w/ permissions issues as a directory (it doesn't have a chevron).

I created the following directory and files within my project:
```
~/dev/connect-content/bundles/python-shiny-express main > mkdir no-rights
~/dev/connect-content/bundles/python-shiny-express main > mkdir rights
~/dev/connect-content/bundles/python-shiny-express main > cd rights
~/dev/connect-content/bundles/python-shiny-express/rights main > touch no-touchy.txt
~/dev/connect-content/bundles/python-shiny-express/rights main > chmod 000 no-touchy.txt 
~/dev/connect-content/bundles/python-shiny-express/rights main > cd ..
~/dev/connect-content/bundles/python-shiny-express main > chmod 000 no-rights
```

And our file view showed:
2024-08-28 at 3 15 PM

In using this, I do think it should work slightly different:

My observations:
1. I would think that the directory should not be selectable and that the tooltip shows the reason why.
2. I would expect that we should disable the files which we don't have permission for and give a reason that it cannot be included within the tooltip.

sagerb (changes requested):
code looks great to handle permission error, but I think we want a change in the UX and I'm not sure if that will require a change in the agent or not. See my comment that describes my verification experience.

If we want to address those changes in a separate PR, let me know and I will approve this one.

sagerb:
It does also appear that you are fairly behind on `main`... so I pushed a merge update into it for you. I noticed that the file inclusion/exclusion wasn't inserting the forward slash into the files array... and was worried we somehow broke that. But after the merge, it worked as expected.

mmarchetti:
Thank you for reviewing. The directory showing up without the chevron is definitely a bug, I'll take a look. Having the file show up with an exclusion reason will be trickier, because we aren't opening files when scanning - we only `stat` them, so we don't know whether we could successfully open them.

sagerb:
> Thank you for reviewing. The directory showing up without the chevron is definitely a bug, I'll take a look. Having the file show up with an exclusion reason will be trickier, because we aren't opening files when scanning - we only `stat` them, so we don't know whether we could successfully open them.

But having that permission error, isn't that enough for us to exclude them, that they won't go up? I'm just thinking we don't allow selection with a permission error (although I guess that is also confusing if the user has the file already listed in the files array).

mmarchetti:
We don't detect permission errors for files in the Files API (only directories). We open/read directories in order to traverse them, which is when the error becomes visible. We don't open each file, because it would be very slow for large directories, so we don't get permission errors back for those.

We can handle permission errors during bundle creation by skipping affected files and logging. As you mentioned, this might be confusing if you've selected a file for inclusion and it can't be read, but files in your workspace that you can't read should be pretty rare. And it's likely that you might have tried to open the file in VSCode at some point, which would pop up an error.

sagerb:
> the file in VSCode at some point, which would pop up an error...

Ah, ok, thanks for the clarification. Then I'm good with not worrying about the exclusion reasons for files but since a directory with permission errors is empty, we might want to think about (in a future PR) passing an exclusion reason for that (if you haven't already implemented that).

Let me re-verify your latest change.

sagerb (approved):
With the recent changes, the expected functionality works great. Directory w/ permission errors is showing as an empty directory, and as seen below, deployment detects/shows a permission error when deploying.

2024-08-29 at 9 06 AM

Even better if you see the posit publisher log!

2024-08-29 at 9 09 AM

dotNomad (approved):
Fantastic changes. Frontend looks great as well.

kgartland-rstudio:
Verified errors in the original issue https://github.com/posit-dev/publisher/issues/2135 no longer occur.

PR #2213, opened 2024-08-29
Preprend slash in config files for all types

### LINKED ISSUE #2211: shiny-express files array missing directory
When creating a config for shiny-express the files array gets: `app.py` rather than `/app.py` like other entrypoint files.

Should that be updated?

dotNomad:
This PR ensures that we prepend the `/` to avoid accidentaly including files with the same name as the entrypoint in subdirectories for the configuration files we generate.

This is a follow-up to #2174 that applies this to all types of Deployments - HTML, Quarto, Shiny Express, etc

## Intent

Resolves #2211 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/internal/initialize/initialize_test.go
+++ b/internal/initialize/initialize_test.go
@@ -175,7 +175,7 @@ func (s *InitializeSuite) TestGetPossibleConfigs() {
 
 	s.Equal(config.ContentTypeHTML, configs[1].Type)
 	s.Equal("index.html", configs[1].Entrypoint)
-	s.Equal([]string{"index.html"}, configs[1].Files)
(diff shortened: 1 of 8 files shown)


kgartland-rstudio (approved):
verified this is now prepending a / for shiny-express and quarto projects now.

sagerb (approved):
Changes look good. Kevin has already verified.

sagerb:
I pushed up changes to fix the unit tests.

PR #2221, opened 2024-08-29
Only show added decorator when file is new and included

### LINKED ISSUE #2218: Added files decorator should only apply to included files
Currently the Project Files tree view shows the `A` for added files whenever there is a file that was not included in the last deployment despite it being excluded.

dotNomad:
This PR fixes the added decorator (the green A after files in the Project Files view) to only show if the file is new (not included in the last deployment) and included.

Previously any file that was new was being decorated.

## Intent

Resolves #2218

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
@@ -24,7 +24,7 @@
       #postDecor
       v-if="
         file.isFile &&
-        file.reason?.source !== FileMatchSource.BUILT_IN &&
+        isFileIncluded(file) &&
         !home.flatFiles.lastDeployedFiles.has(file.rel)
       "
     >


extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
```diff
@@ -24,7 +24,7 @@
       #postDecor
       v-if="
         file.isFile &&
-        file.reason?.source !== FileMatchSource.BUILT_IN &&
```
dotNomad, line 27: If the file is included it isn't `FileMatchSource.BUILT_IN` so we can reduce the if here.

sagerb (approved):
Hardly can say much about this change other than it looks great!

PR #2235, opened 2024-08-30
Show `indeterminate` state in Project Files view based on inclusions

### LINKED ISSUE #2219: Use `indeterminate` state in Project Files with new API attributes
#2163 gave us new attributes to do `indeterminate` states in the Project Files tree view. This will greatly improve readability for directories in that view.

dotNomad:
This PR adds an `indeterminate` state to the checkboxes in the tree Project Files view based on inclusion and the new `allIncluded` and `allExcluded` attributes.

## Intent

Resolves #2219

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

You can see the different states by:
- including a directory, but not one of its children
- excluding a directory, but including one of its children

Keep in mind that the include/exclude files API doesn't handle directories perfectly. I'd recommend looking closely at the files attribute in the Configuration while testing this feature to ensure erroneous includes are being added.

### DIFF
--- a/extensions/vscode/src/api/types/files.ts
+++ b/extensions/vscode/src/api/types/files.ts
@@ -20,6 +20,8 @@ export type ContentRecordFile = {
   size: number;
   fileCount: number;
   abs: string;
+  allIncluded: boolean;
+  allExcluded: boolean;
 };
 
 export enum FileMatchSource {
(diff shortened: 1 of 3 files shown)


dotNomad:
There are some open questions in #2219 so this is left in draft.

dotNomad:
Closing this one up. If we pick-up #2219 we can use this as a starting point.

PR #2240, opened 2024-08-30
Use `file.id` comparison to match content record files list

### LINKED ISSUE #2112: Windows: After a deployment the `A` in Project Files remain for all subdirectory files
After a deployment in Windows, the `A` indicating a new file being added to a Deployment is removed for all files except for those in a subdirectory so it appears all files in a subdirectory are new for subsequent deployments.

This only happens in Windows.

Screenshot 2024-08-14 103221

dotNomad:
This PR fixes the windows path comparisons between the Content Record's `file` list and the files we get from the files API.

Previously I was using `file.rel` which was doing comparisons like:
```
'subdir\\subsubdir\\newtext.txt' == 'subdir/subsubdir/newtext.txt'
```

Instead of 
```
'subdir/subsubdir/newtext.txt' == 'subdir/subsubdir/newtext.txt'
```

`file.rel` in our files API gives the relative path with the OS's separators where `file.id` gives the same string that we expect in our content records and configuration files.

With that corrected this should address the issue of decorators showing up or not showing up correctly on Windows.

## Intent

Resolves #2112

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
@@ -28,7 +28,7 @@
         v-if="
           file.isFile &&
           isFileIncluded(file) &&
-          !home.flatFiles.lastDeployedFiles.has(file.rel)
+          !home.flatFiles.lastDeployedFiles.has(file.id)
         "
         class="text-git-added"
       >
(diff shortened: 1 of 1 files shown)


kgartland-rstudio (approved):
Looks good!

Issue #2153, opened 2024-08-22
Add more debug logging in the backend


mmarchetti:

dotNomad:
We have added additional debug logging, and will be adding more as we work on other issues.

PR #2242, opened 2024-09-04
Adding some more debug lines to inspection and deployment

marcosnav:
## Intent

Adding some more debug logs to the inspection and deployment areas of the backend.

Connected to #2153 

## Approach

- Some new `log.Debug` calls within the inspection and deployment.
- Adding a logger instance to `defaultPublisher` to do logging calls easier across this instance methods.

## Automated Tests

Updated a few unit tests related to expected logging instances.

## Directions for Reviewers

While inspecting and deploying, there should be more debug log lines in the output.

## Checklist

### DIFF
--- a/internal/services/api/post_deployment.go
+++ b/internal/services/api/post_deployment.go
@@ -78,6 +78,7 @@ func PostDeploymentHandlerFunc(
 		w.WriteHeader(http.StatusAccepted)
 		json.NewEncoder(w).Encode(response)
 
+		log := log.WithArgs("local_id", localID)
 		newState.LocalID = localID
 		publisher, err := publisherFactory(newState, emitter, log)
 		log.Debug("New publisher derived from state", "account", b.AccountName, "config", b.ConfigName)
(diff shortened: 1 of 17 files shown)


mmarchetti (approved):

internal/services/api/post_deployment.go
```diff
@@ -76,16 +78,17 @@ func PostDeploymentHandlerFunc(
...
 		newState.LocalID = localID
 		publisher, err := publisherFactory(newState, emitter, log)
+		log.Debug("New publisher derived from state", "account", b.AccountName, "config", b.ConfigName)
 		if err != nil {
 			InternalError(w, req, log, err)
 			return
 		}
 
 		go func() {
-			log := log.WithArgs("local_id", localID)
```
sagerb, line 87: By removing this, are we losing functionality? How will we see the localID, which allows us to discriminate between logs of multiple streams?

sagerb (commented):
Changes look good, just had one question where I think we're losing something.

internal/services/api/post_deployment.go
marcosnav: It is not being removed, it was moved to line 81 above so the derived logger with the `local_id` can be used earlier and passed down to the `publisherFactory`

PR #2262, opened 2024-09-05
Displaying a friendlier error for unknown fields found in configuration files

### LINKED ISSUE #2157: Handle error displayed (500) when configuration is updated and becomes invalid
Edit the active configuration and add an invalid entry (such as `abc = 123`). Save the file.

Notice the error displayed in the notification:

Image

It seems like we can do better and not refresh files based on a Configuration Error. Strange we're not seeing multiple errors actually, since we also refresh packages based on the config.

marcosnav:
## Intent

In order to have more standardized errors handling, this PR introduces error types - to be expanded as we add more "known" errors.

**Screenshot**
Screen Shot 2024-09-17 at 10 56 53

Fixes #2157  

## Type of Change

Improvement

## Approach

On the vscode extension UI code, changes made to `getSummaryStringFromError` allow to get a human consumable error message when the error contains a known JSON agent response - otherwise, a tracing-style message is returned to help diagnose.

## Automated Tests

- Updated backend tests as needed.
- Added new UI tests for the new code. (Run locally and will add something to test on CI too)

## Directions for Reviewers

- Edit the active configuration and add an invalid entry (such as `abc = 123`).
- Save the file.
- Notice the error displayed in the notification.

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
-*   @dotNomad @kgartland-rstudio @mmarchetti @sagerb
+*   @dotNomad @kgartland-rstudio @sagerb @marcosnav
 /internal/ @tdstein
--- a/extensions/vscode/src/utils/errorTypes.ts
+++ b/extensions/vscode/src/utils/errorTypes.ts
@@ -1,0 +1,124 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+import { AxiosError, AxiosResponse, isAxiosError } from "axios";
+
+type ErrorCode =
+  | "unknown"
+  | "resourceNotFound"
+  | "invalidTOML"
+  | "unknownTOMLKey"
+  | "invalidConfigFile";
+
+export type axiosErrorWithJson<T = { code: ErrorCode; details: unknown }> =
+  AxiosError & {
+    response: AxiosResponse<T>;
+  };
+
+export const isAxiosErrorWithJson = (
+  err: unknown,
+): err is axiosErrorWithJson => {
+  if (isAxiosError(err)) {
+    return err.response?.data && err.response.data.code;
+  }
+  return false;
+};
+
+const isOfErrorType = (err: axiosErrorWithJson, code: ErrorCode): boolean => {
+  return err.response.data.code === code;
+};
+
+type MkErrorDataType<T extends ErrorCode, D = Record<PropertyKey, never>> = {
+  code: T;
+  details: D;
+};
+
+const mkErrorTypeGuard = <T>(code: ErrorCode) => {
+  return (err: unknown): err is axiosErrorWithJson<T> => {
+    return isAxiosErrorWithJson(err) && isOfErrorType(err, code);
+  };
+};
+
+// Unknown Errors
+export type ErrUnknown = MkErrorDataType<
+  "unknown",
+  {
+    error: string;
+    data: Record<PropertyKey, any>;
+  }
+>;
+export const isErrUnknown = mkErrorTypeGuard<ErrUnknown>("unknown");
+export const errUnknownMessage = (err: axiosErrorWithJson<ErrUnknown>) => {
+  const { error, data } = err.response.data.details;
+  let msg = `Unknown publisher agent error: ${error}`;
+  Object.keys(data).forEach((key) => {
+    msg += `, ${key}=${data[key]}`;
+  });
+  return msg;
+};
+
+// Resource not found
+export type ErrResourceNotFound = MkErrorDataType<
+  "resourceNotFound",
+  { resource: string }
+>;
+export const isErrResourceNotFound =
+  mkErrorTypeGuard<ErrResourceNotFound>("resourceNotFound");
+
+// Invalid TOML file(s)
+export type ErrInvalidTOMLFile = MkErrorDataType<
+  "invalidTOML",
+  {
+    filename: string;
+    line: number;
+    column: number;
+  }
+>;
+export const isErrInvalidTOMLFile =
+  mkErrorTypeGuard<ErrInvalidTOMLFile>("invalidTOML");
+export const errInvalidTOMLMessage = (
+  err: axiosErrorWithJson<ErrInvalidTOMLFile>,
+) => {
+  const { filename, line, column } = err.response.data.details;
+  return `Invalid TOML file ${filename}:${line}:${column}`;
+};
+
+// Unknown key within a TOML file
+export type ErrUnknownTOMLKey = MkErrorDataType<
+  "unknownTOMLKey",
+  {
+    filename: string;
+    line: number;
+    column: number;
+    key: string;
+  }
+>;
+export const isErrUnknownTOMLKey =
+  mkErrorTypeGuard<ErrUnknownTOMLKey>("unknownTOMLKey");
+export const errUnknownTOMLKeyMessage = (
+  err: axiosErrorWithJson<ErrUnknownTOMLKey>,
+) => {
+  const { filename, line, column, key } = err.response.data.details;
+  return `Unknown field present in configuration file ${filename}:${line}:${column} - unknown key "${key}"`;
+};
+
+// Invalid configuration file(s)
+export type ErrInvalidConfigFiles = MkErrorDataType<
+  "invalidConfigFile",
+  { filename: string }
+>;
+export const isErrInvalidConfigFile =
+  mkErrorTypeGuard<ErrInvalidConfigFiles>("invalidConfigFile");
+
+// Tries to match an Axios error that comes with an identifiable Json structured data
+// defaulting to be ErrUnknown message when
+export function resolveAgentJsonErrorMsg(err: axiosErrorWithJson) {
+  if (isErrUnknownTOMLKey(err)) {
+    return errUnknownTOMLKeyMessage(err);
+  }
+
+  if (isErrInvalidTOMLFile(err)) {
+    return errInvalidTOMLMessage(err);
+  }
+
+  return errUnknownMessage(err as axiosErrorWithJson<ErrUnknown>);
+}
--- a/extensions/vscode/src/utils/errors.ts
+++ b/extensions/vscode/src/utils/errors.ts
@@ -1,27 +1,10 @@
 // Copyright (C) 2023 by Posit Software, PBC.
 
-import axios, { AxiosError, AxiosResponse } from "axios";
+import axios from "axios";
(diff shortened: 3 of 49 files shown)


.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
-*   @dotNomad @kgartland-rstudio @mmarchetti @sagerb
+*   @dotNomad @kgartland-rstudio @sagerb @marcosnav
```
marcosnav, line 1: @mmarchetti 😢

extensions/vscode/src/views/homeView.ts
```diff
@@ -1379,10 +1382,15 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
-        const summary = getSummaryStringFromError(
-          "sendRefreshedFilesLists, files.getByConfiguration",
-          error,
-        );
+        let summary = "";
+        if (isAxiosJsonErrorRes(error)) {
+          summary = error.response.data.message;
+        } else {
+          summary = getSummaryStringFromError(
+            "sendRefreshedFilesLists, files.getByConfiguration",
```
sagerb, line 1390: These strings were added so that we could quickly identify where the code was failing. We should figure out a way to keep them, even if it is just to be used in log messages, rather than the display.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1379,10 +1382,15 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           },
         });
       } catch (error: unknown) {
-        const summary = getSummaryStringFromError(
-          "sendRefreshedFilesLists, files.getByConfiguration",
-          error,
-        );
+        let summary = "";
+        if (isAxiosJsonErrorRes(error)) {
```
sagerb, line 1386: This is where the suggestion above would change the code, to not require the if statement.

internal/services/api/api_helpers.go
```diff
@@ -12,6 +13,20 @@ import (
 	"github.com/posit-dev/publisher/internal/util"
 )
 
+type jsonResponseError struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+	Error   string `json:"error"`
+}
+
+func NewJsonResponseError(code int, message string, err error) jsonResponseError {
```
sagerb, line 22: Have you considered adding another error code to the mix, one that would be specific to our project and not have to be used as the REST API error code? That way we could know that a malformed configuration file with extra files is error `abc`, and we can key off of that very specifically in the UX for context-specific treatment.

internal/services/api/get_config_files.go
```diff
@@ -24,10 +35,13 @@ func GetConfigFilesHandlerFunc(base util.AbsolutePath, filesService files.FilesS
 			// Response already returned by ProjectDirFromRequest
 			return
 		}
-		configPath := config.GetConfigPath(projectDir, name)
-		cfg, err := config.FromFile(configPath)
+		configPath := configGetConfigPath(projectDir, name)
+		cfg, err := configFromFile(configPath)
 		if err != nil {
-			if errors.Is(err, fs.ErrNotExist) {
+			if aerr, ok := err.(*types.AgentError); ok && aerr.Code == util.UnknownTOMLKeyCode {
```
sagerb, line 41: Previous comment example here, the error code passed back to the client would be something like `util.UnknownTOMLKeyCode` within the config file.

internal/util/toml.go
```diff
@@ -56,8 +56,8 @@ func decodeErrFromTOMLErr(e *toml.DecodeError, path AbsolutePath) *DecodeError {
 	}
 }
 
-const invalidTOMLCode types.ErrorCode = "invalidTOML"
-const unknownTOMLKeyCode types.ErrorCode = "unknownTOMLKey"
+const InvalidTOMLCode types.ErrorCode = "invalidTOML"
+const UnknownTOMLKeyCode types.ErrorCode = "unknownTOMLKey"
```
sagerb, line 60: These are the errors which definition should be shared between client and agent.

sagerb (changes requested):
Had a few questions which might take these changes in the direction that the team had previously discussed. (Sorry, prior to your time.)

See what you think about my comments and let me know what questions you might have.

internal/services/api/api_helpers.go
mmarchetti: I would like to see us standardize the HTTP statuses so there are only 1-2 unique error statuses, and instead, differentiate errors by an error code field in the returned JSON.

marcosnav: So, if I understand correctly, we are looking to implement something like what we have in Connect?
```go
type ErrorCode int

// Error codes for the API. These values need to remain stable, as clients may
// use them to drive error-specific behavior.
const (
    _                       = ErrorCode(iota)
    ErrorInvalidTOMLCode
    ErrorUnknownTOMLKeyCode
    ...
)

var errorMessageEntries = map[ErrorCode]struct {
    Code    int
    Message string
}{
    ErrorInvalidTOMLCode:    {http.StatusBadRequest, "This is an invalid toml."},
    ErrorUnknownTOMLKeyCode: {http.StatusBadRequest, "Unknown field present in configuration file."},
    ...
}

type jsonResponseError struct {
	ErrorCode   ErrorCode  `json:"error_code"`
	Message     string     `json:"message"`
	Error       string     `json:"error"`
}
```

mmarchetti: Yes, similar to that. Publisher currently has some error codes defined, but they are strings, so we might want a string value in the JSON.

marcosnav:
@sagerb @mmarchetti Thanks for the feedback 🙇  I'll move some things around and re-request review when it is ready for a second look

marcosnav:
@sagerb @mmarchetti This PR is ready for another look 🙂

extensions/vscode/src/utils/errors.ts
```diff
@@ -1,10 +1,28 @@
...
+import axios, { AxiosError, AxiosResponse } from "axios";
 
 export type ErrorMessage = string[];
 export type ErrorMessages = ErrorMessage[];
 
+export type ErrorResData = {
+  code: string;
+  status: number;
+  msg: string;
+  error: string;
```
sagerb, line 12: Am I correct that you've added an entire structure just so that you can add an error code/string? We already had the code, status and a message within the error response; did you consider simply changing the `error.response.data.msg` from a string to a possible error msg object w/ error string and error string's code?

I think this would have less of an impact on the changes and what you need to do when we have an error string code vs. not. I'm thinking that the agent would change to simply have a new definedError interface which would implement error, plus the new identifying code.

Let me reach out and we can discuss.

sagerb (commented):
Going to reach out and move this to a verbal conversation. Then will update.

.github/CODEOWNERS
dotNomad: Should we pull this out into a separate PR, just to get the `CODEOWNERS` right as quickly as possible?

dotNomad: I created https://github.com/posit-dev/publisher/pull/2278 to get this in so @mmarchetti doesn't get as many non-targeted Publisher notifications.

marcosnav: Thank you!

sagerb (approved):
I hate leaving the technical debt on the table for the suggestion I asked about, but if its not a fast change for you, I'm OK with merging without the change (and creating an issue to do it after the release).

internal/services/api/get_config_files.go
```diff
@@ -13,9 +12,17 @@ import (
...
 	"github.com/posit-dev/publisher/internal/util"
 )
 
+type cfgFromFile func(path util.AbsolutePath) (*config.Config, error)
+type cfgGetConfigPath func(base util.AbsolutePath, configName string) util.AbsolutePath
+
+// TODO: It would be better to have the config package methods as a provider pattern instead of plain functions
+var configFromFile cfgFromFile = config.FromFile
+var configGetConfigPath cfgGetConfigPath = config.GetConfigPath
+
```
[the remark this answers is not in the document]
marcosnav: I don't like leaving TODO's like this but in this case the suggested change would have an impact on many other places using the `config` package. That would expand this PR changes and also delay it's landing.

extensions/vscode/src/utils/errorTypes.ts
dotNomad, line 1: Fantastic work on this file. Great exports, great helpers 💯

dotNomad (approved):
I just had two questions pop up.

extensions/vscode/tsconfig.json
```diff
@@ -18,5 +18,5 @@
       "src/*": ["./src/*"]
     }
   },
-  "exclude": ["node_modules", "webviews"]
+  "exclude": ["node_modules", "webviews", "**/**.test.ts", "vitest.config.ts"]
```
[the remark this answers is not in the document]
marcosnav: Ah that's right, test files won't enforce TS rules 🤔 . I'll try to figure out something, because `tsc` complains about them. When included, resolution need to be set as `"Node16"`, but some dependencies for the extension require resolution as it is now `"commonjs"`.

internal/services/api/api_errors.go
```diff
@@ -0,0 +1,72 @@
...
+		Details: InvalidTOMLFileDetails{
+			Filename: aerr.Data["file"].(string),
+			Line:     aerr.Data["line"].(int),
+			Column:   aerr.Data["column"].(int),
+		},
+	}
+}
+
+func (apierr *APIErrorInvalidTOMLFileDetails) JSONResponse(w http.ResponseWriter) {
+	w.Header().Set("content-type", "application/json")
```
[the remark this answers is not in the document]
marcosnav: Got it, yes this can be confusing. There is a small difference between them though, but it can be handled better.

marcosnav:
@dotNomad pushed a couple commits related to your questions. Things seem to be better now:
- test files evaluated by TS compiler
- Adjusted `JsonResult` to receive status and be used thoroughly

extensions/vscode/tsconfig.json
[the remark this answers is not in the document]
dotNomad: I don't mind adding an issue to the backlog / triage board as a follow-up so we can get this in!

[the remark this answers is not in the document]
dotNomad: Oh you fixed it already! Nice!

dotNomad (approved):
Thanks for those additional changes @marcosnav! 🎉

PR #2273, opened 2024-09-14
Add ability to send secrets to `POST /api/deployments/$NAME` endpoint

### LINKED ISSUE #2254: Secrets: Update `POST /api/deployments/$NAME` to accept secrets
The POST /api/deployments/$NAME endpoint initiates a deployment. It should accept a new `secrets` parameter to pass into the deployment process.

dotNomad:
This PR adds an optional bit to the `POST` body of our `/api/deployments/$name` endpoint.

```
    "secrets": {
        "SOME_SECRET_NAME": "SOME_VALUE"
    }
```

These relate to secret names defined in the Configuration being used to deploy. The API will error with a `400` if a secret is passed that is not included in the Configuration.

Secrets passed via the API are then used to set environment variables on the content in Connect.

## Intent

Resolves #2254 #2255

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I utilized `State` and the factory for it, expanding it to include `Secrets` when passed via the API. Those are then compared to the Configuration to check for invalid secrets. Finally, when deploying to Connect, the secrets from the `State` are added to the environment.

## Automated Tests

I included automated tests for each section where new code was added other than `set_env_vars.go` to ensure that the Connect API was called with the full env var and secrets information since there was no test infrastructure for that file.

## Directions for Reviewers

To test this you have to run the API service. I do this by running:

```
just run ui test/sample-content/fastapi-simple --listen=127.0.0.1:9001
```

then I used [httpie](https://httpie.io/) but you can use `curl` or whatever other tool you'd like to deploy a created deployment, config, and credential with secrets like so:

```
❮ http POST :9001/api/deployments/deployment-DEQ7 \
  account="Posit Demo Server" \
  config="fastapi-simple-QSPH" \
  secrets:='{"SOME_SECRET": "SOME_VALUE"}' -v
```

<details>
  <summary>With Output</summary>
  
```
❮ http POST :9001/api/deployments/deployment-DEQ7 account="Posit Demo Server" config="fastapi-simple-QSPH" secrets:='{"SOME_SECRET": "SOME_VALUE"}' -v
POST /api/deployments/deployment-DEQ7 HTTP/1.1
Accept: application/json, */*;q=0.5
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 107
Content-Type: application/json
Host: localhost:9001
User-Agent: HTTPie/3.2.3

{
    "account": "Posit Demo Server",
    "config": "fastapi-simple-QSPH",
    "secrets": {
        "SOME_SECRET": "SOME_VALUE"
    }
}


HTTP/1.1 202 Accepted
Content-Length: 31
Content-Type: application/json
Date: Sat, 14 Sep 2024 00:40:36 GMT
Vary: Origin

{
    "localId": "j2XHAI4GD0BKN_Ds"
}
```
</details> 

you should see on Connect, in the Vars settings panel for the Content that environment variables are set as normal, but so are any secrets defined and passed

Other cases:
- Deploying without specifying secrets (in the API call or the Configuration) should still work.
- Deploying with secrets declared in the Configuration should work even if no Secrets are passed in the API call


internal/state/state.go
```diff
@@ -143,6 +144,14 @@ func New(path util.AbsolutePath, accountName, configName, targetName string, sav
 			return nil, err
 		}
 	}
+
+	// Check that the secrets passed are in the config
+	for secret := range secrets {
+		if !cfg.HasSecret(secret) {
```
dotNomad, line 150: This double loops since `cfg.HasSecret` does a loop over the Secrets defined in the Configuration, but that felt acceptable since:
- these will most likely be relatively small
- iterating through everything gives us a good error showing the first secret not in the config

Based on feedback I can expanded this to include an error of _all_ secrets passed that were not in the configuration if we think that is helpful. I wrote this as fail fast first.

internal/config/types.go
```diff
@@ -104,6 +104,15 @@ type Config struct {
...
 }
 
+func (c *Config) HasSecret(secret string) bool {
+	for _, s := range c.Secrets {
+		if s == secret {
+			return true
+		}
+	}
+	return false
+}
```
marcosnav, line 114: It would be good to have tests for this new method, even if it is a fairly simple method, it's nice to build confidence for future changes.

internal/publish/set_env_vars.go
```diff
@@ -31,7 +34,17 @@ func (p *defaultPublisher) setEnvVars(
...
+	for name := range secrets {
+		log.Info("Setting secret as environment variable", "name", name)
+	}
+
+	// Combine env and secrets into one environment for Connect
+	combinedEnv := make(map[string]string)
+	maps.Copy(combinedEnv, env)
+	maps.Copy(combinedEnv, secrets)
+
+	err := client.SetEnvVars(contentID, combinedEnv, log)
```
marcosnav, line 47: same here, would be good to have tests for these changes

marcosnav (approved):
LGTM 👍 , a couple suggestions to have bit more testing coverage

internal/config/types.go
dotNomad: Added some unit tests around this in https://github.com/posit-dev/publisher/pull/2273/commits/6feea3a03a050a1547ed043b604bcba7438d7353

sagerb (approved):
Code looks good to me.

I verified the functionality, both good and bad paths. Looks great!

internal/publish/set_env_vars.go
dotNomad: After some pairing I will get something pushed up momentarily.

dotNomad: Added tests in https://github.com/posit-dev/publisher/pull/2273/commits/e085f8b84703f1376f32c7e3afe1918ba36ddcb0

PR #2274, opened 2024-09-16
webview tests

kgartland-rstudio:
## Intent

* Add testing for all webview components after configuration is created
* Added a script to help with flakiness. Connect installations would regularly fail due a linux package lock. i added a script to the fuzzbucket instance that checks for the lock and kills the process using the lock prior to installing Connect. I haven't seen Connect installation failures since adding that script.
* Upgraded tests and fuzzbucket instance to Python 3.12 to avoid a python-setup bug that doesn't set the python version in the shell properly so VSCode doesn't get the expected Python version that's specified in python-setup step.
* Added data-automation tags to the components under test
* Added parameterization in Github Actions so we can select `all`, `bats-tests`, or `vscode-ui-tests` in the Nightly runs. The `bats-tests` includes about 50 deployments so it's overkill to run both test suites if we're just working on the vscode ui. This is a manual override in the GH Actions UI, the scheduled tests will use `all` by default.

The tests currently rely on initial tests that create a config. I have plans on making the webview testing completely stand-alone but that will be on a follow-on PR. That will allow us to re-run a single test if something is failing for troubleshooting.

These tests are for green path only. I will be adding tests for failure scenarios (missing config, missing creds, etc) in future PR.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Checklist


extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
```diff
@@ -35,7 +35,9 @@
         :class="codicon"
       />
       <div class="tree-item-label-container">
-        <span class="tree-item-title">{{ title }}</span>
+        <span class="tree-item-title" :data-automation="`req`">{{
```
sagerb, line 38: I'm not understanding why you are evaluating an expression which just returns the string of "req"? Or do I have this wrong? I think it should be able to be replaced with:
```
<span class="tree-item-title" data-automation="req">{{
```

test/vscode-ui/test/helpers.ts
```diff
@@ -6,9 +6,9 @@ export async function switchToSubframe() {
   const iframe = await browser.$("iframe");
   await browser.switchToFrame(iframe);
 
-  await browser.$("iframe").waitForExist({ timeout: 3000 });
+  await browser.$("iframe").waitForExist({ timeout: 10000 });
```
sagerb, line 9: What was it that required a jump from 3 seconds to 10 seconds for the wait? (just interested, because that delay seems like something we would want to address.)

test/vscode-ui/test/specs/nested-fastapi-deployment.spec.ts
sagerb, line 1: Is your intention to eventually verify all of the tests within this file, against every content type?

test/vscode-ui/wdio.conf.ts
```diff
@@ -161,7 +170,7 @@ export const config: Options.Testrunner = {
     ui: "bdd",
     timeout: 60000,
   },
-
+  // wdio.conf.js
```
sagerb, line 173: Do you need this in here?

sagerb (approved):
Just a few questions, but it looks good! I'll be happy to see this put into place. :-)

extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
kgartland-rstudio: i think i had been using a merge field there but then realized i didn't need it but forgot to remove the "`".  Thanks, I'll fix that.

test/vscode-ui/test/helpers.ts
kgartland-rstudio: I think I increased that up to 10 secs for troubleshooting. i can probably bump that back down.

test/vscode-ui/test/specs/nested-fastapi-deployment.spec.ts
kgartland-rstudio: No, I plan on having separate test suites for each content type (at least 1 Python, R, Quarto content type). I'm not yet sure if I'll run all the tests on each PR or have some run nightly instead.

test/vscode-ui/wdio.conf.ts
kgartland-rstudio: i do not, I'll remove it.

PR #2286, opened 2024-09-18
Associate predeployment with existing content

### LINKED ISSUE #2285: Associate a Pre-Deployment with an existing Content Item on a server
This will serve as the simplest implementation of the import function we have discussed.

It should basically allow for a user to enter the URL that they use to access the content item they wish to update, and understand that it will update rather than create a new content item.

sagerb:
## Intent

This PR provides the functionality to associate a pre-deployment with an existing deployment on a connect server.

Resolves #2285 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

- Pre-Deployments now can have an ID as well as DashboardURL, DirectURL and LogsURL (since they are built off of the ID).
- The POST /deployment and PATCH /deployment APIs have been enhanced to allow an ID to be specified.
- The UX now displays some helpful text below the deployment status when a deployment has not yet been deployed. It offers the ability for those predeployments that do not have an ID to be associated with one, via an achor. The Anchor will initiate a prompt for the URL used to access content, from which the code extracts the GUID and then patches the deployment file. The UX then updates the text to reflect that deploying this project will update an existing one.

Demo:

https://github.com/user-attachments/assets/a23f4fc8-62e0-4e22-816c-91e619cba0dd

## Automated Tests

Unit tests updated

## Directions for Reviewers

I have broken down the updates into three commits and would recommend reviewing them in that order.

1. [Move URL factory methods out to separate utility to avoid circular dependency when needed](https://github.com/posit-dev/publisher/commit/20a3977dfd9b219042f50331f72fea3330dea289)
2. [Move ID, DashboardURL, DirectURL and LogsURL into pre-deployment file definitions.](https://github.com/posit-dev/publisher/commit/5e01b9500685fc014cee46b1e6c4300abd6fc7ff) 
3. [Support Patching the ID on a predeployment file and supplying ID on initial post deployments call.](https://github.com/posit-dev/publisher/commit/807a30026f5311e522a4c5e47579a13771343f98)
4. [Update UX to support patching ID to associate predeployment with existing content item.](https://github.com/posit-dev/publisher/commit/6a19a61b347c3ad92bb686ba7c70739b9a28ac5b)
5. [keep prompt window open so user can switch windows to copy/paste URL](https://github.com/posit-dev/publisher/pull/2286/commits/c4d0e41d52cb7862c78e6120a30f7e0573f8042c)
6. 
## Checklist


extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -129,6 +129,16 @@
...
+        <div v-if="isPreContentRecordWithoutID">
+          Deploying will create a new Content item on the Connect Server. Click
+          <a class="webview-link" role="button" @click="onAssociateDeployment"
+            >here</a
+          >
+          to associate it with an existing Content item on the server.
+        </div>
+        <div v-if="isPreContentRecordWithID">
+          Deploying will update an existing Content item on the Connect Server.
+        </div>
```
jonkeane, line 141: Trying to streamline this a bit. The code/links might not line up exactly from the suggestion part itself

```suggestion
        <div v-if="isPreContentRecordWithoutID">
          Is this already deployed to a Connect server? You can 
          <a class="webview-link" role="button" @click="onAssociateDeployment"
            >update that previous deployment</a
          >
        </div>
        <div v-if="isPreContentRecordWithID">
          ✔️ will update <a class="webview-link" role="button" @click="viewContentURL">the existing content</a> when deployed.
        </div>
```

extensions/vscode/src/api/types/contentRecords.ts
```diff
@@ -31,6 +31,13 @@ export type ContentRecordError = {
   state: ContentRecordState.ERROR;
 } & ContentRecordLocation;
 
+type OptionalPreDeploymentFields = {
+  id?: string;
+  bundleUrl?: string;
+  dashboardUrl?: string;
+  directUrl?: string;
+};
```
dotNomad, line 39: I noticed that in `EvenEasierDeploy.vue` `isPreContentRecordWithID` and `isPreContentRecordWithoutID` was added.

We could potentially split up `PreContentRecord` into these two types to accomplish the same thing that is sort of being split up here (between type files and component files)

extensions/vscode/src/utils/guid.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+export const GuidRegex =
+  /(?:\{{0,1}(?:[0-9a-fA-F]){8}-(?:[0-9a-fA-F]){4}-(?:[0-9a-fA-F]){4}-(?:[0-9a-fA-F]){4}-(?:[0-9a-fA-F]){12}\}{0,1})/;
```
dotNomad, line 4: Very nice! Played around a bit in `regexr` to double check this to the best of my ability.

extensions/vscode/src/utils/guid.ts
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+export const GuidRegex =
```
dotNomad, line 3: Nitpicky: Typically I see these types of constants named like `GUID_REGEX`

extensions/vscode/src/views/homeView.ts
```diff
@@ -979,6 +984,53 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     return commands.executeCommand(Commands.Logs.Focus);
   }
 
+  private async showAssociateGUID() {
```
dotNomad, line 987: Is there a good way for us to break this out of `homeView`? The file is getting rather cumbersome, and this feels like a candidate that we could extract into a single file.

extensions/vscode/src/api/resources/ContentRecords.ts
```diff
@@ -104,12 +104,18 @@ export class ContentRecords {
   // 204 - no content
   // 404 - contentRecord or config file not found
   // 500 - internal server error
-  patch(deploymentName: string, configName: string, dir: string) {
+  patch(
+    deploymentName: string,
+    dir: string,
+    configName?: string,
+    guid?: string,
```
dotNomad, line 111: Usability suggestion we could make this a `body` with optional attributes which would mean we wouldn't need to pass `undefined` when sending the `guid` for example.

extensions/vscode/src/views/homeView.ts
```diff
@@ -979,6 +984,53 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+    let urlOrGuid = "";
+    const result = await window.showInputBox({
+      title: "Enter the URL of the Existing Content Item on the Server.",
+      prompt:
+        "You can supply either the content's dashboard URL or the solo URL.",
+      value: urlOrGuid,
+      placeHolder:
+        "For example: https://connect.company.co/connect/#/apps/adffa505-08c7-450f-88d0-f42957f56eff",
```
dotNomad, line 995: This is already pretty long, I would simplify a bit by just providing the placeholder value.

```suggestion
        "https://connect.company.co/connect/#/apps/adffa505-08c7-450f-88d0-f42957f56eff",
```

extensions/vscode/src/views/homeView.ts
```diff
@@ -979,6 +984,53 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     return commands.executeCommand(Commands.Logs.Focus);
   }
 
+  private async showAssociateGUID() {
+    let urlOrGuid = "";
+    const result = await window.showInputBox({
+      title: "Enter the URL of the Existing Content Item on the Server.",
+      prompt:
+        "You can supply either the content's dashboard URL or the solo URL.",
```
dotNomad, line 992: To match wording on other inputs from VS Code:
CleanShot 2024-09-18 at 12 11 52@2x


```suggestion
        "Please provide the content's dashboard URL or solo URL.",
```

extensions/vscode/src/views/homeView.ts
```diff
@@ -979,6 +984,53 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+      value: urlOrGuid,
+      placeHolder:
+        "For example: https://connect.company.co/connect/#/apps/adffa505-08c7-450f-88d0-f42957f56eff",
+      validateInput: (text) => {
+        const guid = extractGUID(text);
+        if (guid === null) {
+          return Promise.resolve({
+            message: `Unexpected format for a Posit Connect Content URL. Confirm URL value loads content from server and contains a content GUID.`,
```
dotNomad, line 1000: Wording is tough, and I quite liked this error 👌 

```suggestion
            message: `Unexpected format for a Posit Connect Content URL. Confirm the URL loads content from the server and contains a content GUID.`,
```

internal/deployment/deployment.go
```diff
@@ -104,9 +104,15 @@ func FromFile(path util.AbsolutePath) (*Deployment, error) {
...
+	if d.DashboardURL == "" && d.ID != "" {
+		d.DashboardURL = util.GetDashboardURL(d.ServerURL, d.ID)
+	}
+	if d.DirectURL == "" && d.ID != "" {
+		d.DirectURL = util.GetDirectURL(d.ServerURL, d.ID)
+	}
 	if d.LogsURL == "" && d.DashboardURL != "" {
-		// Migration
-		d.LogsURL = util.URLJoin(d.DashboardURL, "logs")
+		d.LogsURL = util.GetLogsURL(d.ServerURL, d.ID)
```
dotNomad, line 115: Is the migration necessary here? If I'm understanding correctly pre-deployments didn't have `id` and full deployments did. In the full deployment case the URLs were always provided right?

It looks like the update to `patch_deployment` also writes the URLs.

internal/util/urls.go
```diff
@@ -13,3 +16,19 @@ func NormalizeServerURL(serverURL string) (string, error) {
...
+
+func GetLogsURL(accountURL string, contentID types.ContentID) string {
+	return GetDashboardURL(accountURL, contentID) + "/logs"
+}
+
+func GetDirectURL(accountURL string, contentID types.ContentID) string {
+	return fmt.Sprintf("%s/content/%s/", accountURL, contentID)
+}
+
+func GetBundleURL(accountURL string, contentID types.ContentID, bundleID types.BundleID) string {
```
dotNomad, line 32: Fantastic split out for these functions.

dotNomad (approved):
This looks fantastic. I have quite a few comments, but most of them are praise, suggestions for improvement, or nitpicks.

The only thing that I think should be addressed is some of the wording suggestions.

I'll happily give a re-review if it is desired if any of those things are addressed.

extensions/vscode/src/views/homeView.ts
jonkeane: What about even simpler: "Please provide the content's URL from the Connect server"  or "Please provide the content's URL from Connect"

Both `dashboard` and `solo` are terms we use in Connect, but are likely to be misunderstood by actual users of connect ("Yes, I know my dashboard is on Connect, what do you mean 'the content's dashboard' the content *is* the dashboard."). Solo is similar, and we don't need to list all the things we accept here, just the most common one. If someone tries something else and it works that's fine

dotNomad: I think that is an even better suggestion. With this working to accept any "content URL" (in the form of dashboard URL or solo URL) we don't need to specify since any will work.

"Please provide the content's URL from Connect" is my preference so far. "from the Connect server" also feels verbose given how we talk about instances of Connect.

internal/deployment/deployment.go
sagerb: No, you're right. I do not think the migration is needed at this time. Mike placed the original case in there because we didn't always have the logs URL present in deployment files, but my changes are going in as one merge, so we won't have the issue of having predeployment files with the ID but without the URLs. (I had them within my development environment because of the incremental development process.)

Thanks for pointing this out. I'll remove my migration and make a minor change to what Mike had there before.

extensions/vscode/src/views/homeView.ts
sagerb: updated!

extensions/vscode/src/views/homeView.ts
sagerb: changed.

extensions/vscode/src/api/resources/ContentRecords.ts
sagerb: Great recommendation. I've updated the third parameter to now be a body structure with optional attributes and retrofit the calling code. I like it much better.

extensions/vscode/src/utils/guid.ts
sagerb: updated.

extensions/vscode/src/utils/guid.ts
sagerb: Thanks!

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
dotNomad: Was thinking about this a bit more to see if I could provide some solid suggestions. The only other note I'll add is that the unicode `✔️` is a new pattern. I like the tone that icon can provide, I think its just the first non-error (`X`) icon we use.

extensions/vscode/src/api/types/contentRecords.ts
sagerb: Discussed with Jordan and we decided to hold off on this change. There are complications in implementing two additional types which then multiply into the other variants (such as with and without config).

extensions/vscode/src/views/homeView.ts
sagerb: I've created an `actions` folder, and have moved this method into its own file. The only thing that it requires is that the `PublisherState` be passed into it as a parameter.

If you like this approach, I can see that we could easily move all of the "on" actions that are dispatched from `onConduitMessage`. I think we could go further, but in order to limit the testing impact of a refactoring, I think we should come back after the pending release.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -148,7 +158,11 @@
           />
         </div>
         <div
-          v-if="!isPreContentRecord(home.selectedContentRecord)"
+          v-if="
+            !isPreContentRecordWithoutID &&
+            home.selectedContentRecord.dashboardUrl !== undefined &&
+            home.selectedContentRecord.dashboardUrl !== ''
+          "
```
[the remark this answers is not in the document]
sagerb: Good catch.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: Hesitant to introduce the symbolic check mark since we're not using that style anywhere currently. I do think that when we enhance this box to show what will be happening when the user clicks the Deploy button, it will be a great candidate for the UX (list of checks and X's), but for now, I'd suggest we go with something a little different:

```
        <div v-if="isPreContentRecordWithoutID">
          Is this already deployed to a Connect server? You can 
          <a class="webview-link" role="button" @click="onAssociateDeployment"
            >update that previous deployment</a
          >.
        </div>
        <div v-if="isPreContentRecordWithID">
          Existing content will be updated when deployed.
        </div>
```

dotNomad (approved):
Thanks for those changes - pulling out `showAssociateGUID` is so much better, and looks like the `PublisherState` was helpful there!

marcosnav (approved):
LGTM 💯

Issue #2257, opened 2024-09-04
Secrets: Clear UI secret state


mmarchetti:
Clear the list of secrets in the UI on a successful deployment, or when the user selects a different deployment.

PR #2288, opened 2024-09-18
Add Secrets view and deploy with Secrets to extension

### LINKED ISSUE #2252: Secrets: New UI panel for Secrets
Create a new collapsible panel at the same level as Python packages, R Packages, etc. It shows the list of secrets from the configuration, with an indicator showing whether each one has been loaded in the current session.

dotNomad:
This PR adds a new Secrets section to the Home view.

Some new features:
- Secret values present are sent up on Deploy and added as environment variables on Connect. If the env var is present it's value is overwritten
- When changing Deployments the provided Secret values are cleared
- When a successful Deployment occurs the provided Secret values are cleared
- When a Configuration is changed, but the Deployment stays the same, the shared Secrets keep their values in the UI
- The number of Secrets with values is displayed as a Count badge on the Secrets section header to help the user know what is going to be sent up
- When editing a Secret the input will be selected so all of the value is highlighted for easy overwrite (similar to the rename functionality in VS Code)

<details>
  <summary>Preview</summary>
  
  https://github.com/user-attachments/assets/8134bad2-e327-4146-bbd8-a011c8e8e2b0
</details> 

## Intent

Resolves #2252, #2257, #2258

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach here was to expand previously made components, and create new ones (when necessary) to allow for the new model of "editing" a Tree Item description.

In this case editing the value of the Secret items.

Actions are used to edit and clear Secrets. Instead of clicking directly the action button is more clear, and avoids issues with duplicated click events when the input field is already up.

When editing a Secret value the action buttons are removed from the UI until editing is complete. This was done to avoid confusing the user and dealing with tricky cases like what clicking "edit" or "clear" should do while the input is open. It also provides more space for the input field which is horizontally constrained.

## Directions for Reviewers

Ensure that Tree Item descriptions function and are styled as they previously were.

Ensure the new Tree Item with description input (the Secrets) function as expected.

Ensure that Secrets with values are sent up and are correct on the Connect Server. Secrets without values should remain the same across multiple re-deployments.

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
+++ b/extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
@@ -36,7 +36,10 @@
       />
       <div class="tree-item-label-container">
         <span class="tree-item-title" data-automation="req">{{ title }}</span>
-        <span v-if="description" class="tree-item-description">
+        <span v-if="$slots.description" class="tree-item-description">
+          <slot name="description" />
+        </span>
+        <span v-else-if="description" class="tree-item-description">
           {{ description }}
         </span>
       </div>
@@ -79,6 +82,7 @@ withDefaults(defineProps<Props>(), {
 
 defineSlots<{
   default(props: { indentLevel: number }): any;
+  description(): any;
   postDecor(): any;
 }>();
 
@@ -151,6 +155,7 @@ const toggleExpanded = () => {
 
     .tree-item-label-container {
       flex: 1;
+      display: flex;
       min-width: 0;
       overflow: hidden;
       text-overflow: ellipsis;
@@ -159,13 +164,24 @@ const toggleExpanded = () => {
         line-height: 22px;
         color: inherit;
         white-space: pre;
+        min-width: 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
       }
 
       .tree-item-description {
+        line-height: 22px;
         font-size: 0.9em;
         margin-left: 0.5em;
-        opacity: 0.7;
+        flex-shrink: 100000;
         white-space: pre;
+        min-width: 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
+
+        &:not(:has(input)) {
+          opacity: 0.7;
+        }
       }
     }
 
--- a/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
@@ -3,6 +3,7 @@
     :title="name"
     :actions="actions"
     codicon="codicon-lock-small"
+    :list-style="secretValue ? 'default' : 'deemphasized'"
     :tooltip="tooltip"
     align-icon-with-twisty
   >
--- a/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
@@ -1,0 +1,75 @@
+<template>
+  <form v-if="showInput" @submit.prevent="updateSecret">
+    <input
+      ref="input"
+      v-model="inputValue"
+      aria-label="Type secret value. Press Enter to confirm or Escape to cancel."
+      @keydown.escape="showInput = false"
+    />
+  </form>
+  <TreeItem
+    v-else
+    :title="name"
+    :description="value ? '••••••' : undefined"
+    :actions="actions"
+    codicon="codicon-lock-small"
+    :tooltip="tooltip"
+    align-icon-with-twisty
+    @click="inputSecret"
+  />
+</template>
+
+<script setup lang="ts">
+import { computed, ref, nextTick } from "vue";
+
+import TreeItem from "src/components/tree/TreeItem.vue";
+import { useHomeStore } from "src/stores/home";
+import { ActionButton } from "src/components/ActionToolbar.vue";
+
+interface Props {
+  name: string;
+  value?: string;
+}
+
+const props = defineProps<Props>();
+
+const input = ref<HTMLInputElement | null>(null);
+const showInput = ref(false);
+const inputValue = ref(props.value);
+
+const home = useHomeStore();
+
+const inputSecret = () => {
+  showInput.value = true;
+  nextTick(() => input.value?.focus());
+};
+
+const updateSecret = () => {
+  home.secrets.set(props.name, inputValue.value);
+  showInput.value = false;
+};
+
+const tooltip = computed(() => {
+  if (props.value) {
+    return "On the next deploy the new value will be set for the deployment.";
+  }
+
+  return "No value has been set. The value will not change on the next deploy.";
+});
+
+const actions = computed<ActionButton[]>(() => {
+  const result = [];
+
+  if (props.value) {
+    result.push({
+      label: "Clear Value",
+      codicon: "codicon-x",
+      fn: () => {
+        home.secrets.set(props.name, undefined);
+      },
+    });
+  }
+
+  return result;
+});
+</script>
--- a/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
@@ -41,7 +41,7 @@ const home = useHomeStore();
 
 const inputSecret = () => {
   showInput.value = true;
-  nextTick(() => input.value?.focus());
+  nextTick(() => input.value?.select());
 };
 
 const updateSecret = () => {
--- a/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
@@ -3,7 +3,7 @@
     :title="name"
     :actions="actions"
     codicon="codicon-lock-small"
-    :list-style="secretValue ? 'default' : 'deemphasized'"
+    :list-style="secretValue || isEditing ? 'default' : 'deemphasized'"
     :tooltip="tooltip"
     align-icon-with-twisty
   >
--- a/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
@@ -81,7 +81,7 @@ const actions = computed<ActionButton[]>(() => {
   if (secretValue.value) {
     result.push({
       label: "Clear Value",
-      codicon: "codicon-x",
+      codicon: "codicon-remove",
       fn: () => {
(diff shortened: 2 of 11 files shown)


extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
```diff
@@ -0,0 +1,94 @@
...
+
+<script setup lang="ts">
+import { computed, ref, nextTick } from "vue";
+
+import TreeItem from "src/components/tree/TreeItem.vue";
+import { useHomeStore } from "src/stores/home";
+import { ActionButton } from "src/components/ActionToolbar.vue";
+
+interface Props {
+  name: string;
```
sagerb, line 31: I would have expected that this component would be a UI component that would accept its name and value through a prop and emit a change when the value changed. I would then have expected the view using this component to have to deal with the storage of the secret value. 

While this approach will work, it makes this component extremely dedicated to this particular input vs. having a component that allows its value to be edited. I think we would benefit from having it be a bit more general purpose, especially when you think of using it for environment variables as well.

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -57,6 +60,32 @@ export const useHomeStore = defineStore("home", () => {
     },
   );
 
+  watch(
+    [selectedContentRecord, selectedConfiguration],
+    ([contentRecord, config], [prevContentRecord, prevConfig]) => {
+      const result = new Map<string, string | undefined>();
+
+      if (config === undefined || isConfigurationError(config)) {
+        return result;
```
sagerb, line 69: return out of a watch? Does this do something that I'm not aware of?

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -57,6 +60,32 @@ export const useHomeStore = defineStore("home", () => {
     },
   );
 
+  watch(
+    [selectedContentRecord, selectedConfiguration],
```
sagerb, line 64: I was wondering about simplifying this into two different watches.
1. on change of content record, remove all secrets.
2. on change of config, set the secret from the values present and clear the ones that have gone away.

I've spent a few minutes making sure I understand what your implementation was doing and I think splitting it may avoid that for others in the future.

sagerb (commented):
Some early feedback, happy to discuss or explain my observations. Thanks!

extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
dotNomad: I actually started writing it this way at the start, but found it more difficult to keep the `home.secrets` Map in line with the value that was being `v-model`ed in `Secret.vue` instead I made it just reflect the state of `home.secrets.get('name')` for the `name` key provided

dotNomad: The main difficulty when using `v-model` was grabbing the most recent value when swapping over to the input mode. It worked great when setting the value, but not when the value changed out from under the component.

extensions/vscode/webviews/homeView/src/stores/home.ts
dotNomad: Oops this should be setting `secrets.value` then bailing out.

dotNomad: Fixed in https://github.com/posit-dev/publisher/pull/2288/commits/08bccd44ed6aef1313ff2db0aaaf41d8bda1e4f0

extensions/vscode/webviews/homeView/src/stores/home.ts
dotNomad: I initially planned to do this as two separate watches due to the reason you stated, but I was worried about any timing issues that could have spawned from that. For example, some code somewhere could potentially change the content record first then the configuration, and code elsewhere does the opposite potentially leaving secrets in a different state.

Keeping it as I have it avoids that by ensuring secrets is in the correct state no matter if one or both change.

Let me know if I'm making a bad assumption though. I would like to simplify if possible.

sagerb: Would the opposite order really end up not equaling the same outcome?

1. Content-Record changes and then Configuration changes
- Content-Record change removes all secrets
- Configuration change has no secrets set, so it does nothing

2. Configuration changes and then Content-Record changes
- Configuration change sets the secret from the values present and clears the ones that have gone away 
- Content-Record change removes all secrets

In both cases, no secrets are set after both of these events.

Am I seeing this wrong?

dotNomad: No I think you got it, let me play around with this split into two watches and see if I can break anything. Your logic is sound.

extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
```diff
@@ -151,6 +155,7 @@ const toggleExpanded = () => {
 
     .tree-item-label-container {
       flex: 1;
+      display: flex;
```
dotNomad, line 158: Had to bring in flex for the label container so that the input in the description `span` could be displayed the way we want. This required bringing down the

```css
min-width: 0;
overflow: hidden;
text-overflow: ellipsis;
```

CSS for dealing with overflowing with ellipsis.

extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
```diff
@@ -159,13 +164,24 @@ const toggleExpanded = () => {
         line-height: 22px;
         color: inherit;
         white-space: pre;
+        min-width: 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
       }
 
       .tree-item-description {
+        line-height: 22px;
```
dotNomad, line 173: `line-height` has returned. Now that we are using `display: flex` here the `span` is no longer being rendered the same way. Without this the height changes, and the description wasn't on the baseline.

extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
```diff
@@ -159,13 +164,24 @@ const toggleExpanded = () => {
...
         font-size: 0.9em;
         margin-left: 0.5em;
-        opacity: 0.7;
```
dotNomad, line 167: Since `description` can contain an `input` now we don't want the description to _always_ be `opacity: 0.7`. Now we apply this only if the description doesn't contain an `input` (see below)

extensions/vscode/webviews/homeView/src/components/tree/TreeItem.vue
```diff
@@ -159,13 +164,24 @@ const toggleExpanded = () => {
...
+        flex-shrink: 100000;
```
dotNomad, line 176: This retains the behavior we had before where the description would shrink before the title.

extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
```diff
@@ -0,0 +1,93 @@
...
+      label: "Edit Value",
+      codicon: "codicon-pencil",
+      fn: inputSecret,
+    },
+  ];
+
+  if (secretValue.value) {
+    result.push({
+      label: "Clear Value",
+      codicon: "codicon-x",
```
dotNomad, line 84: I'm unsure about the `x` icon for this.

extensions/vscode/webviews/homeView/src/components/views/secrets/Secrets.vue
```diff
@@ -0,0 +1,38 @@
...
+const sectionActions = computed<ActionButton[]>(() => [
+  {
+    label: "Clear Values for all Secrets",
+    codicon: "codicon-clear-all",
+    fn: () => {
+      home.secrets.forEach((_, key) => {
+        home.secrets.set(key, undefined);
+      });
+    },
+  },
```
dotNomad, line 36: I could optionally show this only if 1 or more secrets have values?

extensions/vscode/webviews/homeView/src/style.css
```diff
@@ -8,6 +8,10 @@ body {
   padding: 0;
 }
 
+::selection {
+  background-color: rgba(195, 183, 184, 0.15);
+}
```
dotNomad, line 13: This preserves the `input` selection color, and matches the highlight color across VS Code. This doesn't appear to be a variable, instead it is hardcoded which is interesting. I did check, it is consistent across themes.

extensions/vscode/webviews/homeView/src/stores/home.ts
dotNomad: I played around with this a bit, and it ended up being _really_ close to a duplication for both watchers. If you have a code block suggestion we can take a look.

extensions/vscode/webviews/homeView/src/components/views/secrets/Secrets.vue
dotNomad: The more I use this the more I want to do this, since the button moves if the count badge is showing or not.

extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
sagerb: Did you get a chance to play with the approach we discussed via zoom? That would be where outside component would pass in the map value from the store and then update the store when receiving the change emitted event from the component?

extensions/vscode/webviews/homeView/src/stores/home.ts
sagerb: I'm probably just missing something in my quick analysis, so I'll defer to yours. Let's go with what you've got working. Thanks for indulging me!

sagerb (approved):
Added a few responses, which I'll leave to your discretion. I'm approving it so you can determine where you do any followup.

extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
dotNomad: https://github.com/posit-dev/publisher/pull/2288/commits/45e5056d6cdd330377521cc7ac7f23079ab00cc0

extensions/vscode/webviews/homeView/src/components/views/secrets/Secret.vue
dotNomad: I did mess around with this, and it ended up re-creating the Map in the `Secrets` component a bit as an array so each `v-model` accessed an index in that array. It felt duplicative to me.

PR #2291, opened 2024-09-19
cleanup cypress tests

kgartland-rstudio:
This PR just removes the cypress tests that were used for the old standalone UI.

The old tests caused this failure so I figured I'd just delete them.

https://github.com/posit-dev/publisher/actions/runs/10942280011


sagerb (approved):
LGTM, we can always bring them back if we need to.

Issue #1831, opened 2024-06-17
Credential persistence for Workbench


mmarchetti:
Libraries needed for keyring support aren't always available in Workbench. #1824 introduces a temporary workaround that allows keys to persist for during your workbench session, so we can test. We need a solution to let them persist beyond the workbench session.

mmarchetti:
For 1.1.5, we will go with the implementation in #2063.

mmarchetti:
Updated to use a file-based temporary solution in #2089.

dotNomad:
Verified that this behaves as expected on Workbench, and the keychain is still used on my machine.

PR #2292, opened 2024-09-19
Credentials service with a logger

marcosnav:
## Intent

PR introduces quite simple changes for CredentialsService to have it's own logging instance, allowing to have further logging calls within credentials operations.

This is an initial work for #1831

### Adding a logger to credentials service

This PR adds a logger to `CredentialsService` which is now private - `credentialService` - such change for two reasons:
1. Continuity to #2153 - more debug logging where we are lacking
2. These changes help and are introductory work to #1831

### Why make credentials service private?

Making credentials service private is to prevent panics, since `credentialsService` will call a logger instance internally, making the struct private and the interface and constructor public, forces consumers of this service to pass a logger. E.g:

```go
// The following would bring a panic, program will compile with no issues but with the risk of runtime errors
// "runtime error: invalid memory address or nil pointer dereference"
cs := credentialsService{}
cs.Get(someGuid) // <- If logger is called for any reason within this method, program will panic
```

```go
// Using the private struct but a public constructor prevents that panic
cs := NewCredentialsService(lgr) // <- constructor expects and forces us to pass a logger for the resulting struct
cs.Get(someGuid) // Won't panic, no runtime errors
```

### No-op logger

The new no-op logger - `NewDiscardLogger` - is only used for the `cli`. Endpoints and any code related to the vscode extension is passing a proper logger.

I'm not going to bring more scope to this work. I agree that it'll be great to make the `cli` fully work, but that's not our priority at the moment. When we get to make the `cli` a priority we then can figure out a proper logging strategy for it.

## Type of Change

Enhancement on credentials operations for logging purposes.

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

- Have a `NewCredentialsService` constructor to force new instances to have a logger.
- Previous `CredentialService` is now a private package struct and we now have an interface for it.

## Automated Tests

Updated unit tests accordingly.

## Directions for Reviewers

Nothing specific, changes are pretty much repetitive

## Checklist

### DIFF
--- a/cmd/publisher/commands/credentials.go
+++ b/cmd/publisher/commands/credentials.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/posit-dev/publisher/internal/cli_types"
 	"github.com/posit-dev/publisher/internal/credentials"
+	"github.com/posit-dev/publisher/internal/logging"
 )
 
 type CredentialsCommand struct {
@@ -24,7 +25,7 @@ type CreateCredentialCommand struct {
 }
 
 func (cmd *CreateCredentialCommand) Run(args *cli_types.CommonArgs, ctx *cli_types.CLIContext) error {
-	cs := credentials.CredentialsService{}
+	cs := credentials.NewCredentialsService(logging.NewDiscardLogger())
 	cred, err := cs.Set(cmd.Name, cmd.URL, cmd.ApiKey)
 	if err != nil {
 		return err
@@ -44,7 +45,7 @@ type DeleteCredentialCommand struct {
 }
 
 func (cmd *DeleteCredentialCommand) Run(args *cli_types.CommonArgs, ctx *cli_types.CLIContext) error {
-	cs := credentials.CredentialsService{}
+	cs := credentials.NewCredentialsService(logging.NewDiscardLogger())
 	err := cs.Delete(cmd.GUID)
 	if err != nil {
 		return err
@@ -59,7 +60,7 @@ type GetCredentialCommand struct {
 }
 
 func (cmd *GetCredentialCommand) Run(args *cli_types.CommonArgs, ctx *cli_types.CLIContext) error {
-	cs := credentials.CredentialsService{}
+	cs := credentials.NewCredentialsService(logging.NewDiscardLogger())
 	cred, err := cs.Get(cmd.GUID)
 	if err != nil {
 		return err
@@ -78,7 +79,7 @@ type ListCredentialsCommand struct {
 }
 
 func (cmd *ListCredentialsCommand) Run(args *cli_types.CommonArgs, ctx *cli_types.CLIContext) error {
-	cs := credentials.CredentialsService{}
+	cs := credentials.NewCredentialsService(logging.NewDiscardLogger())
 	creds, err := cs.List()
 	if err != nil {
 		return err
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -32,6 +32,7 @@ import (
 	"net/url"
 
 	"github.com/google/uuid"
+	"github.com/posit-dev/publisher/internal/logging"
 	"github.com/posit-dev/publisher/internal/util"
 	"github.com/spf13/afero"
 	"github.com/zalando/go-keyring"
@@ -74,8 +75,21 @@ func (cr *CredentialRecord) ToCredential() (*Credential, error) {
 	}
 }
 
-type CredentialsService struct {
+type CredentialsService interface {
+	FileCredentialRecordFactory() (*CredentialRecord, error)
+	Delete(guid string) error
+	Get(guid string) (*Credential, error)
+	List() ([]Credential, error)
+	Set(name string, url string, ak string) (*Credential, error)
+}
+
+type credentialsService struct {
 	afs afero.Fs
+	log logging.Logger
+}
+
+func NewCredentialsService(log logging.Logger) *credentialsService {
+	return &credentialsService{log: log}
 }
 
 // FileCredentialRecordFactory creates a Credential based on the presence of the
@@ -86,7 +100,7 @@ type fileCredential struct {
 	Key string `toml:"key"`
 }
 
-func (cs *CredentialsService) FileCredentialRecordFactory() (*CredentialRecord, error) {
+func (cs *credentialsService) FileCredentialRecordFactory() (*CredentialRecord, error) {
 	homeDir, err := util.UserHomeDir(cs.afs)
 	if err != nil {
 		return nil, err
@@ -148,7 +162,7 @@ func (cs *CredentialsService) FileCredentialRecordFactory() (*CredentialRecord, 
 	return nil, nil
 }
 
-func (cs *CredentialsService) checkForConflicts(
+func (cs *credentialsService) checkForConflicts(
 	table *map[string]CredentialRecord,
 	c *Credential) error {
 	// Check if Credential attributes (URL or name) are already used by another credential
@@ -187,7 +201,7 @@ func (cs *CredentialsService) checkForConflicts(
 
 // Delete removes a Credential by its guid.
 // If lookup by guid fails, a NotFoundError is returned.
-func (cs *CredentialsService) Delete(guid string) error {
+func (cs *credentialsService) Delete(guid string) error {
 
 	table, err := cs.load()
 	if err != nil {
@@ -196,6 +210,7 @@ func (cs *CredentialsService) Delete(guid string) error {
 
 	_, exists := table[guid]
 	if !exists {
+		cs.log.Debug("Credential does not exist", "credential", guid)
 		return &NotFoundError{GUID: guid}
 	}
 
@@ -209,7 +224,7 @@ func (cs *CredentialsService) Delete(guid string) error {
 }
 
 // Get retrieves a Credential by its guid.
-func (cs *CredentialsService) Get(guid string) (*Credential, error) {
+func (cs *credentialsService) Get(guid string) (*Credential, error) {
 	table, err := cs.load()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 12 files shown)


cmd/publisher/commands/credentials.go
```diff
@@ -24,7 +25,7 @@ type CreateCredentialCommand struct {
 }
 
 func (cmd *CreateCredentialCommand) Run(args *cli_types.CommonArgs, ctx *cli_types.CLIContext) error {
-	cs := credentials.CredentialsService{}
+	cs := credentials.NewCredentialsService(logging.NewDiscardLogger())
```
marcosnav, line 28: We are using a new `NewDiscardLogger()` to pass a no-op logging instance where we don't care too much about log output. (...at least for now)

internal/credentials/credentials_test.go
```diff
@@ -26,20 +28,33 @@ var fileCred = Credential{
 	Name:   "connect.localtest.me",
 }
 
-func credentialsFilePath(afs afero.Fs) (util.AbsolutePath, error) {
+type CredentialsTestSuite struct {
+	utiltest.Suite
```
marcosnav, line 32: Updated this tests to be a testify suite for consistency with our other test files. Most changes in this file are just swapping `assert` calls to a `Suite s.*()` method. New relevant changes assert calls to a mock logger instance.

cmd/publisher/commands/credentials.go
sagerb: Why don't we care about logging? It's really the only diagnostic tool we have at this point if something goes wrong...?

internal/credentials/credentials.go
```diff
@@ -86,7 +100,7 @@ type fileCredential struct {
 	Key string `toml:"key"`
 }
 
-func (cs *CredentialsService) FileCredentialRecordFactory() (*CredentialRecord, error) {
+func (cs *credentialsService) FileCredentialRecordFactory() (*CredentialRecord, error) {
```
sagerb, line 103: I'm not following why you changed this to an unexported struct? 
Couldn't you have just changed the existing CredentialsService to have a logger? 

Perhaps it would explain your statement from this morning of why you had to break it out to a separate logger?

internal/credentials/credentials.go
```diff
@@ -196,6 +210,7 @@ func (cs *CredentialsService) Delete(guid string) error {
 
 	_, exists := table[guid]
 	if !exists {
+		cs.log.Debug("Credential does not exist", "credential", guid)
```
sagerb, line 213: Further not following why you've got logging messages in here if you are initially going to a discarding logger which won't do anything?

sagerb (commented):
Can you explain further what your reasons were for the two major changes here (dedicated no-op logger and a switch away from the existing CredentialService interface to a non-exported one? I'm just missing the context, and from my understanding, this much change should not be needed here. I'm probably wrong, just need the extra explanation. Thanks!

cmd/publisher/commands/credentials.go
marcosnav: The credentials service being created here is for `cli` tooling. IIRC we are not prioritizing `cli` behavior, passing a no-op logger here to comply with the args without having to dig on the logging strategy for `cli` commands.

internal/credentials/credentials.go
marcosnav: This change is to prevent panics, since `credentialsService` will call a logger instance internally, making the struct private and the interface and constructor public, forces consumers of this service to pass a logger. E.g:

```go
// The following would bring a panic, program will compile with no issues but with the risk of runtime errors
// "runtime error: invalid memory address or nil pointer dereference"
cs := credentialsService{}
cs.Get(someGuid) // <- If logger is called for any reason within this method, program will panic
```

```go
// Using the private struct but a public constructor prevents that panic
cs := NewCredentialsService(lgr) // <- constructor expects and forces us to pass a logger for the resulting struct
cs.Get(someGuid) // Won't panic, no runtime errors
```

internal/credentials/credentials.go
marcosnav: We are using proper loggers within the endpoints that create `NewCredentialService`, the only no-op logger is for `cli` tooling

cmd/publisher/commands/credentials.go
tdstein: Using verbosity arguments may be a better solution. But, I agree with @sagerb, its better to be overly verbose here instead of quiet.

aside; it would be really great to make the cli real

sagerb: The `CredentialService` is used to respond to existing API requests as well currently. All we wanted to do in this PR was to add an additional implementation that would not use the keychain for persistence. I'm not following you why we have to make this much of a change.

marcosnav:
@sagerb @tdstein Sorry for not being clear in advance on some of this changes, I'm also adding the following to the PR description.

### Adding a logger to credentials service

This PR adds a logger to `CredentialsService` which is now private - `credentialService` - such change for two reasons:
1. Continuity to #2153 - more debug logging where we are lacking
2. This changes help and are introductory work to #1831

### Why make credentials service private?

Making credentials service private is to prevent panics, since `credentialsService` will call a logger instance internally, making the struct private and the interface and constructor public, forces consumers of this service to pass a logger. E.g:

```go
// The following would bring a panic, program will compile with no issues but with the risk of runtime errors
// "runtime error: invalid memory address or nil pointer dereference"
cs := credentialsService{}
cs.Get(someGuid) // <- If logger is called for any reason within this method, program will panic
```

```go
// Using the private struct but a public constructor prevents that panic
cs := NewCredentialsService(lgr) // <- constructor expects and forces us to pass a logger for the resulting struct
cs.Get(someGuid) // Won't panic, no runtime errors
```

### No-op logger

The new no-op logger - `NewDiscardLogger` - is only used for the `cli`. Endpoints and any code related to the vscode extension is passing a proper logger.

I'm not going to bring more scope to this work. I agree that it'll be great to make the `cli` fully work, but that's not our priority at the moment. When we get to make the `cli` a priority we then can figure out a proper logging strategy for it.

tdstein (commented):
Hey @marcosnav - here is another approach you could try that avoids having to reimplementing the credentials service.

Take a look at https://github.com/posit-dev/publisher/blob/c1303e7ac860e326612d6a5ac4f15d71252ec35a/internal/credentials/keyring.go#L11

Can you turn this into an interface and implement a second version that writes to a file instead of using go-keyring? This would be much simpler as you just need to create a simple kv store with an optional service namespace.

Let me know if you would like to pair on this to get started.

sagerb:
>Can you turn this into an interface and implement a second version...

@marcosnav This is more inline with what my expectations were for this change. I'm sorry if we didn't communicate it well within the original issue or our internal discussions.

tdstein:
Here's what ChatGPT o1-preview threw together. I don't know if it even compiles, but its something to work off of.

```go

package credentials

import (
	"encoding/json"
	"errors"
	"os"
	"sync"
)

type JSONCredentialStore struct {
	filePath string
	mu       sync.Mutex
	data     map[string]map[string]string
}

func NewJSONCredentialStore(filePath string) *JSONCredentialStore {
	return &JSONCredentialStore{
		filePath: filePath,
		data:     make(map[string]map[string]string),
	}
}

func (store *JSONCredentialStore) load() error {
	file, err := os.Open(store.filePath)
	if err != nil {
		if os.IsNotExist(err) {
			// Initialize empty data if file doesn't exist
			store.data = make(map[string]map[string]string)
			return nil
		}
		return err
	}
	defer file.Close()

	decoder := json.NewDecoder(file)
	if err := decoder.Decode(&store.data); err != nil {
		return err
	}
	return nil
}

func (store *JSONCredentialStore) save() error {
	file, err := os.OpenFile(store.filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
	if err != nil {
		return err
	}
	defer file.Close()

	encoder := json.NewEncoder(file)
	if err := encoder.Encode(store.data); err != nil {
		return err
	}
	return nil
}

func (store *JSONCredentialStore) Set(service, user, password string) error {
	store.mu.Lock()
	defer store.mu.Unlock()

	if err := store.load(); err != nil {
		return err
	}

	if store.data[service] == nil {
		store.data[service] = make(map[string]string)
	}
	store.data[service][user] = password

	if err := store.save(); err != nil {
		return err
	}
	return nil
}

func (store *JSONCredentialStore) Get(service, user string) (string, error) {
	store.mu.Lock()
	defer store.mu.Unlock()

	if err := store.load(); err != nil {
		return "", err
	}

	if users, ok := store.data[service]; ok {
		if password, ok := users[user]; ok {
			return password, nil
		}
	}
	return "", errors.New("password not found")
}

func (store *JSONCredentialStore) Delete(service, user string) error {
	store.mu.Lock()
	defer store.mu.Unlock()

	if err := store.load(); err != nil {
		return err
	}

	if users, ok := store.data[service]; ok {
		delete(users, user)
		if len(users) == 0 {
			delete(store.data, service)
		}
	}

	if err := store.save(); err != nil {
		return err
	}
	return nil
}
```

marcosnav:
@tdstein Thanks! that's exactly what I'm doing 😃 . I'm creating a new on-file service that'll implement the same methods 👍

marcosnav:
@sagerb It was pretty well communicated! That's my current WIP, that does not change the fact that credentials service needs a logger to follow and log out what errors that the Keyring service or the upcoming in file credentials service will throw

tdstein:
> @tdstein Thanks! that's exactly what I'm doing 😃 . I'm creating a new on-file service that'll implement the same methods 👍

Awesome. Let me know if you get stuck. Happy to help.

sagerb (approved):
@marcosnav Thanks for the clarifications. I've looked at some of the existing agent code (in particular `r.go` for the R inspector, and your approach feels much more inline with the existing implementations than I thought. Thanks so much for your patience!

PR #2294, opened 2024-09-19
Update QuickPickItem and EvenEasierDeploy to show entrypoint more prominently.

### LINKED ISSUE #2248: Make the entrypoint file more prominent within the deployment selector
Currently the entrypoint file is shown, but appears less significant than the deployment (which is showing title and credential nickname).

Since the entrypoint is the top-level object in a deployment, we need to show it as more prominent. One idea is to update the select button to show it ahead of the title (perhaps with 3 lines), or place it on the same line as the credential. (But horizontal space is limited, so we should probably explore 3 lines first.)

When this is changed, we should also examine how the multi-selector can also show the entrypoint as part of the 2 line entry for each deployment, so that the concepts are consistent.

sagerb:
## Intent

Resolves #2248 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

With this change, the entrypoint is now shown on the button displaying the current selected deployment as well as within the quick pick list of available deployments, as seen below:

2024-09-19 at 2 13 PM
2024-09-19 at 2 14 PM

## Directions for Reviewers

Simply build and verify new functionality is displayed.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -1083,7 +1083,12 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           details.push(`Missing Credential for ${contentRecord.serverUrl}`);
           problem = true;
         }
-        if (!isRelativePathRoot(contentRecord.projectDir)) {
+
+        if (isRelativePathRoot(contentRecord.projectDir)) {
+          if (config && !isConfigurationError(config)) {
+            details.push(config.configuration.entrypoint);
+          }
+        } else {
           if (config && !isConfigurationError(config)) {
             details.push(
               `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`,
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -271,6 +271,15 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
   }
 
   private async onInitializingMsg() {
+    // inform webview of the platform specific path separator
+    // (path package is a node library, wrapper for browser doesn't seem to work in webview correctly)
+    this.webviewConduit.sendMsg({
+      kind: HostToWebviewMessageType.SET_PATH_SEPARATOR,
+      content: {
+        separator: path.sep,
+      },
+    });
+
     // send back the data needed. Optimize request for initialization...
     await this.refreshAll(false, true);
     this.setInitializationContext(HomeViewInitialized.initialized);
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -1068,14 +1068,20 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
         }
 
         let details = [];
-        if (!isRelativePathRoot(contentRecord.projectDir)) {
-          details.push(`${contentRecord.projectDir}${path.sep}`);
-        }
         if (credential?.name) {
           details.push(credential.name);
         } else {
           details.push(`Missing Credential for ${contentRecord.serverUrl}`);
           problem = true;
+        }
+        if (
+          !isRelativePathRoot(contentRecord.projectDir) &&
+          config &&
+          !isConfigurationError(config)
+        ) {
+          details.push(
+            `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`,
+          );
         }
         const detail = details.join(" • ");
 
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -1074,14 +1074,14 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           details.push(`Missing Credential for ${contentRecord.serverUrl}`);
           problem = true;
         }
-        if (
-          !isRelativePathRoot(contentRecord.projectDir) &&
-          config &&
-          !isConfigurationError(config)
-        ) {
-          details.push(
-            `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`,
-          );
+        if (!isRelativePathRoot(contentRecord.projectDir)) {
+          if (config && !isConfigurationError(config)) {
+            details.push(
+              `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`,
+            );
+          } else {
+            details.push(`${contentRecord.projectDir}${path.sep}`);
+          }
         }
         const detail = details.join(" • ");
 
--- a/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
+++ b/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
@@ -332,19 +332,15 @@ const entrypointSubTitle = computed(() => {
   ) {
     const contentRecord = home.selectedContentRecord;
     const config = home.selectedConfiguration;
-    if (
-      contentRecord &&
-      config.configuration.entrypoint &&
-      !isConfigurationError(config)
-    ) {
-      return path.join(
-        contentRecord.projectDir,
-        config.configuration.entrypoint,
-      );
-      // return `${contentRecord.projectDir}/${config.configuration.entrypoint}`;
+    if (contentRecord) {
+      if (isConfigInError.value) {
+        return `${contentRecord.projectDir}${path.sep}`;
+      } else {
+        return `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`;
+      }
     }
   }
-  return "Entrypoint not determined";
+  return "ProjectDir and Entrypoint not determined";
 });
 
 const isCredentialMissing = computed((): boolean => {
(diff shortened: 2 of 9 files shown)


extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -326,13 +309,44 @@ const deploymentTitle = computed(() => {
...
+    if (
+      contentRecord &&
+      config.configuration.entrypoint &&
+      !isConfigurationError(config)
+    ) {
+      return path.join(
+        contentRecord.projectDir,
+        config.configuration.entrypoint,
+      );
+      // return `${contentRecord.projectDir}/${config.configuration.entrypoint}`;
```
dotNomad, line 344: Looks like this was left over.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -326,13 +309,44 @@ const deploymentTitle = computed(() => {
...
+  if (
+    home.selectedConfiguration &&
+    !isConfigurationError(home.selectedConfiguration)
+  ) {
+    const contentRecord = home.selectedContentRecord;
+    const config = home.selectedConfiguration;
+    if (
+      contentRecord &&
+      config.configuration.entrypoint &&
+      !isConfigurationError(config)
```
dotNomad, line 338: Looks like a duplicate check here. We check if the config is in error above. We could simplify by setting `const config` earlier in the function

extensions/vscode/src/views/homeView.ts
```diff
@@ -1068,15 +1068,21 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
         }
+        if (
+          !isRelativePathRoot(contentRecord.projectDir) &&
+          config &&
+          !isConfigurationError(config)
+        ) {
+          details.push(
+            `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`,
+          );
+        }
```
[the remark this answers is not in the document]
sagerb: Good catch! I've updated the logic and rather than optionally appending it, I've gone with a simple if statement. Given the two choices, it seemed the if statement was more readable.

extensions/vscode/webviews/homeView/vite.config.ts
```diff
@@ -16,6 +16,7 @@ export default defineConfig({
   resolve: {
     alias: {
       src: fileURLToPath(new URL("./src", import.meta.url)),
+      path: "path-browserify",
```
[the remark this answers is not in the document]
sagerb: If you look into the `package-lock.json` file, you'll see that it is a dependency of `@volar/typescript`. That is why it is working, but it would throw us if we removed that, so let me go ahead and make it a first-class dependency of our project.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -27,24 +27,6 @@
...
-        v-if="
-          home.selectedConfiguration &&
-          !isConfigurationError(home.selectedConfiguration) &&
-          home.selectedConfiguration?.configuration?.entrypoint
-        "
-        class="deployment-details-container"
-      >
-        <div class="deployment-details-row">
-          <span
-            class="deployment-details-label"
```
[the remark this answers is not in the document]
sagerb: Great! Thanks for finding this!

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: removed. Thanks!

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: I refactored this to match the earlier input regarding just showing the project dir when we can't determine the entrypoint, but I believe we still need a check here (which I've changed to use the computed function that validates the config). When a configuration is in error, that is shown below the button, but we still need to show some values within the button right? Let me know if I'm not following the same path as you are, once you've looked at my changes. Thanks!

kgartland-rstudio:
This change will cause the UI tests to fail. @sagerb want me to jump on this branch and update the tests?

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -326,13 +309,40 @@ const deploymentTitle = computed(() => {
...
+    if (contentRecord) {
+      if (isConfigInError.value) {
+        return `${contentRecord.projectDir}${path.sep}`;
+      } else {
+        return `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`;
+      }
+    }
+  }
+  return "ProjectDir and Entrypoint not determined";
+});
```
marcosnav, line 344: Since this relies heavily on `home` store data, have you considered using a store getter to handle most of this? 
e.g:
```js
const entrypointSubTitle = computed(() => {
  return home.entryPointSubtitle || "ProjectDir and Entrypoint not determined";
});
```

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -357,11 +367,20 @@ const lastStatusDescription = computed(() => {
...
 
 const toolTipText = computed(() => {
+  let entrypoint = "unknown";
+  if (
+    home.selectedConfiguration &&
+    !isConfigurationError(home.selectedConfiguration) &&
+    home.selectedConfiguration.configuration.entrypoint
+  ) {
+    entrypoint = home.selectedConfiguration.configuration.entrypoint;
+  }
```
marcosnav, line 377: Same here, a new store getter could help on keeping this component's responsibilities a bit cleaner.
```js
const toolTipText = computed(() => {
  const entrypoint = home.currentEntrypoint || "unknown";
  
  return `Deployment Details
...
```

marcosnav (approved):
LGTM, a couple suggestions

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -326,13 +310,40 @@ const deploymentTitle = computed(() => {
...
+    home.selectedConfiguration &&
+    !isConfigurationError(home.selectedConfiguration)
+  ) {
+    const contentRecord = home.selectedContentRecord;
+    const config = home.selectedConfiguration;
+    if (contentRecord) {
+      if (isConfigInError.value) {
+        return `${contentRecord.projectDir}${path.sep}`;
+      } else {
+        return `${contentRecord.projectDir}${path.sep}${config.configuration.entrypoint}`;
```
dotNomad, line 340: Here we are including the project dir and path separator even when the configuration is at the relative path root. I think this looks a bit strange.

CleanShot 2024-09-20 at 12 17 51@2x

It is definitely what we want when we aren't at the relative path root though

CleanShot 2024-09-20 at 12 18 07@2x

extensions/vscode/src/views/homeView.ts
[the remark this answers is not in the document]
dotNomad: Looks like the update didn't include this suggestion

> If we have a valid configuration then add ${path.sep}${config.configuration.entrypoint} to the end of it

At the relative path root I'm not seeing the entrypoint in the selection at all:

CleanShot 2024-09-20 at 12 20 27@2x

extensions/vscode/webviews/homeView/vite.config.ts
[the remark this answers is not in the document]
dotNomad: Definitely agree that any direct dependenies we have should be in our `package.json` just in case other dependencies remove theirs.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
dotNomad: The tooltip text feels appropriate in the component to me since it is the only use of it. Not opposed to having it in the store, but I would look for it in the component first instinctually

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
dotNomad: A helper to generate the entrypoint string based on relative path root, the selected configuration, and the path separator seems like a create function. It is just unfortunate that we need it in the home view, and in the extension code. That makes it a bit harder to generalize - although it is still possible.

dotNomad (commented):
The passing of the `path.sep` is a nice way to do this without any additional dependencies 👌 

Once we address the other concerns around showing the entrypoint I can take another look at this

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: Good catch. I've cleaned up that logic. It's now closer to what you originally asked for. Thanks for finding this.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: Thanks for the suggestion, but I agree that we want to keep this presentation calculation out of the store.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: If we were sharing the value, I could justify it being in the store, but right now, it feels better to me in the component as well.

extensions/vscode/webviews/homeView/vite.config.ts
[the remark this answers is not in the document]
dotNomad: Can we remove this now that `path-browserify` is not being used and we are passing in the path separator?

dotNomad:
Updated this PR to remove `path-browserify`, the unused import, and included the entrypoint in the deployment quick pick as described.

@sagerb should be ready for a look-see, and then merge.

sagerb (commented):
Thanks for making those changes, they all look good to me. Going to merge!

PR #2295, opened 2024-09-20
Add vscode-ui tests to Pull Request job in CI

kgartland-rstudio:
## Intent
This will enable the vscode-ui tests to run on Pull Request jobs in CI.  It includes some changes to how we build docker for now. In a subsequent PR I'll try to do some docker caching.

I also remove the package step since I need to run that in the UI tests to get everything compiled properly. I still have the `dist/` dir saved as an artifact so it's available to download after the run.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Automated Tests

These are them.

## Directions for Reviewers

You can see the tests results here:
https://github.com/posit-dev/publisher/actions/runs/10964651592/job/30448800387#step:13:4660

In case of a failure, the screenshots taken during the run will be saved for 1 day as an artifact (i.e windows-latest-screenshots).  The last screenshot will display the screen where the error occurred.

## Checklist


.github/workflows/vscode-ui.yaml
```diff
@@ -22,22 +22,13 @@ jobs:
...
       - uses: actions/setup-python@v5
         with:
-          python-version: 3.9
-      - uses: quarto-dev/quarto-actions/setup@v2
-        with:
-          version: 1.4.549
+          python-version: ${{ env.PYTHON_VERSION }}
+      # - uses: quarto-dev/quarto-actions/setup@v2
+      #   with:
+      #     version: ${{ env.QUARTO_VERSION }}
```
dotNomad, line 31: Can these be removed?

dotNomad (commented):
I might need some clarity on

> I also remove the package step since I need to run that in the UI tests to get everything compiled properly. I still have the dist/ dir saved as an artifact so it's available to download after the run.

Is the reason why because `vscode-ui` needs the extension to be _not_ packaged prior to running? If that isn't the case why can't we keep the packaging separate from the `vscode-ui` workflow?

.github/workflows/vscode-ui.yaml
kgartland-rstudio: I'll be adding Quarto tests later on so I'm just going to leave that there until it's ready.

kgartland-rstudio:
> I might need some clarity on
> 
> > I also remove the package step since I need to run that in the UI tests to get everything compiled properly. I still have the dist/ dir saved as an artifact so it's available to download after the run.
> 
> Is the reason why because `vscode-ui` needs the extension to be _not_ packaged prior to running? If that isn't the case why can't we keep the packaging separate from the `vscode-ui` workflow?

If we kept the package step, we'd be running `just package` twice. The webdriver vscode service doesn't use the packaged extension, it needs to be pointed at the source code. I didn't dig into what exactly `just package` does but rather than passing around the compiled source code between steps, it made more sense to run `just package` here so I can use it.

dotNomad:
> I didn't dig into what exactly `just package` does but rather than passing around the compiled source code between steps, it made more sense to run `just package` here so I can use it.

It just struck me as odd that the `vscode-ui` workflow was now the uploader of the distributed artifacts to `dist` in the pull request, main, and nightly workflow.

The `package` workflow took care of that.

Is it not possible to keep that isolation of workflow actions? Perhaps I'm missing something on why that isn't possible.

kgartland-rstudio:
I played around with the different options and I think running `just package` in parallel works for now.  We can optimize later.  The most recent run adds 1:30 to the run, so actually not too bad.
https://github.com/posit-dev/publisher/actions/runs/11019941736

kgartland-rstudio:
One last commit. I found that we could unzip the vsix and the point the tests at the extracted directory.  So now we still maintain the `package` step and use the dist artifact from it in the `vscode-ui` test.

dotNomad:
Awesome I'll take another look first thing tomorrow morning. Thanks for figuring this one out.

.github/workflows/main.yaml
```diff
@@ -14,16 +14,12 @@ jobs:
...
-  package:
-    needs: build
-    uses: ./.github/workflows/package.yaml
   archive:
     needs: build
     uses: ./.github/workflows/archive.yaml
   upload:
     needs:
       - archive
-      - package
```
dotNomad, line 26: It looks like the `upload` flow needs `package` to run since it uses the `dist` created.

dotNomad (commented):
This looks very close. I did notice a change in `main.yaml` that I think might break the workflow?

.github/workflows/main.yaml
kgartland-rstudio: Thanks for catching that! I reverted the changes I made to main.yaml since we now are keeping the package step around.

dotNomad (approved):
Thanks for the changes here. Might be a few conflicts with #2333 but should be easy to resolve. If you'd like me to take another look feel free to re-request!

PR #2282, opened 2024-09-17
Upgrade home view webview packages

dotNomad:
Upgrades the packages in the home webview to bump Vite from `5.4.1` to `5.4.6` along with updating other packages using `npm update`

## Intent

Supersedes https://github.com/posit-dev/publisher/pull/2280

PR #2296, opened 2024-09-21
Update npm packages at root dir

dotNomad:
While investigating #2282 I ran `npm update` to up the root level npm packages.

This doesn't adjust the extension or webview packages which look to need some migration.

Prettier was upgraded, but no formatting was needed.


sagerb (approved):
LGTM! Thanks!

PR #2297, opened 2024-09-21
Update npm packages for home view

dotNomad:
Supersedes #2282 and #2302. This PR updates quite a few packages (using `npm update` so keeping major versions the same). This includes updates for Vite, Vue, rollup, and other related packages.

This PR also locks the version of Axios since minor versions have, and currently do break types when upgraded.

A note that after upgrading the packages I started to see a deprecation warning for `sass`

> Deprecation Warning: The legacy JS API is deprecated and will be removed in Dart Sass 2.0.0.
> More info: https://sass-lang.com/d/legacy-js-api

To fix this I followed the recommendation in the Vite documentation to
> use api: "modern-compiler" with `sass-embedded` package.

`sass-embedded` is an alternative to `sass` and instead of being JavaScript is a native Dart executable. So `sass` was replaced with `sass-embedded` and an update to the Vite config was done.

### References

- https://vitejs.dev/config/shared-options.html#css-preprocessoroptions
- https://www.npmjs.com/package/sass-embedded

### DIFF
--- a/extensions/vscode/webviews/homeView/package.json
+++ b/extensions/vscode/webviews/homeView/package.json
@@ -22,7 +22,7 @@
     "@vscode/webview-ui-toolkit": "^1.4.0",
     "@vue/tsconfig": "^0.5.1",
     "jsdom": "^24.1.1",
-    "sass": "^1.72.0",
+    "sass-embedded": "^1.79.3",
     "typescript": "~5.4.2",
     "vite": "^5.1.6",
     "vitest": "^2.0.4",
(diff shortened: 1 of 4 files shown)


extensions/vscode/webviews/homeView/package.json
```diff
@@ -8,7 +8,7 @@
     "test": "TZ=UTC vitest"
   },
   "dependencies": {
-    "axios": "^1.7.4",
+    "axios": "1.7.4",
```
dotNomad, line 11: Axios is misbehaving with types right now. We need to address it, but have kept it separate. I'm changing it to be an exact version here to avoid any minor version changes from hitting us now and in the future.

sagerb (approved):
Changes look good, glad to see the safeguards you put in place on the Axios types.

PR #2307, opened 2024-09-24
Updates to New Deployment Multi-Stepper

### LINKED ISSUE #2200: Add Content Type selector to new config flows where inspection returns "unknown"
This will prevent the creation of `unknown` configurations, that have downstream impacts.

### LINKED ISSUE #2237: Prompt User to confirm scanning directories if number of directories is over some threshold
I was testing the permissions failures yesterday and to do so, I opened my Home directory and then tried to Select an entrypoint.  

The scan (understandably) took nearly two full minutes.  I'm wondering if we should prompt the user if they'd really like us to scan their entire directory structure in cases like these.  Maybe we could do a quick count of directories and it's over x, we prompt the user to confirm.

I did see we got a report from a user doing exactly this, opening a home directory and wanting to scan from there.

>  I also noticed that I need to open a specific folder before trying to create the deployment or it will try and scan my entire home directory.

https://positpbc.slack.com/archives/C05D7NZD52S/p1724761830685139

I wonder if users might think they're going to get to select which directory to scan when clicking `Select...`, rather than scanning every file for an entrypoint without warning.

sagerb:
## Intent

Resolves #2237 
Resolves #2200 

This PR addresses the performance issue with a recursive inspection behavior from VSCode's workspace root, when creating a new deployment. It also solves the problem when the user picks an entrypoint for which our inspection cannot determine the content type.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The first step of the new Deployment multi-stepper now displays a list of the open files that have one of the file extensions we recognize as entrypoints, performs inspection only on these files, and then offers the ability to use VSCode's file open dialog to select any other file. 

If the file selected has several options for deployment content type, a choice between the config types is presented.

If a config is selected with an UNKNOWN content type, a list of content types is presented for the user to tell us what type of project it is.

In another scenario, if a known type of entrypoint file has its publisher button clicked, and there are multiple deployments, then selecting a new deployment from the list will now maintain the file selected as the entrypoint file through the new deployment multi-stepper.

## Automated Tests

The UX tests are being updated to handle these changes.

## Directions for Reviewers

To validate the functionality, you'll need to validate:

1. Clicking the plus sign when there are no files open within VSCode, should start the new Deployment multi-stepper off with just an option to open the file dialog.
-  Selecting a file outside of the VSCode workspace should display an error and then return to the file dialog.
- Clicking cancel from the File Dialog should abort the flow
- Clicking a valid entrypoint should proceed as expected. Use an RMD file to display multiple content types and other valid entrypoints which will not need to prompt for type.
2. Open a valid entrypoint file along with some others. Click the plus sign to confirm that only the valid entrypoint is shown.
3. Create multiple deployments for the entrypoint that you have open. Then click the file publisher button and confirm that you see a selection dialog to chose between the existing deployments or create a new one. Creating a new one should build a new deployment for the entrypoint you triggered the button from, without you having to select it.
4. Create a simple hello-world.py file, which is not currently supported as content. Browse or open the file and initiate a new deployment. Select it and confirm that you are forced to select a content type, and that the configuration file created properly reflects the choices you made. Deployment success/failure does not need to be confirmed.

## Checklist

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -46,388 +46,6 @@ import { showProgress } from "src/utils/progress";
 import { relativeDir, relativePath, vscodeOpenFiles } from "src/utils/files";
 import { ENTRYPOINT_FILE_EXTENSIONS } from "src/constants";
 
-// type stepInfo = {
-//   step: number;
-//   totalSteps: number;
-// };
-
-// type possibleSteps = {
-//   noCredentials: {
-//     singleEntryPoint: stepInfo;
-//     multipleEntryPoints: {
-//       singleContentType: stepInfo;
-//       multipleContentTypes: stepInfo;
-//     };
-//   };
-//   newCredentials: {
-//     singleEntryPoint: stepInfo;
-//     multipleEntryPoints: {
-//       singleContentType: stepInfo;
-//       multipleContentTypes: stepInfo;
-//     };
-//   };
-//   existingCredentials: {
-//     singleEntryPoint: stepInfo;
-//     multipleEntryPoints: {
-//       singleContentType: stepInfo;
-//       multipleContentTypes: stepInfo;
-//     };
-//   };
-// };
-
-// const steps: Record<string, possibleSteps | undefined> = {
-//   inputEntryPointFileSelection: {
-//     noCredentials: {
-//       singleEntryPoint: {
-//         step: 0, // not yet shown
-//         totalSteps: 4,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 1,
-//           totalSteps: 5,
-//         },
-//         multipleContentTypes: {
-//           step: 1,
-//           totalSteps: 6,
-//         },
-//       },
-//     },
-//     newCredentials: {
-//       singleEntryPoint: {
-//         step: 0, // not yet shown
-//         totalSteps: 5,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 1,
-//           totalSteps: 6,
-//         },
-//         multipleContentTypes: {
-//           step: 1,
-//           totalSteps: 7,
-//         },
-//       },
-//     },
-//     existingCredentials: {
-//       singleEntryPoint: {
-//         step: 0, // not yet shown
-//         totalSteps: 2,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 1,
-//           totalSteps: 3,
-//         },
-//         multipleContentTypes: {
-//           step: 1,
-//           totalSteps: 4,
-//         },
-//       },
-//     },
-//   },
-//   inputEntryPointContentTypeSelection: {
-//     noCredentials: {
-//       singleEntryPoint: {
-//         step: 0, // still 0
-//         totalSteps: 4,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 1, // not shown
-//           totalSteps: 5,
-//         },
-//         multipleContentTypes: {
-//           step: 2,
-//           totalSteps: 6,
-//         },
-//       },
-//     },
-//     newCredentials: {
-//       singleEntryPoint: {
-//         step: 0, // still 0
-//         totalSteps: 5,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 1, // not shown
-//           totalSteps: 6,
-//         },
-//         multipleContentTypes: {
-//           step: 2,
-//           totalSteps: 7,
-//         },
-//       },
-//     },
-//     existingCredentials: {
-//       singleEntryPoint: {
-//         step: 0, // still 0
-//         totalSteps: 2,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 1, // not shown
-//           totalSteps: 3,
-//         },
-//         multipleContentTypes: {
-//           step: 2,
-//           totalSteps: 4,
-//         },
-//       },
-//     },
-//   },
-//   inputTitle: {
-//     noCredentials: {
-//       singleEntryPoint: {
-//         step: 1,
-//         totalSteps: 4,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 2,
-//           totalSteps: 5,
-//         },
-//         multipleContentTypes: {
-//           step: 3,
-//           totalSteps: 6,
-//         },
-//       },
-//     },
-//     newCredentials: {
-//       singleEntryPoint: {
-//         step: 1,
-//         totalSteps: 5,
-//       },
-//       multipleEntryPoints: {
-//         singleContentType: {
-//           step: 2,
-//           totalSteps: 6,
-//         },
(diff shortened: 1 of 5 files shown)


kgartland-rstudio:
Updated the tests, looks good on all OS'es.

kgartland-rstudio (approved):
Tested on all OS'es. Looks good!

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -770,125 +350,222 @@ export async function newDeployment(
...
+            defaultUri: baseUri,
+            openLabel: "Select",
+            canSelectFolders: false,
+            canSelectMany: false,
+            title: "Select Entrypoint File (main file for your project)",
+          });
+          if (!fileUris || !fileUris[0]) {
+            // cancelled.
+            return;
+          }
```
dotNomad, line 392: I found it a bit odd that canceling out of the show open dialog cancelled the entire flow. I'm not saying this _needs_ to change, perhaps others have differing opinions. For me I'd prefer if the multi-stepper stayed open and I had to dismiss it with `ESC` as normal.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -770,125 +350,222 @@ export async function newDeployment(
...
+          const fileUri = fileUris[0];
+
+          if (relativeDir(fileUri)) {
+            selectedEntrypointFile = relativePath(fileUri);
+          } else {
+            window.showErrorMessage(
+              `Entrypoint files must be located within the folder opened by VSCode. 
+							File ${fileUri.fsPath} is not located within the VSCode Workspace: ${baseUri.fsPath}.`,
```
dotNomad, line 400: We probably want to avoid VSCode specific language since this can be used in Positron. Both use "workspace".

```suggestion
              `Entrypoint files must be located within the open workspace. 
							File ${fileUri.fsPath} is not located within the open workspace: ${baseUri.fsPath}.`,
```

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -770,125 +350,222 @@ export async function newDeployment(
...
+    if (
+      newDeploymentData.entrypoint.inspectionResult.configuration.type ===
+      ContentType.UNKNOWN
+    ) {
+      // have to prompt user for a content type, since we were unable to determine it.
+      const quickPicks: QuickPickItemWithIndex[] = [];
+      allValidContentTypes.forEach((contentType, index) => {
+        quickPicks.push({
+          label: contentType,
+          detail: contentTypeStrings[contentType],
```
dotNomad, line 512: I found this style to be a bit large given the strings we use. I believe a swap from `detail` to `description` would be a bit more dense, and easier to parse:

CleanShot 2024-09-25 at 15 12 39@2x

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -588,88 +186,80 @@ export async function newDeployment(
...
+          const relEntryPointDir = path.dirname(openFile);
+          const relEntryPointFile = path.basename(openFile);
+
+          const inspectResponse = await api.configurations.inspect(
+            relEntryPointDir,
+            python,
+            {
+              entrypoint: relEntryPointFile,
+            },
+          );
```
dotNomad, line 237: I thought we only did inspection once the file was selected. With the open file list now I see the same file for each content type is available to it, which I don't think is what we want:

CleanShot 2024-09-25 at 15 17 39@2x

dotNomad: Just double checked, if I select one of the `static.Rmd` files above I still get the content type question

dotNomad (changes requested):
I am able to get an Unexpected type guard failure if I drill into the steps (I went to credential creation), head back up, click "Open..." to do a file selection then click cancel.

---

Noticed a few things in the comments. A handful of nitpicks, design questions, or cleanup recommendations, but there are a few things we should address.

sagerb:
After discussion with @dotNomad , I agree that some rework will be needed on this PR. We'd like to return the flow back to where only the file is shown, with then inspections only done on the file selected. Returning the issue back to "in-progress".

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -458,113 +105,64 @@ export async function newDeployment(
       return false;
```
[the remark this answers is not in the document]
sagerb: Great catch. Cleaned up.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -588,88 +186,80 @@ export async function newDeployment(
...
-      if (entrypointFilesUnopened.length) {
-        entryPointListItems.push({
-          label: "Discovered Entrypoints",
-          kind: QuickPickItemKind.Separator,
-        });
-        entrypointFilesUnopened.forEach((entryPointFile) => {
-          entryPointListItems.push({
-            iconPath: new ThemeIcon("file"),
-            label: entryPointFile,
+    });
```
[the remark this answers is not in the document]
sagerb: updated.

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: Per our discussion, this has been refactored to return to the behavior you described.

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: Fair enough. This has been updated.

extensions/vscode/src/multiStepInputs/newDeployment.ts
sagerb: updated.

sagerb:
> I am able to get an Unexpected type guard failure if I drill into the steps (I went to credential creation), head back up, click "Open..." to do a file selection then click cancel.

Missed this one, checking...

sagerb:
> > I am able to get an Unexpected type guard failure if I drill into the steps (I went to credential creation), head back up, click "Open..." to do a file selection then click cancel.
> 
> Missed this one, checking...

Fixed in latest commit.

dotNomad (approved):
Thanks for those additional changes. I wasn't able to re-create any of the bugs I encountered, and the small changes added up to make for a better flow too. :shipit:

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -758,137 +303,226 @@ export async function newDeployment(
...
     } else {
-      state.data.entryPoint = inspectionQuickPicks[0];
+      newDeploymentData.entrypoint.inspectionResult =
+        inspectionQuickPicks[0].inspectionResult;
       // We're skipping this step, so we must silently just jump to the next step
       return inputTitle(input, state);
     }
   }
 
+  // // ***************************************************************
```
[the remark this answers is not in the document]
sagerb: Yes, we should. I'll remove it. Thanks!

PR #2308, opened 2024-09-24
File credentials service (unplugged at the moment)

marcosnav:
## Intent

Incremental work pushed for #1831 

This PR adds a new credentials service and it's constructor, which will handle cases where a system keyring service is unavailable.

- `fileCredentialsService`
- `NewFileCredentialsService(log logging.Logger) (*fileCredentialsService, error)`

This new file credentials service is unplugged in this PR, meant to facilitate incremental review of the file credentials work.

A subsequent PR will be created pointing to this one (at least until this one is merged).

Main points of this new file credentials service:
- Has a setup step that runs within the constructor to verify if the `.connect-credentials` file exists, creates it when needed.
- Implements the methods of the existing credentials service that uses the keyring provider. To allow plugging it in and be swapped when needed. (More on this on the sub-sequent PR)
```
Set(name, url, ak string) (*Credential, error)
Get(guid string) (*Credential, error)
List() ([]Credential, error)
Delete(guid string) error
```

Unit tests of this service serve as an example on how the `.connect-credentials` file will updated, while actually testing read and write actions to a TOML format file.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Unit tests for this new service included

## Directions for Reviewers

No manual actions needed at this moment, unit tests verify that the file service can read and write credentials to the provided file.

## Checklist

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -53,6 +53,16 @@ type Credential struct {
 
 type CredentialV0 = Credential
 
+func (c *Credential) ConflictCheck(compareWith Credential) error {
+	if compareWith.URL == c.URL {
+		return NewURLCollisionError(c.Name, c.URL)
+	}
+	if compareWith.Name == c.Name {
+		return NewNameCollisionError(c.Name, c.URL)
+	}
+	return nil
+}
+
 type CredentialRecord struct {
 	GUID    string          `json:"guid"`
 	Version uint            `json:"version"`
@@ -95,7 +105,7 @@ func NewCredentialsService(log logging.Logger) *credentialsService {
 // FileCredentialRecordFactory creates a Credential based on the presence of the
 // a file containing CONNECT_SERVER and CONNECT_API_KEY.
 
-type fileCredential struct {
+type deprcfileCredential struct {
 	URL string `toml:"url"`
 	Key string `toml:"key"`
 }
@@ -113,7 +123,7 @@ func (cs *credentialsService) FileCredentialRecordFactory() (*CredentialRecord, 
 	if !exists {
 		return nil, nil
 	}
-	var credential fileCredential
+	var credential deprcfileCredential
 	err = util.ReadTOMLFile(filePath, &credential)
 	if err != nil {
 		return nil, err
--- a/internal/credentials/errors.go
+++ b/internal/credentials/errors.go
@@ -8,6 +8,10 @@ type CorruptedError struct {
 	GUID string
 }
 
+func NewCorruptedError(guid string) *CorruptedError {
+	return &CorruptedError{guid}
+}
+
 func (e *CorruptedError) Error() string {
 	return fmt.Sprintf("credential '%s' is corrupted", e.GUID)
 }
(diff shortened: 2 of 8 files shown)


internal/credentials/credentials.go
```diff
@@ -95,7 +105,7 @@ func NewCredentialsService(log logging.Logger) *credentialsService {
 // FileCredentialRecordFactory creates a Credential based on the presence of the
 // a file containing CONNECT_SERVER and CONNECT_API_KEY.
 
-type fileCredential struct {
+type deprcfileCredential struct {
```
marcosnav, line 108: Marking the current `fileCredential` struct as deprecated, will be removed in the next PR and be replaced by `file.go#fileCredential`

internal/credentials/errors.go
```diff
@@ -44,6 +60,10 @@ type EnvURLCollisionError struct {
 	URL  string
 }
 
+func NewEnvURLCollisionError(name, url string) *EnvURLCollisionError {
```
marcosnav, line 63: EnvVar credentials related errors to be wiped out on sub-sequent PR, as we talked about removing breadcrumbs of what once was support for environment credentials

internal/credentials/file.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// Set home dir credentials file path
+	homeDir, err := util.UserHomeDir(fservice.afs)
+	if err != nil {
+		return nil, err
+	}
+	fservice.credsFilepath = homeDir.Join(ondiskFilename)
+
+	// Verify file can be modified, will create if not exists
+	err = fservice.setup()
```
sagerb, line 87: In general, I prefer the approach that we don't create or validate the file until a request comes in which requires it. That way, the error of it not being able to be accessed is deferred until a requested action, which can then receive the appropriate error. When doing it within the constructor, it is questionable of who is going to see the error and why an empty file is created with no actions.

My requested change is to simply move the `fservice.setup()` method into each of the API request handlers as a first step, returning any errors within those responses.

Please let me know what you think, perhaps I'm missing an advantage for your approach over my recommendation. Thanks!

sagerb (changes requested):
Just have the one requested change and I'll be interested in your thoughts. Everything else looks good. :-)

internal/credentials/file.go
marcosnav: Thanks Bill! This is happening just like you mention, the credentials services constructors run only when endpoints receive requests. Any error that could come up will always be sent and received within an API response.

This PR only has the service introduction, but in the next PR you'll be able to see that such constructors are imported and used only within endpoint handlers :)

dotNomad (approved):
Fantastic setup, splitting the PR was a great idea, the code is extremely readable, and thoroughly tested 💯

internal/credentials/file.go
sagerb: Thanks Marcos! I think my mind was also raising some concerns about this being a construction method rather than a factory method, so I was not really wanting to return errors from it, but after your explanation and a second look, I thank you very much for your response. I'm good with all of this, I'm approving.

sagerb (approved):
Concerns have been resolved. Thanks!

internal/credentials/file.go
dotNomad: Just for clarity to ensure I'm understanding this correctly this will create `.connect-credentials` on machines that do not have keyrings where our extension is loaded and opened since we call `GET /api/credentials` to fetch the Credentials on the machine. Am I correct?

marcosnav: Yes, that's right 👍

PR #2312, opened 2024-09-24
Add POST configuration secrets endpoint

dotNomad:
This PR adds a new API endpoint `POST /api/configurations/$NAME/secrets` that:
- takes an action and secret name similar to the POST files API
- returns a `configDTO` containing the `projectDir`, `configuration`, and `configLocation` for easy plug into our frontend store

## Intent

Part of #2305

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Directions for Reviewers

Test out the new endpoint:
- testing the add and remove action
- testing in and out of multi-directory roots

`http POST :9001/api/configurations/fastapi-simple-QSPH/secrets action=add secret=new`
`http POST :9001/api/configurations/fastapi-simple-QSPH/secrets action=remove secret=new`
`http POST :9001/api/configurations/fastapi-simple-QSPH/secrets action=remove secret=new dir==fastapi-simple`
`http POST :9001/api/configurations/fastapi-simple-QSPH/secrets action=add secret=new dir==fastapi-simple`

### DIFF
--- a/internal/services/api/post_config_secrets.go
+++ b/internal/services/api/post_config_secrets.go
@@ -1,0 +1,125 @@
+package api
+
+// Copyright (C) 2024 by Posit Software, PBC.
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/posit-dev/publisher/internal/config"
+	"github.com/posit-dev/publisher/internal/logging"
+	"github.com/posit-dev/publisher/internal/types"
+	"github.com/posit-dev/publisher/internal/util"
+)
+
+const (
+	secretActionAdd    = "add"
+	secretActionRemove = "remove"
+)
+
+type postConfigSecretsRequest struct {
+	Action string `json:"action"`
+	Secret string `json:"secret"`
+}
+
+func applySecretAction(cfg *config.Config, action string, secret string) error {
+	switch action {
+	case secretActionAdd:
+		// Check if the secret already exists before adding
+		for _, s := range cfg.Secrets {
+			if s == secret {
+				return nil // Secret already exists, no need to add
+			}
+		}
+		cfg.Secrets = append(cfg.Secrets, secret)
+	case secretActionRemove:
+		for i, s := range cfg.Secrets {
+			if s == secret {
+				cfg.Secrets = append(cfg.Secrets[:i], cfg.Secrets[i+1:]...)
+				break
+			}
+		}
+	default:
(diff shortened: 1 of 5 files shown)


internal/services/api/post_config_secrets.go
```diff
@@ -0,0 +1,125 @@
...
+	Secret string `json:"secret"`
+}
+
+func applySecretAction(cfg *config.Config, action string, secret string) error {
+	switch action {
+	case secretActionAdd:
+		// Check if the secret already exists before adding
+		for _, s := range cfg.Secrets {
+			if s == secret {
+				return nil // Secret already exists, no need to add
```
sagerb, line 35: So if duplicate, no error? Are we sure that is the desired behavior?

sagerb (approved):
Code looks good to me.

NOTE: I did not attempt to verify the functionality (reviewing code on my phone).

internal/services/api/post_config_secrets.go
dotNomad: I figured since Secrets should be a set, not containing duplicates, that the add should be idempotent. I can definitely change that if there is disagreement.

internal/services/api/post_config_secrets.go
```diff
@@ -0,0 +1,125 @@
...
+	secretActionAdd    = "add"
+	secretActionRemove = "remove"
+)
+
+type postConfigSecretsRequest struct {
+	Action string `json:"action"`
+	Secret string `json:"secret"`
+}
+
+func applySecretAction(cfg *config.Config, action string, secret string) error {
```
marcosnav, line 29: I know that this new endpoint is highly based on `post_config_files.go`. Sharing here some suggestions though, it is a good practice to keep API handlers lean and only consume service providers or instances that handle the main process. Meaning, an ideal flow would be
- Enpdoint handler gets called with some payload
- Payload is validated
- Payload is handled over to some service which is the responsible for the main action
- Endpoint handler returns the result

With that said, and if you are up to, you could create a new method into the `Config` struct responsible for all of these operations. IMO the config package is better suited to house this func. So your endpoint down below would look something like:
```go
    err = cfg.ApplySecretAction(b.Action, b.Secret)
    if err != nil {
        BadRequest(w, req, log, err)
        return
    }
```

Also, actions could be constants from the config package
```go
// config.go
const SecretActionAdd = "add"
const SecretActionRemove = "remove"
```

marcosnav (approved):
Looks good in general, would love to see some code moved over to the config package. Nothing to block this PR to land though.

PR #2314, opened 2024-09-25
File Credentials Support

marcosnav:
## Intent

Full support for credentials persistence via a file `~/.connect-credentials`.

Following up the work done at #2308, this PR completes the feature by:
- Having a main credentials service constructor `NewCredentialsService`
- `NewCredentialsService` is called by endpoints that do credentials operations
- `NewCredentialsService` tries to:
    -  set up the default system's keyring service, `NewKeringCredentialsService`
    - falling back to file based credentials when there is no keyring available. `NewFileCredentialsService`
- This allows for different funtionality on credentials persistence to share the same interface for credentials operations `List(), Get(), Set(), Delete()`

**Screenshots of Workbench using the file backed credentials service**
Screen Shot 2024-09-25 at 15 05 32

Screen Shot 2024-09-25 at 15 11 30

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Unit tests created and updated

## Directions for Reviewers

To manually verify this changes: 
- Start a new workbench session,
- Install the compatible packaged VSIX file, generated by CI on this PR's job
- Reload session if needed, from the command pallette `> Developer: Reload Window`
- Create and delete credentials as desired
- Remove workbench sessions, see how credentials are persisted next time you start a new session

## Checklist

### DIFF
--- a/internal/credentials/credentials_test.go
+++ b/internal/credentials/credentials_test.go
@@ -3,270 +3,124 @@
 package credentials
 
 import (
+	"errors"
 	"testing"
 
-	"github.com/pelletier/go-toml/v2"
 	"github.com/posit-dev/publisher/internal/logging/loggingtest"
-	"github.com/posit-dev/publisher/internal/util"
 	"github.com/posit-dev/publisher/internal/util/utiltest"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/suite"
 	"github.com/zalando/go-keyring"
 )
 
-type Cred struct {
-	GUID   string
-	Server string
-	ApiKey string
-	Name   string
-}
-
-var fileCred = Credential{
-	GUID:   "00000000-0000-0000-0000-000000000000",
-	URL:    "https://connect.localtest.me/rsc/dev-password-copy",
-	ApiKey: "3456789",
-	Name:   "connect.localtest.me",
-}
-
-type CredentialsTestSuite struct {
+type CredentialsServiceTestSuite struct {
 	utiltest.Suite
 	log *loggingtest.MockLogger
 }
 
-func TestCredentialsTestSuite(t *testing.T) {
-	suite.Run(t, new(CredentialsTestSuite))
+func TestCredentialsServiceTestSuite(t *testing.T) {
+	suite.Run(t, new(CredentialsServiceTestSuite))
 }
 
-func (s *CredentialsTestSuite) SetupTest() {
+func (s *CredentialsServiceTestSuite) SetupTest() {
 	s.log = loggingtest.NewMockLogger()
 }
 
-func (s *CredentialsTestSuite) credentialsFilePath(afs afero.Fs) (util.AbsolutePath, error) {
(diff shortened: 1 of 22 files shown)


internal/credentials/keyring.go
```diff
@@ -3,44 +3,177 @@
...
+	if !exists {
+		ks.log.Debug("Credential does not exist", "credential", guid)
+		return nil, NewNotFoundError(guid)
+	}
+
+	return cr.ToCredential()
+}
+
+// List retrieves all Credentials
+func (ks *keyringCredentialsService) List() ([]Credential, error) {
```
dotNomad, line 69: Noticed that the `keyring_test.go` file didn't have a test for this function. Maybe handy to add a test ensuring that an empty array is returned instead of `nil` to avoid the bug we have seen in other places: https://github.com/posit-dev/publisher/issues/1937

internal/credentials/credentials_test.go
```diff
@@ -3,270 +3,124 @@
...
-	}
+func (s *CredentialsServiceTestSuite) TestNewCredentialsService_KeyringErrFallbackFile() {
+	// Use an in memory filesystem for this test
+	// avoiding to manipulate users ~/.connect-credentials
+	fsys = afero.NewMemMapFs()
+	defer func() { fsys = afero.NewOsFs() }()
 
-	// Add credentials into environment
-	s.createFileCredentials(&cs, true)
+	keyringErr := errors.New("this is a teapot, unsupported system")
```
dotNomad, line 117: Classic 😆

dotNomad (approved):
Verified that the changes here work for add and deletion of credentials locally and on Workbench. Great work! 💯

internal/credentials/keyring.go
marcosnav: Good catch, pushed a new commit adding a test for it 👍

dotNomad (approved):
Thanks for adding the test, this is looking good!

sagerb (approved):
Code looks great!

The only nit I have is I'm wondering about the name of the credentials file, and if it should be a little more "name-spaced"... Something like `.posit-connect-credentials`. It also occurred to me that to be friendlier to corporate interests we could establish `.posit-credentials` as a file, with connect being a section or type on a section. Another option is to name it as `.posit-connect` and establish it for some other future needs outside of just credentials.

Not trying to incorrectly invest in a future feature we do not yet know about, but I think the  naming is important here (and hard). My favorite is probably `.posit-connect`.

dotNomad:
> if it should be a little more "name-spaced"... Something like `.posit-connect-credentials`

I've had this same question. I think it is fine to get this merged in and have a separate discussion around it since that is distinct from the implementation here.

> My favorite is probably `.posit-connect`

To mimic our `.posit` directory in projects we could do `.posit/.connect-credentials`, but like I said I think we should have this as a follow-up discussion.

PR #2315, opened 2024-09-25
Add contextual notification alert icons to section headers

### LINKED ISSUE #2245: Add action indicators to section headers to indicate that attention is needed
- Add ability to show icons in the TreeSection headers
- Show warning icons in headers when that section has a warning
    - Like a missing package file

sagerb:
…ion when alerting

## Intent

This PR adds action indicators to section headers to indicate that attention is needed.

resolves #2245 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Business logic that determined the display of actionable messages was moved into the central store for `EvenEasierDeploy.vue`, `Credentials.vue`, `RPackages.vue` and `PythonPackages.vue`. Existing logic was maintained, which was easy since each of these methods were already using the store to perform their computations.

`TreeSection.vue` component was updated to allow the specification of a `headerActions` prop, which is shown when the section is collapsed (vs. the `actions` prop which is only shown when the section is expanded. Also had to support an `expanded` prop, to allow the actions (who's logic is implemented in the parent component) to be able to expand the section.

Screenshot:
2024-09-25 at 11 05 AM

## Directions for Reviewers

Each of the updated sections (credentials, python packages, and r packages) will show the alert icon when collapsed under the same conditions as they would display their "welcome" messages when expanded.

To reproduce some of these:
- for a python project:
    - remove your `requirements.txt` file
    - empty the `requirements.txt` file
- for an R project:
    - remove your `renv.lock` file 
    - empty your `renv.lock` file

## Checklist

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
+++ b/extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
@@ -11,15 +11,21 @@
           class="twisty-container codicon"
           :class="expanded ? 'codicon-chevron-down' : 'codicon-chevron-right'"
         />
-        <h3 class="title">{{ title }}</h3>
-        <span v-if="$slots.description" class="description">
-          <slot name="description" />
-        </span>
-        <span v-else-if="description" class="description">
-          {{ description }}
-        </span>
+        <div style="display: flex; justify-content: space-between">
+          <h3 class="title">{{ title }}</h3>
+          <span v-if="$slots.description" class="description">
+            <slot name="description" />
+          </span>
+          <span v-else-if="description" class="description">
+            {{ description }}
+          </span>
+          <div
+            v-if="codicon && !expanded"
+            class="tree-section-icon codicon"
+            :class="codicon"
+          />
+        </div>
       </div>
-      <div v-if="codicon" class="tree-section-icon codicon" :class="codicon" />
       <div v-if="actions" class="actions">
         <ActionToolbar :title="title" :actions="actions" />
       </div>
@@ -62,17 +68,11 @@ const toggleExpanded = () => {
   .tree-section-icon {
     font-size: 13px;
     margin-right: 4px;
-    display: initial;
   }
 
   &.expanded:hover .actions,
   &.expanded:focus-within .actions {
     display: initial;
-  }
-
-  &.expanded .tree-section-icon {
-    display: none;
-    margin-left: auto;
   }
 
   &.expanded:hover
--- a/extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
+++ b/extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
@@ -11,20 +11,18 @@
           class="twisty-container codicon"
           :class="expanded ? 'codicon-chevron-down' : 'codicon-chevron-right'"
         />
-        <div style="display: flex; justify-content: space-between">
-          <h3 class="title">{{ title }}</h3>
-          <span v-if="$slots.description" class="description">
-            <slot name="description" />
-          </span>
-          <span v-else-if="description" class="description">
-            {{ description }}
-          </span>
-          <div
-            v-if="codicon && !expanded"
-            class="tree-section-icon codicon"
-            :class="codicon"
-          />
-        </div>
+        <h3 class="title">{{ title }}</h3>
+        <span v-if="$slots.description" class="description">
+          <slot name="description" />
+        </span>
+        <span v-else-if="description" class="description">
+          {{ description }}
+        </span>
+        <div
+          v-if="codicon && !expanded"
+          class="tree-section-icon codicon"
+          :class="codicon"
+        />
       </div>
       <div v-if="actions" class="actions">
         <ActionToolbar :title="title" :actions="actions" />
@@ -68,6 +66,7 @@ const toggleExpanded = () => {
   .tree-section-icon {
     font-size: 13px;
     margin-right: 4px;
+    margin-left: auto;
   }
 
   &.expanded:hover .actions,
--- a/extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
+++ b/extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
@@ -27,6 +27,9 @@
       <div v-if="actions" class="actions">
         <ActionToolbar :title="title" :actions="actions" />
       </div>
+      <div v-if="headerActions" class="header-actions">
+        <ActionToolbar :title="title" :actions="headerActions" />
+      </div>
       <div v-if="count" class="count">
         <CountBadge :count="count" />
       </div>
@@ -40,20 +43,27 @@
 <script setup lang="ts">
 import ActionToolbar, { ActionButton } from "src/components/ActionToolbar.vue";
 import CountBadge from "src/components/CountBadge.vue";
+import { watch } from "vue";
 
 const expanded = defineModel("expanded", { required: false, default: false });
 
-defineProps<{
+const props = defineProps<{
   title: string;
   description?: string;
   codicon?: string;
+  headerActions?: ActionButton[];
   actions?: ActionButton[];
   count?: number;
+  expanded?: boolean;
 }>();
 
 const toggleExpanded = () => {
   expanded.value = !expanded.value;
 };
+
+watch([props.expanded], () => {
+  expanded.value = props.expanded;
+});
 </script>
 
 <style lang="scss" scoped>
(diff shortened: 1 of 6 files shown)


extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
```diff
@@ -27,6 +27,9 @@
       <div v-if="actions" class="actions">
         <ActionToolbar :title="title" :actions="actions" />
       </div>
+      <div v-if="headerActions" class="header-actions">
+        <ActionToolbar :title="title" :actions="headerActions" />
+      </div>
```
dotNomad, line 32: I do not think this is the right approach to showing that a section has an error. I would have expected to use the `codicon` prop to pass in `codicon-alert`. The reason for this:
- this is not an action, clicking on it doesn't do anything that the header doesn't already do (expand)
- the icon goes away when expanded removing the glanceability of the icon showing that there is an error

If we do want to go with this method we need to remove the `expanded` prop and instead use the available `v-model` for `expanded`. This will prevent the bug currently present where the button stops working after the initial expansion.

dotNomad (changes requested):
Added my concerns in a comment, we can discuss to come up with an approach.

extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
sagerb: I started with a codicon, but that puts it as a left-hand prefix to the label,  and I wanted it at the right side. Reaching out to chat.

sagerb:
Code has been modified to use the header's codicon, but move it to the right-hand side, as well as hide on expansion, as we discussed.

I think this should be ready for your re-review, with all concerns addressed. Thanks!

extensions/vscode/webviews/homeView/src/components/views/Credentials.vue
```diff
@@ -3,9 +3,12 @@
     title="Credentials"
     data-automation="credentials"
     :actions="sectionActions"
+    :codicon="home.credentialsAlert ? `codicon-alert` : ``"
```
dotNomad, line 6: I'm not sure the having no Credentials should be an alert. This makes it so the very first experience, without a Deployment, shows an alert to the user.

CleanShot 2024-09-26 at 14 37 22@2x

If we are in the
> A Credential for the Deployment's server URL was not found

state then I think that is valid.

extensions/vscode/webviews/homeView/src/components/views/Credentials.vue
```diff
@@ -3,9 +3,12 @@
...
   >
-    <WelcomeView v-if="home.sortedCredentials.length === 0">
-      <p>No credentials have been added yet.</p>
+    <WelcomeView v-if="showWelcomeView">
```
dotNomad, line 8: Any reason to now use `home.alertNoCredentials` directly?

extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
```diff
@@ -24,6 +19,7 @@
           {{ description }}
         </span>
       </div>
+      <div v-if="codicon" class="tree-section-icon codicon" :class="codicon" />
```
dotNomad, line 22: I think we actually still want this in the `pane-header-title-container` so the click event on the entire header triggers the expansion. With the current position of this in the DOM if you click the alert it will not open the section.

extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
```diff
@@ -63,11 +59,22 @@ const toggleExpanded = () => {
...
+
   &.expanded:hover .actions,
   &.expanded:focus-within .actions {
     display: initial;
   }
 
+  &.expanded .tree-section-icon {
+    display: none;
+    margin-left: auto;
+  }
```
dotNomad, line 76: Personally since `expanded` is available to us I think we should not render the codicon rather than use CSS to not display it.

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -232,6 +232,111 @@ export const useHomeStore = defineStore("home", () => {
...
+  const alertPythonEmptyRequirements = computed(() => {
+    return (
+      pythonProject.value &&
+      pythonPackageFile.value &&
+      pythonPackages.value &&
+      pythonPackages.value.length === 0
+    );
+  });
+
+  const alertPythonMissingRequirements = computed(() => {
```
dotNomad, line 249: I found this name a bit confusing when I was reading it in the component.

I'd suggestion

`isMissingPythonRequirementsFile`

or even better if we could group the python alerts in an object we could do

`home.python.isMissingRequirementsFile`

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -232,6 +232,111 @@ export const useHomeStore = defineStore("home", () => {
       .length;
   });
 
+  // Not considered an alert at this time.
+  const isNotPythonProject = computed(() => {
+    return !pythonProject.value;
+  });
+
+  const alertPythonEmptyRequirements = computed(() => {
```
dotNomad, line 240: The `alert` prefix on a few of these I found a bit strange. I think it makes sense for the aggregate ones

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -232,6 +232,111 @@ export const useHomeStore = defineStore("home", () => {
...
+      pythonPackages.value &&
+      pythonPackages.value.length === 0
+    );
+  });
+
+  const alertPythonMissingRequirements = computed(() => {
+    return pythonProject.value && !pythonPackageFile.value;
+  });
+
+  const pythonAlert = computed(() => {
```
dotNomad, line 253: I always like the `is`, `has`, etc prefixes for boolean variables. I'd suggest `hasPythonAlert` in this case. Or if we go with the object grouping we could do `python.hasAlert`

dotNomad (commented):
Had a few comments.

Overall the move the home store looks great. Some of the naming on the functions felt a bit off or hard to parse for me.

I did have an idea for a bit of organization in the home store. We could group the functions for the Python and R sections. So usage would look like:
- `home.r.packages`
- `home.python.packageFile`

extensions/vscode/webviews/homeView/src/stores/home.ts
sagerb: I've modified all of the method names / organization that I've added to the store.

sagerb:
>I did have an idea for a bit of organization in the home store.

I've organized them into objects, as you suggested. For now, they are in the applicable hierarchies of `config.active.*`, `credential.*`, `r.active.*`, and `python.active.*`. This allows us to move the other functionality available in the home store into these objects, using the top-level hierarchy for general or collective concepts and the functionality regarding the selected values into `.active`.

extensions/vscode/webviews/homeView/src/stores/home.ts
sagerb: updated.

extensions/vscode/webviews/homeView/src/stores/home.ts
sagerb: refactored with the re-organization with the store.

extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
sagerb: This one I need to leverage your CSS expertise. I want the icon to be forced to the right, as if we were in a `flex`, with `space-between`, but as I place it back into the `pane-header-title-container`, I haven't been able to figure out the necessary changes to make it work, while maintaining the existing functionality. So I've pushed it up there, but currently it simply displays at the end of the title.

sagerb:
@dotNomad Requests/concerns have been addressed in the latest commits. If you have suggestion as to how to design the CSS around the notification icon, please let me know.

extensions/vscode/webviews/homeView/src/components/tree/TreeSection.vue
dotNomad: Adjusted the CSS and HTML in https://github.com/posit-dev/publisher/pull/2315/commits/a2fc8c08e10461e2cba36a29db439c8c39585a63 for the `tree-section-icon`

Instead of creating a flex container I just used `margin-left: auto` to ensure that the codicon was always on the far left.

This does a few things:
- prevents us having to continue passing down the `min-width: 0` trick to have the header overflow with `text-overflow: ellipsis`
- since the icon wouldn't flex anyway `margin-left: auto` provides the same end result
- we keep the same rendering mode so we don't have to re-align the icon in the flex container

Here is an example of it working with a long title

CleanShot 2024-09-27 at 15 00 36@2x

dotNomad (approved):
Thanks for all of those changes. I think this is looking great! Let me take a look at the PR this is based on so we can get these both merged up.

PR #2324, opened 2024-09-27
Suppress calls to agent when active config is in error

### LINKED ISSUE #2323: Refresh File List API errors being displayed when editing and saving an invalid active configuration file.
API calls which depend on a valid configuration file are being called even if the configuration file is in error, which causes API failure notices to be displayed in VSCode.

The following example is after I added an `[r]` section to the configuration file and saved the file.

Image

This particular issue would be resolved if we suppressed the API calls which require a valid configuration are suppressed if we know it is error.

Also notice that the configuration dependent sections are visible when the active configuration is in error - which is not reliable (since the backend can't read / parse the invalid configuration file).

sagerb:
## Intent

This PR resolves the issue where API calls that depend on a valid configuration file are being called even if the configuration file is in error, which causes API failure notices to be displayed in VSCode.

Resolves #2323 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

There are a number of API calls that we know depend upon a valid configuration file (as we are required to pass in the name of the active configuration file). I have added checks (where missing) into the methods in which we get the active configuration file and return immediately if the active configuration is in error. This removes the reported error message, which was occurring off of file watcher actions.

I have also updated the homeView to not display the configuration-reliant sections (files, python, r) when the selected configuration is in error.

## Directions for Reviewers

Edit the active configuration to have an invalid TOML field. Saving the file should not cause a 500 error to be displayed in the lower-right corner of VSCode. Confirm the sections `files`, `r packages` and `python packages` are hidden after save.

Make other changes directly to the configuration which are valid, and upon save, you should see the appropriate updates to the UX where previously expected. Confirm the sections `files`, `r packages` and `python packages` are visible after save.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -338,7 +338,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       return;
     }
     if (isConfigurationError(activeConfig)) {
-      console.log(
+      console.error(
         "homeView::updateFileList: Skipping - error in active configuration.",
       );
       return;
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -337,6 +337,12 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       console.error("homeView::updateFileList: No active configuration.");
       return;
     }
+    if (isConfigurationError(activeConfig)) {
+      console.log(
+        "homeView::updateFileList: Skipping - error in active configuration.",
+      );
+      return;
+    }
     try {
       await showProgress("Updating File List", Views.HomeView, async () => {
         const api = await useApi();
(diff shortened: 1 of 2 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -337,6 +337,12 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       console.error("homeView::updateFileList: No active configuration.");
       return;
     }
+    if (isConfigurationError(activeConfig)) {
+      console.log(
```
marcosnav, line 341: nit: `console.log` -> `console.error` like the others

marcosnav (approved):
LGTM

dotNomad (approved):
Nice changes. Agree with @marcosnav's point for consistency

PR #2331, opened 2024-09-30
Display configuration error w/ TOML parsing details

### LINKED ISSUE #2330: Display of Error when configuration file has an invalid TOML error
With a fix in place for #2323, an invalid configuration file is shown to be in error, but there is no information presented which will help the user correct the issue. 

We have the error being sent back as part of a Configuration Error. We should be displaying it to the user.

sagerb:
## Intent

The changes present within this PR will display the error details present within `ConfigurationError`s. Previously, the details were not shown and only a generic message was provided to the user, with no details on what to fix. A great example of this was when a TOML parsing error was returned within a ConfigurationError object. The AgentError includes specific error data for this type of error (invalidTOML).

Resolves #2330 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The UX currently detects some errors associated with the configuration file. I first extracted those out into a new `ConfigErrors` component, with each of the existing being subcomponents to display their own details. 

I then defined a new type declaration for the Agent Error which contained the invalidTOML error, and created a subcomponent to display within the `ConfigErrors` component.

I also added a generic subcomponent that displays the information present within the AgentError, if the error had not been handled by another subcomponent.

Screenshots:

Existing Configuration Errors
- Missing Configuration Error:
2024-09-30 at 4 27 PM

- No configuration file entry in deployment file
2024-09-30 at 4 30 PM

- Invalid TOML Error - Parsing error reported
(added abc to file)
2024-09-30 at 4 32 PM

- Invalid TOML Error (actually UnknownTOMLKey from agent, but handled as invalidTOML error)
(changed `entrypoint =` to `entrypoint2 =`)
2024-10-01 at 9 03 AM

- Invalid TOML Error
(Changed `files = []` to `files = "abc"`)
2024-10-01 at 9 08 AM

- Invalid TOML Error
(Deleted value after = sign)
2024-10-01 at 9 22 AM

- Invalid TOML Error
(Removed last quote for key)
2024-10-01 at 9 23 AM

- Generic Handled Error - Error not returned as specific error from agent
(duplicated one key to another).
2024-10-01 at 9 54 AM

## Directions for Reviewers

Conditions for reproducing the errors above are listed. Simply recreate them and validate that you see the updated UX.

## Checklist

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/configErrors/ConfigErrors.vue
+++ b/extensions/vscode/webviews/homeView/src/components/configErrors/ConfigErrors.vue
@@ -1,0 +1,100 @@
+<!-- Copyright (C) 2024 by Posit Software, PBC. -->
+
+<template>
+  <EntryMissing
+    v-if="ifEntryMissingError"
+    :config-selection-prompt="promptForConfigSelection"
+    @select-configuration="selectConfiguration"
+  />
+  <FileMissing
+    v-if="ifFileMissingError"
+    :config-selection-prompt="promptForConfigSelection"
+    @select-configuration="selectConfiguration"
+  />
+  <InvalidTOMLFile
+    v-if="ifInvalidTOMLFileError"
+    :error="invalidTOMLFileError"
+    @edit-active-configuration="onEditActiveConfiguration"
+  />
+  <UnhandledError v-if="ifUnhandledError" :error="configError" />
+</template>
+<script setup lang="ts">
+import { useHostConduitService } from "src/HostConduitService";
+import { useHomeStore } from "src/stores/home";
+import { computed } from "vue";
+import { WebviewToHostMessageType } from "../../../../../src/types/messages/webviewToHostMessages";
+import { isConfigurationError } from "../../../../../src/api";
+import {
+  AgentError,
+  InvalidTOMLFileCodeError,
+  isInvalidTOMLFileCode,
+} from "../../../../../src/api/types/error";
+
+import InvalidTOMLFile from "./InvalidTOMLFile.vue";
+import EntryMissing from "./EntryMissing.vue";
+import FileMissing from "./FileMissing.vue";
+import UnhandledError from "./UnhandledError.vue";
+
+const home = useHomeStore();
+const hostConduit = useHostConduitService();
+
+const ifEntryMissingError = computed(() => {
+  return home.config.active.isEntryMissing;
+});
+
+const ifFileMissingError = computed(() => {
+  return home.config.active.isMissing;
+});
+
+const ifInvalidTOMLFileError = computed(() => {
+  return invalidTOMLFileError.value;
+});
+
+const ifUnhandledError = computed(() => {
+  return (
+    !ifEntryMissingError.value &&
+    !ifFileMissingError.value &&
+    !ifInvalidTOMLFileError.value
+  );
+});
+
+const invalidTOMLFileError = computed(
+  (): InvalidTOMLFileCodeError | undefined => {
+    if (configError.value && isInvalidTOMLFileCode(configError.value)) {
+      return configError.value;
+    }
+    return undefined;
+  },
+);
+
+const promptForConfigSelection = computed((): string => {
+  return home.config.active.matchingConfigs.length > 0
+    ? "Select a Configuration"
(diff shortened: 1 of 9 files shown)


marcosnav:
With a deeper look today I'm a bit confused, I thought this was solving a new problem but wasn't [this previous PR](https://github.com/posit-dev/publisher/pull/2262) handling this with an error message - providing the configuration file, the field, line and column.

Does that message stopped working? Interested to know if something changed in the codebase that the notification message stopped working, hinting that a lack of automated tests aren't covering us from side-effects from our changes?.

Also, some of these messages seem to replicate what the even better TOML extension already does (mainly schema validations, duplicated keys, etc). IMO, if there is already a more robust tool that does this for us. Do we want to jump on the same boat? this is code that we will maintain, the less we have the better.

marcosnav:
Ah, now I get it, the previous PR suppressing the calls when the config has errors isn't getting the error from the API anymore. That's why the message doesn't show anymore.

marcosnav:
I know I'm late to say this - not calling the agent when configuration files have errors made sense - But now seeing these changes seems we are creating a new solution to something that was already there. We just needed a new error type and struct with the validation details for `missing properties`.

Feels like we are pulling error handling in different directions with different practices. This is were I think we should have further discussions in the future about error handling, settling on something, trying to be consistent as possible.

sagerb:
> I know I'm late to say this - not calling the agent when configuration files have errors made sense - But now seeing these changes seems we are creating a new solution to something that was already there. We just needed a new error type and struct with the validation details for `missing properties` or a broader `invalid toml` with generic details since the inline linting by Even Better TOML does a great job.
> 
> Feels like we are pulling error handling in different directions with different practices. This is were I think we should have further discussions about error handling, settling on something, trying to be consistent as possible.

I think there is still some misunderstanding here. I'll add this topic as well to tomorrow's discussion.

extensions/vscode/webviews/homeView/src/components/configErrors/InvalidTOMLFile.vue
```diff
@@ -0,0 +1,44 @@
+<!-- Copyright (C) 2024 by Posit Software, PBC. -->
+
+<template>
+  <div v-if="error" class="error-block">
+    Configuration Error!
+    <ul class="bullets">
+      <li>Invalid TOML</li>
+      <li>
+        {{ error.data.problem }}
+      </li>
```
jonkeane, line 10: This is coming directly from the backend, yeah? Or do we process it somewhere?

Asking because I'm not sure if errors like the following are all that helpful for folks most likely to see them:

```
cannot decode TOML
string into struct field config.Config.Files of type []string
```

extensions/vscode/webviews/homeView/src/components/configErrors/ConfigErrors.vue
```diff
@@ -0,0 +1,100 @@
...
+const configError = computed((): AgentError | undefined => {
+  if (
+    home.selectedConfiguration &&
+    isConfigurationError(home.selectedConfiguration)
+  ) {
+    return home.selectedConfiguration.error;
+  }
+  return undefined;
+});
+</script>
```
jonkeane, line 100: This looks like a lot of behavior (and especially finicky behavior) to not have any tests alongside this. Could you add some?

extensions/vscode/webviews/homeView/src/components/configErrors/InvalidTOMLFile.vue
sagerb: This is being returned from the TOML parser go package we use as the error encountered. We "could" consider translating / mapping it to another error, but it would be a maintenance issue when the package changes.

dotNomad (commented):
Looking at this deeply I echo some of @marcosnav's concerns about just plainly showing TOML errors.

I do think that the componentization of the configuration errors is a good one. I'd recommend splitting that work out into a separate PR so we can isolation the conversation / feedback to just the topic of errors.

sagerb:
Going to rework changes in this PR and will present a new one.

Issue #2329, opened 2024-09-30
`inotify-tools` is required for File Watcher to work in linux installations


kgartland-rstudio:
Adding this as Documentation as I'm not sure if we can include `inotify-tools` as a dependency to the extension installation.

PR #2332, opened 2024-09-30
inotify doc udpate

kgartland-rstudio:
Adds inotify instructions to docs for linux users.

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [x] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -62,10 +62,10 @@ Or install a specific version with:
 > Linux users need to have `inotify-tools` installed for the VSCode filewatcher to work properly.
 > This impacts how the extension presents file-specific data.
 > To install `inotify` on Debian, use the following command:<br />
+>
 > ```
 > apt install inotify-tools
 > ```
-
 
 ## Optional: Install Quarto
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -58,6 +58,15 @@ Or install a specific version with:
 ./install-publisher.bash 1.0.beta1
 ```
 
+> [!WARNING]
+> Linux users need to have `inotify-tools` installed for the VSCode filewatcher to work properly.
+> This impacts how the extension presents file-specific data.
+> To install `inotify` on Debian, use the following command:<br />
+> ```
+> apt install inotify-tools
+> ```
+
+
 ## Optional: Install Quarto
 
 If you plan on deploying Quarto content having Quarto installed will allow


docs/installation.md
```diff
@@ -58,6 +58,15 @@ Or install a specific version with:
 ./install-publisher.bash 1.0.beta1
 ```
 
+> [!WARNING]
+> Linux users need to have `inotify-tools` installed for the VSCode filewatcher to work properly.
```
jonkeane, line 62: Is there a link we could add to the vscode docs that explain this?

Also is this true for positron? Or does positron bundle/otherwise get around this?

kgartland-rstudio: vscode talks about needing to update settings in inotify but i don't see anywhere they call it out as a dependency:

https://code.visualstudio.com/docs/setup/linux#_visual-studio-code-is-unable-to-watch-for-file-changes-in-this-large-workspace-error-enospc


unfortunately, I updated my mac today and Rosetta will not work with Linux AMD on Parallels (which is all Positron provides, no ARM installer) so I can't confirm.

dotNomad: > Is there a link we could add to the vscode docs that explain this

Are you referring to our `docs/vscode.md` file or _the_ VS Code docs?

dotNomad (approved):
Great to have some details in our install docs about this. I could see it being a part of a troubleshooting doc if we eventually need one as well.

Slight typo in the title of the PR, but otherwise nice addition :)

dotNomad:
Resolves #2329

docs/installation.md
jonkeane: >  _the_ VS Code docs?

This. It would be great if we showed that this isn't a special publisher issues, it's a general vscode thing.

kgartland-rstudio: > > _the_ VS Code docs?
> 
> This. It would be great if we showed that this isn't a special publisher issues, it's a general vscode thing.

Everywhere I look I only see references about updating inotify to work with large directories.  It seems most Linux distributions come with `inotify` out of the box so maybe this is more of a VM issue.

jonkeane: Sorry I missed this on Monday

> Everywhere I look I only see references about updating inotify to work with large directories. It seems most Linux distributions come with inotify out of the box so maybe this is more of a VM issue.

Yeah, that sounds right — I've had the (displeasure) of working on things that use inotify separately before and never had an issue personally with it not being generally available. 

I do still think it would be beneficial (to future-us if no one else) to link out to a canonical resource of some sort that this is a vscode thing, it's not because we're doing something special or bespoke

PR #2333, opened 2024-10-01
Extract dist rather than running just package

kgartland-rstudio:
`just package` has started failing on Windows since last Wednesday. This PR instead grabs the dist directory built in the `package` step and extracts it to we can run the vscode-ui tests on Windows.  This only impacts Nightly runs.


dotNomad (approved):
So bizarre we need to extract it on Windows, but it works!

Issue #2206, opened 2024-08-28
Content failed to start message could be better


nealrichardson:
I deployed an app, and because it needed an env var set on Connect, it failed to start. The Publisher bar has this:

![image](https://github.com/user-attachments/assets/643a8277-1ea0-4dbf-ab62-04e517474d83)

* The (x) looks like something to click to dismiss the message
* If you get that message, instead of having a button that says "View Content", how about "View Logs" and link directly to the logs in Connect? I clicked the button to go to Connect, then had to remember that logs are behind the notepad-looking icon there.

PR #2334, opened 2024-10-01
Deployment preflight errors due to bad content GUID

### LINKED ISSUE #2287: Better error text when GUID is not found on server for updating an existing deployment
If you try to redeploy to the server, but the GUID of the deployment can not be found, the agent responds with a 404 error, with the UX not showing anything helpful or actionable.

Image

Publisher log isn't much more helpful:

Image

Agent output doesn't show much:

```
time=2024-09-18T08:23:20.597-07:00 level=ERROR msg="Deployment failed" local_id=d167SN51G8U42qzK error="unexpected response from the server (404)"
```

Only the tooltip on the publisher log shows the actual call info, which does include the text "the requested object was not found", which can then be associated with the API call.

This failure is pretty bad and we can do better, just by capturing and massaging the message.

marcosnav:
## Intent

**Main intent:**
Better error messaging on deployment errors due to bad content ID on deployment toml

**Common errors handled by this PR**
- The GUID does not match the content or does not exist - 404
Screen Shot 2024-10-03 at 15 03 51

- The user does not have permissions anymore to the content - 403
Screen Shot 2024-10-03 at 15 09 55

- Content is locked
Screen Shot 2024-10-03 at 15 03 31

- Unknown error
Screen Shot 2024-10-03 at 15 56 55

Fixes #2287 

## Type of Change

Adding an extra step on the backend when deploying. Before building the bundle and doing any further operations, we will do a preflight check, trying to query the content target to validate that it can be reached, it exists and the user has permissions over it.

Returning an error with enough details to be handled by the extension code.

The changes are taking a similar approach to the http Axios errors but in this case these are messages sent by via SSE, so this is another area of the product where we are now filtering by error codes to provide a more meaningful message.

## Automated Tests

Created and updated unit tests accordingly.

## Directions for Reviewers

- [ ] For an existing content already deployed, change the `id` within the `deployment-*.toml` to:
    - [ ] To a content GUID that you don't have permissions to (or use a new credential without enough permissions to that content)
    - [ ] To a GUID for some content that does not exist

## Checklist

### DIFF
--- a/extensions/vscode/src/api/types/events.ts
+++ b/extensions/vscode/src/api/types/events.ts
@@ -1,5 +1,7 @@
 // Copyright (C) 2023 by Posit Software, PBC.
 
+import { ErrorCode } from "../../utils/errorTypes";
+
 export enum EventSourceReadyState {
   CONNECTING = 0,
   OPEN = 1,
@@ -298,10 +300,11 @@ export function getLocalId(arg: EventStreamMessage) {
   return arg.data.localId;
 }
 
-export interface EventStreamMessage {
+export interface EventStreamMessage<T = Record<string, string>> {
   type: EventSubscriptionTarget;
   time: string;
-  data: Record<string, string>;
+  data: T;
+  errCode?: ErrorCode;
   error?: string;
 }
 
--- a/extensions/vscode/src/eventErrors.test.ts
+++ b/extensions/vscode/src/eventErrors.test.ts
@@ -5,91 +5,68 @@ import { EventStreamMessage } from "./api";
 import {
   EventStreamMessageErrorCoded,
   isCodedEventErrorMessage,
-  isEvtErrTargetNotFound,
-  isEvtErrTargetForbidden,
+  isEvtErrDeploymentFailed,
   handleEventCodedError,
 } from "./eventErrors";
 import { ErrorCode } from "./utils/errorTypes";
 
-const mkEventStreamMsg = (
-  data = {},
-  errCode?: ErrorCode,
-): EventStreamMessage => {
-  return {
+function mkEventStreamMsg(data: Record<PropertyKey, any>): EventStreamMessage;
+function mkEventStreamMsg(
+  data: Record<PropertyKey, any>,
+  errCode: ErrorCode,
+): EventStreamMessageErrorCoded;
+function mkEventStreamMsg(data: Record<PropertyKey, any>, errCode?: ErrorCode) {
+  const smsg: EventStreamMessage = {
     type: "publish/failure",
     time: "Tue Oct 01 2024 10:00:00 GMT-0600",
     data,
-    errCode,
     error: "failed to publish",
   };
-};
+  if (errCode) {
+    smsg.errCode = errCode;
+  }
+  return smsg;
+}
 
 describe("Event errors", () => {
   test("isCodedEventErrorMessage", () => {
     // Message without error code
-    let streamMsg = mkEventStreamMsg();
+    let streamMsg = mkEventStreamMsg({});
     let result = isCodedEventErrorMessage(streamMsg);
     expect(result).toBe(false);
 
     // Message with error code
-    streamMsg = mkEventStreamMsg({}, "deploymentTargetNotFound");
+    streamMsg = mkEventStreamMsg({}, "deployFailed");
     result = isCodedEventErrorMessage(streamMsg);
     expect(result).toBe(true);
   });
 
-  test("isEvtErrTargetNotFound", () => {
+  test("isEvtErrDeploymentFailed", () => {
     // Message with another error code
     let streamMsg = mkEventStreamMsg({}, "unknown");
-    let result = isEvtErrTargetNotFound(
-      streamMsg as EventStreamMessageErrorCoded,
-    );
+    let result = isEvtErrDeploymentFailed(streamMsg);
     expect(result).toBe(false);
 
     // Message with error code
-    streamMsg = mkEventStreamMsg({}, "deploymentTargetNotFound");
-    result = isEvtErrTargetNotFound(streamMsg as EventStreamMessageErrorCoded);
-    expect(result).toBe(true);
-  });
-
-  test("isEvtErrTargetForbidden", () => {
-    // Message with another error code
-    let streamMsg = mkEventStreamMsg({}, "unknown");
-    let result = isEvtErrTargetForbidden(
-      streamMsg as EventStreamMessageErrorCoded,
-    );
-    expect(result).toBe(false);
-
-    // Message with error code
-    streamMsg = mkEventStreamMsg({}, "deploymentTargetIsForbidden");
-    result = isEvtErrTargetForbidden(streamMsg as EventStreamMessageErrorCoded);
+    streamMsg = mkEventStreamMsg({}, "deployFailed");
+    result = isEvtErrDeploymentFailed(streamMsg);
     expect(result).toBe(true);
   });
 
   test("handleEventCodedError", () => {
     const msgData = {
       dashboardUrl: "https://here.it.is/content/abcdefg",
+      message: "Deployment failed - structured message from the agent",
       error: "A possum on the fridge",
     };
     let streamMsg = mkEventStreamMsg(msgData, "unknown");
-    let resultMsg = handleEventCodedError(
-      streamMsg as EventStreamMessageErrorCoded,
-    );
+    let resultMsg = handleEventCodedError(streamMsg);
     expect(resultMsg).toBe("Unknown error: A possum on the fridge");
 
-    streamMsg = mkEventStreamMsg(msgData, "deploymentTargetNotFound");
-    resultMsg = handleEventCodedError(
-      streamMsg as EventStreamMessageErrorCoded,
-    );
+    streamMsg = mkEventStreamMsg(msgData, "deployFailed");
+    resultMsg = handleEventCodedError(streamMsg);
     expect(resultMsg).toBe(
-      `Content at https://here.it.is/content/abcdefg could not be found. Please, verify the content "id" is accurate.`,
-    );
-
-    streamMsg = mkEventStreamMsg(msgData, "deploymentTargetIsForbidden");
-    resultMsg = handleEventCodedError(
-      streamMsg as EventStreamMessageErrorCoded,
-    );
-    expect(resultMsg).toBe(
-      "You don't have enough permissions to deploy to https://here.it.is/content/abcdefg. Please, verify the credentials in use.",
+      "Deployment failed - structured message from the agent",
     );
   });
 });
--- a/extensions/vscode/src/eventErrors.test.ts
+++ b/extensions/vscode/src/eventErrors.test.ts
@@ -1,0 +1,95 @@
+// Copyright (C) 2024 by Posit Software, PBC.
+
+import { describe, expect, test } from "vitest";
+import { EventStreamMessage } from "./api";
+import {
+  EventStreamMessageErrorCoded,
+  isCodedEventErrorMessage,
+  isEvtErrTargetNotFound,
+  isEvtErrTargetForbidden,
+  handleEventCodedError,
+} from "./eventErrors";
+import { ErrorCode } from "./utils/errorTypes";
+
+const mkEventStreamMsg = (
+  data = {},
+  errCode?: ErrorCode,
+): EventStreamMessage => {
+  return {
+    type: "publish/failure",
+    time: "Tue Oct 01 2024 10:00:00 GMT-0600",
+    data,
+    errCode,
+    error: "failed to publish",
+  };
+};
+
+describe("Event errors", () => {
+  test("isCodedEventErrorMessage", () => {
+    // Message without error code
+    let streamMsg = mkEventStreamMsg();
+    let result = isCodedEventErrorMessage(streamMsg);
+    expect(result).toBe(false);
+
+    // Message with error code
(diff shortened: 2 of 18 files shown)


extensions/vscode/src/api/types/events.ts
```diff
@@ -298,10 +300,11 @@ export function getLocalId(arg: EventStreamMessage) {
   return arg.data.localId;
 }
 
-export interface EventStreamMessage {
+export interface EventStreamMessage<T = Record<string, string>> {
   type: EventSubscriptionTarget;
   time: string;
-  data: Record<string, string>;
+  data: T;
+  errCode?: ErrorCode;
```
marcosnav, line 307: Allowing backend error codes to be passed to SSE messages too

extensions/vscode/src/eventErrors.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2024 by Posit Software, PBC.
```
marcosnav, line 1: Similar to HTTP axios errors, creating some tooling to filter and handle incoming SSE error messages that have an error code

internal/clients/connect/client_connect.go
```diff
@@ -179,6 +179,11 @@ type connectGetContentDTO struct {
 	// Owner        *ownerOutputDTO   `json:"owner,omitempty"`
 }
 
+func (c *ConnectClient) ContentDetails(contentID types.ContentID, body *ConnectContent, log logging.Logger) error {
```
marcosnav, line 182: New client method to pull content data, helps to validate the user has access to it.

internal/clients/http_client/http_client.go
```diff
@@ -241,3 +241,12 @@ func NewHTTPClientForAccount(account *accounts.Account, timeout time.Duration, l
 		Transport: authTransport,
 	}, nil
 }
+
+func IsHTTPAgentErrorStatusOf(err error, status int) (*types.AgentError, bool) {
```
marcosnav, line 245: Handy func to check if an error that is bubbling up is an AgentError of certain http status

internal/events/events.go
```diff
@@ -17,13 +17,13 @@ type EventData = types.ErrorData
 var NoData = struct{}{}
 
 type Event struct {
-	Time time.Time
-	Type EventType
-	Data EventData
+	Time    time.Time
+	Type    EventType
+	Data    EventData
+	ErrCode ErrorCode
```
marcosnav, line 23: Include the error code on the event payload sent to the extension client

internal/publish/publish_test.go
```diff
@@ -108,119 +123,152 @@ func (s *PublishSuite) TestNewFromState() {
 }
 
 func (s *PublishSuite) TestPublishWithClientNewSuccess() {
-	s.publishWithClient(nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil)
+	s.publishWithClient(nil, &publishErrsMock{}, nil)
```
marcosnav, line 126: Changed a bit the signature here, passing a lot of `nil`s in a certain order was growing too much

internal/types/error.go
```diff
@@ -16,6 +16,9 @@ const (
 	ErrorUnknownTOMLKey               ErrorCode = "unknownTOMLKey"
 	ErrorInvalidConfigFiles           ErrorCode = "invalidConfigFiles"
 	ErrorCredentialServiceUnavailable ErrorCode = "credentialsServiceUnavailable"
+	ErrorDeploymentTargetNotFound     ErrorCode = "deploymentTargetNotFound"
+	ErrorDeploymentTargetIsForbidden  ErrorCode = "deploymentTargetIsForbidden"
+	ErrorDeploymentTargetUnreachable  ErrorCode = "deploymentTargetUnreachable"
```
jonkeane, line 21: Do we need three unique `deploymentTarget` error types here? Or is it sufficient to pass this information within the error itself? The lines of code are free, but the more types we have the more names we have to keep track of + write in all the places, etc.

sagerb: The concept was that these were three types of errors, which we needed to pass through within the error, so that the UX could identify them and place them into a specific error type (to be handled according per the type). Having them listed in one place allows us to keep them synchronized from one place, since we're not sharing code between the agent and the UX. 

I believe this concept still holds and is a good practice, since the error code here is functioning as a const or enum.

internal/publish/publish.go
```diff
@@ -265,6 +267,28 @@ func (p *defaultPublisher) createDeploymentRecord(
 	return p.writeDeploymentRecord()
 }
 
+// Does the target exist and the user has permissions to it?
+func (p *defaultPublisher) targetPreflightChecks(client connect.APIClient, contentID types.ContentID) error {
```
dotNomad, line 271: I noticed that because the error is retrieved from `ContentDetails` which does a `GET` call and these errors have codes plugged in, but nothing else, the error we write to the deployment record `message` is largely un-helpful. The message is just what the API call error has in `message`

Currently that is the error we show for the "Last Deployment Failed" message in the sidebar. So while I get a great message when the piece of content cannot be found in the notification window the sidebar error is quite generic:

CleanShot 2024-10-02 at 13 00 20@2x

Not sure if we should consider that part of this work, but I'd argue because its a one line addition it should be.

extensions/vscode/src/eventErrors.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    streamMsg = mkEventStreamMsg({}, "deploymentTargetNotFound");
+    result = isCodedEventErrorMessage(streamMsg);
+    expect(result).toBe(true);
+  });
+
+  test("isEvtErrTargetNotFound", () => {
+    // Message with another error code
+    let streamMsg = mkEventStreamMsg({}, "unknown");
+    let result = isEvtErrTargetNotFound(
+      streamMsg as EventStreamMessageErrorCoded,
```
dotNomad, line 44: The `as` keyword always makes me nervous. This is test code so feel free to ignore this, but I would make `mkEventStreamMsg` have a generic so you could do something like

```
let streamMsg = mkEventStreamMsg<EventStreamMessageErrorCoded>({}, "unknown");
```

or similar. That would remove the need for the `as`. You could even change the return type if `errCode` is passed by overloading the definition of the method.

extensions/vscode/src/eventErrors.ts
```diff
@@ -0,0 +1,46 @@
...
+  emsg: EventStreamMessageErrorCoded,
+): emsg is EventStreamMessageErrorCoded<deploymentEvtErr> => {
+  return emsg.errCode === "deploymentTargetIsForbidden";
+};
+
+export const handleEventCodedError = (
+  emsg: EventStreamMessageErrorCoded,
+): string => {
+  if (isEvtErrTargetNotFound(emsg)) {
+    return `Content at ${emsg.data.dashboardUrl} could not be found. Please, verify the content "id" is accurate.`;
```
dotNomad, line 39: I don't think this is an error that most users will be able to fix on their own. 

This is because the `id` is inside of the Content Record under `.posit/publish/deployments` which we never prompt the user to open, and is a little tough to find the correct deployment for your configuration if you have multiple / don't know what to look for.

This is a much better error that what we have now, so perhaps what I'm suggesting should be a follow-up, but we should have a way for users to solve this issue.

extensions/vscode/src/eventErrors.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+  if (isEvtErrTargetForbidden(emsg)) {
+    return `You don't have enough permissions to deploy to ${emsg.data.dashboardUrl}. Please, verify the credentials in use.`;
```
dotNomad, line 42: Should we include something about requesting Collaborator permissions from the owner as another option?

extensions/vscode/src/eventErrors.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+  const unknownErrMsg = emsg.data.error || emsg.data.message;
+  return `Unknown error: ${unknownErrMsg}`;
+};
```
dotNomad, line 46: After a thorough review of this PR I think I understand @jonkeane's point a bit more about the codes for event messages.

These errors don't look to be doing the same amount of construction we expect the frontend to do for other errors (components, links, buttons, etc).

A highlight of this is the locked error. If you lock a piece of content and attempt to deploy you get a pretty good error even though it is unknown:

CleanShot 2024-10-02 at 13 17 01@2x

That didn't need a error code, if statements, or really much at all for it to make sense to the user.

Of course we could add a nice link to the content in the notification to adjust the lock setting, and if we want that we need a code to understand which error was passed up.

dotNomad (commented):
I had some concerns about the errors here that I think we can bring to a discussion with the team.

internal/publish/publish.go
marcosnav: Maybe improvements to this messaging be related to this other issue #2206?

extensions/vscode/src/eventErrors.ts
marcosnav: That was the initial implementation a few weeks ago, the agent sending the error message and the extension only showing it. The argument of having the flexibility of sending only the error code so the UI can do many things is the approach we have now. Some error codes will require some construction on the UI while others will be fairly minimal.

IMO, at this point, consistency should be the key on where do we build this messages to avoid guessing games for each error message.

In this locked content example, I think it is a nice catch, it should have it's error code and be a known error and not show up as an `Unknown error` message.

extensions/vscode/src/eventErrors.test.ts
marcosnav: That's a great idea, I'll try that.

extensions/vscode/src/eventErrors.ts
marcosnav: I think this provides enough info for now, considering it might be very likely to happen when associating content, the context of the interaction feels appropriate. Totally agree that there's room for improvement but user feedback may provide a better direction.

extensions/vscode/src/eventErrors.ts
marcosnav: Makes sense, WDYT of?
```
... .You may need to request collaborator permissions or verify the credentials in use
```

internal/publish/publish.go
dotNomad: The methodology to set that message is the same, but they are different parts of the publishing process. Ideally we would solve one at a time.

extensions/vscode/src/eventErrors.ts
dotNomad: That sounds great 👍

extensions/vscode/src/eventErrors.ts
dotNomad: > IMO, at this point, consistency should be the key on where do we build this messages to avoid guessing games for each error message.

Completely agree. I lean toward consistency. I was trying to voice my understanding of some of the other concerns brought up, and think out loud how I'm reasoning about which to choose. It sounds like that didn't come across that way - sorry about that.

marcosnav: No worries, glad to hear that. Sharing concerns and thinking out loud always helps on getting to common ground and understanding, I appreciate that 👍

sagerb: I will say that I expect we'll take more and more advantage of having the finely defined codes returned from the agent. I'm going to have to implement a richer error type coming back from the agent for deployment failure, so that I can catch when it has occurred due to TLS Certificate Verification failure, and offer the user the link to adjust the setting. Just wanted to add this context into the discussion. I also lean towards consistency here.

internal/publish/publish.go
sagerb: I would think if we created some kind of error message factory, which took in the original error (present either in the stream or in the content record file), then we can consistently produce the enhanced message.

marcosnav:
@dotNomad @jonkeane Pushed some changes, addressing the feedback and also found a way to have one single error code for the deployment errors, respecting the message as it comes from the agent. Still having the option to have more specific deployment error codes in the future if there's the need.

internal/publish/publish.go
```diff
@@ -265,6 +267,49 @@ func (p *defaultPublisher) createDeploymentRecord(
...
+	if aerr, isAgentErr := types.IsAgentError(err); isAgentErr {
+		return p.preflightAgentError(aerr, contentID)
+	}
+	if err != nil {
+		p.log.Debug("Deployment target cannot be reached. Halting deployment", "content_id", contentID)
+		return types.NewAgentError(events.DeploymentFailedCode, err, nil)
+	}
+
+	// Verify content is not locked
+	lockedErr := content.LockedError()
```
dotNomad, line 304: Nice that you got the locked error in here too 😄

internal/publish/publish_test.go
dotNomad: Much better!

extensions/vscode/src/eventErrors.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+  EventStreamMessageErrorCoded,
+  isCodedEventErrorMessage,
+  isEvtErrDeploymentFailed,
+  handleEventCodedError,
+} from "./eventErrors";
+import { ErrorCode } from "./utils/errorTypes";
+
+function mkEventStreamMsg(data: Record<PropertyKey, any>): EventStreamMessage;
+function mkEventStreamMsg(
```
dotNomad, line 14: Little bit more complex up here, but it looks so much more usable in the tests themselves. Great work on the typing!

dotNomad (approved):
Absolutely fantastic work here. Thanks for addressing those bits of feedback 🙇

PR #2335, opened 2024-10-01
Be quiet for symlink walking errors

### LINKED ISSUE #2267: Agent encountering/responding with Error Code 500 while refreshing files associated with configuration file
From slack thread: https://positpbc.slack.com/archives/C05D7NZD52S/p1726068297152959

This might be something specific to my environment, but while editing files in Emacs, the publisher was constantly trying to refresh the project and sometimes produced errors like the following, which are temporary files used while in-flight edits are being performed. I received several of these errors:

Image

Q> How can I get these refresh errors to stop? They are happening even while the publisher view is not open. Do I need to remove the extension?

Unfortunately, there is no specific way to disable them. We rely on them for a large amount of functionality, so a bug in this area is not very mitigable. Without uninstalling the extension, you could disable it (and then enable it back when you want to try its functionality). Otherwise, you could uninstall it until we resolve that issue.

Q> OK. I’ll disable for now. I didn’t expect it to be doing background things while not thinking about publishing.

Looking closer at the error code, it is your symlink `.#bug.py` that is giving the agent problems. This is a good bug and one which we hadn’t seen before. Thanks!
It is an interesting comment to consider: “I didn’t expect it to be doing background things while not thinking about publishing.“. As an extension, part of our functionality is loaded immediately by VSCode, ahead of our activation of our UX. The background functionality supports the command pallet commands and our file-specific entrypoint detection. We have taken a simple path to start with, but could certainly be much more thoughtful about what happens ahead of showing the UX.

marcosnav:
## Intent

This PR aims to not pollute VSCode with notification errors about symlinks that do not exist. 

When a directory being watched by the Publisher extension is being modified by other software, there's the chance for symbolic links to be created to files that do not exist, like Emacs which creates symbolic files for unsaved modifications.

With this PR changes, we'll log the behavior and errors for symbolic links but won't show error notifications.

Screen Shot 2024-10-01 at 13 54 59

Fixes #2267 

## Approach

Log symbolic link walking errors, do not bubble them up to the extension code as errors.

## Automated Tests

Updated unit tests accordingly.

## Directions for Reviewers

Creating a symbolic link to a file that does not exist should show the `"Error following symlink" ... "no such file or directory"`  debug log but not notification error should show up.

E.g: 
`ln -s foo bar` should show something like
```
time=2024-10-01T14:45:03.224-06:00 level=DEBUG msg="Error following symlink, ignoring file" filepath=/Users/mn/Posit/_ide/top-5-income-share-streamlit/bar error="lstat /Users/mn/Posit/_ide/top-5-income-share-streamlit/foo: no such file or directory"
```

## Checklist

### DIFF
--- a/internal/util/symlink_walker.go
+++ b/internal/util/symlink_walker.go
@@ -42,7 +42,12 @@ func (w *symlinkWalker) visit(fn AbsoluteWalkFunc) AbsoluteWalkFunc {
 			linkTarget, err := filepath.EvalSymlinks(path.String())
 			targetPath := NewPath(linkTarget, path.Fs())
 			if err != nil {
-				return fmt.Errorf("error following symlink %s: %w", path, err)
+				// There's software that creates symbolic links to files which do not exist,
+				// like Emacs which creates symbolic files for unsaved modifications to files.
+				// We'll log the behavior but not return errors to not polute the user with error notifications
+				// when another piece of software is dealing with the same directory.
+				w.log.Debug("Error following symlink, ignoring file", "filepath", path, "error", err.Error())
+				return nil
 			}
 			targetInfo, err := targetPath.Stat()
 			if err != nil {
--- a/internal/util/symlink_walker.go
+++ b/internal/util/symlink_walker.go
@@ -3,7 +3,6 @@ package util
 // Copyright (C) 2023 by Posit Software, PBC.
 
 import (
-	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -45,14 +44,15 @@ func (w *symlinkWalker) visit(fn AbsoluteWalkFunc) AbsoluteWalkFunc {
 				// like Emacs which creates symbolic files for unsaved modifications to files.
 				// We'll log the behavior but not return errors to not polute the user with error notifications
 				// when another piece of software is dealing with the same directory.
-				w.log.Debug("Error following symlink, ignoring file", "filepath", path, "error", err.Error())
+				w.log.Warn("Error following symlink, ignoring file", "filepath", path, "error", err.Error())
 				return nil
 			}
 
 			targetPath := NewPath(linkTarget, path.Fs())
 			targetInfo, err := targetPath.Stat()
 			if err != nil {
-				return fmt.Errorf("error getting target info for symlink %s: %w", targetPath, err)
+				w.log.Warn("Error getting info for symlink", "filepath", targetPath, "error", err.Error())
+				return nil
 			}
 			// Visit symlink target info but use the path to the link.
 			err = w.visit(fn)(path, targetInfo, nil)
--- a/internal/util/symlink_walker.go
+++ b/internal/util/symlink_walker.go
@@ -40,7 +40,6 @@ func (w *symlinkWalker) visit(fn AbsoluteWalkFunc) AbsoluteWalkFunc {
 		if info.Mode().Type()&os.ModeSymlink != 0 {
 			w.log.Info("Following symlink", "path", path)
 			linkTarget, err := filepath.EvalSymlinks(path.String())
-			targetPath := NewPath(linkTarget, path.Fs())
 			if err != nil {
 				// There's software that creates symbolic links to files which do not exist,
 				// like Emacs which creates symbolic files for unsaved modifications to files.
@@ -49,6 +48,8 @@ func (w *symlinkWalker) visit(fn AbsoluteWalkFunc) AbsoluteWalkFunc {
 				w.log.Debug("Error following symlink, ignoring file", "filepath", path, "error", err.Error())
 				return nil
 			}
+
+			targetPath := NewPath(linkTarget, path.Fs())
 			targetInfo, err := targetPath.Stat()
 			if err != nil {
 				return fmt.Errorf("error getting target info for symlink %s: %w", targetPath, err)
--- a/internal/bundles/bundle_test.go
+++ b/internal/bundles/bundle_test.go
@@ -396,6 +396,8 @@ func (s *BundlerSuite) TestNewBundleFromDirectorySymlinks() {
 	}, s.getTarFileNames(dest))
 }
 
+// We log the issues with symbolic links but not return them to not polute the user with error notifications
+// when another piece of software is dealing with the same directory.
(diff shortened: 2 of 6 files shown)


jonkeane:
I'm not opposed to this PR itself, but we need something broader to prevent errors _like these_ from bubbling up in front of people and making them believe things are wrong. We know about symlinks today, but there will be some other issue / file type that causes these tomorrow.

sagerb:
> I'm not opposed to this PR itself, but we need something broader to prevent errors _like these_ from bubbling up in front of people and making them believe things are wrong. We know about symlinks today, but there will be some other issue / file type that causes these tomorrow.

@jonkeane perhaps we can discuss this tomorrow in sync? I'd like to discuss further any ideas you have for solving this challenging problem.

dotNomad:
> I'm not opposed to this PR itself, but we need something broader to prevent errors _like these_ from bubbling up in front of people and making them believe things are wrong. We know about symlinks today, but there will be some other issue / file type that causes these tomorrow.

If we want to avoid showing users these types of errors unilaterally we could fail silently when our `GET /api/configurations/$name/files` endpoint responds with an error. We could log the error rather than using `window.showErrorMessage` like [we are here](https://github.com/posit-dev/publisher/blob/7e07f78554babe5549e0846af266659dc2870514/extensions/vscode/src/views/homeView.ts#L1507C16-L1507C32).

Unfortunately that means the Project Files view will not be responsive, and inaccurate since we aren't getting the files response from the API. Do you think that is the right trade of @jonkeane?

The other option would be avoiding sending back errors on the API side whenever possible, like this PR does for the symlink case.

sagerb:
@jonkeane I agree.

@marcosnav I'm thinking that the solution to the issue (#2267) should cast a wider net. You've identifed one location where an error is returned from a called function, and we're suppressing it. I'd recommend that you walk through the code which walks directories and files, and suppress the errors which we don't care about. For example, right below the code you modified is call to `lstat` which could fail (https://github.com/posit-dev/publisher/blob/96c67e8552b246f478a62543d6a977519ae6954c/internal/util/symlink_walker.go#L53). If it does return an error, its not one we want to both the user with. (The current code will return a nasty error to be displayed to the user.) This is another case where we should ignore the file and move on.

I think this approach is most applicable to our agent code which is walking and examining files. If we encounter any issue accessing a file or directory as we walk, let's log it and move on. It doesn't have to be a big deal (I don't know that I'd place large comments in the codebase describing why, but we can certainly document it to show we are ignoring it.).

I'll add a topic to the agenda for tomorrow's sync meeting, so we can discuss it.

dotNomad (commented):
Tested this locally, and it worked great. I do agree that #2267 will require a bit more to be considered complete.

I have no qualms with merging this as a stepping stone though other than the CI failure on Windows. That looks like it needs some investigation.

marcosnav:
Thanks folks! All these arguments make sense, I like @sagerb's idea of scoping the changes to the agent code, specially because aside from the symlink walker there is only one more doing the same.

Happy to chat the details in our next sync.

dotNomad (approved):
Thanks for fixing that test up. This looks great 😎 
Looking forward to discussing our error catching strategy tomorrow.

internal/util/symlink_walker.go
```diff
@@ -42,7 +42,12 @@ func (w *symlinkWalker) visit(fn AbsoluteWalkFunc) AbsoluteWalkFunc {
 			linkTarget, err := filepath.EvalSymlinks(path.String())
 			targetPath := NewPath(linkTarget, path.Fs())
 			if err != nil {
-				return fmt.Errorf("error following symlink %s: %w", path, err)
+				// There's software that creates symbolic links to files which do not exist,
```
aronatkins, line 45: nit: Just above: Move the computation of `targetPath` after this error handling.

marcosnav: Ah! nice catch, thanks!

internal/util/symlink_walker.go
```diff
@@ -42,7 +42,12 @@ func (w *symlinkWalker) visit(fn AbsoluteWalkFunc) AbsoluteWalkFunc {
...
+				// like Emacs which creates symbolic files for unsaved modifications to files.
+				// We'll log the behavior but not return errors to not polute the user with error notifications
+				// when another piece of software is dealing with the same directory.
+				w.log.Debug("Error following symlink, ignoring file", "filepath", path, "error", err.Error())
+				return nil
 			}
 			targetInfo, err := targetPath.Stat()
```
sagerb, line 52: I think we want to ignore this error.

as well as the `targetPath.ReadDir()` on line 62.

I'd also be a little concerned with what errors may be returned on line 72 in total.

sagerb (changes requested):
I believe the general consensus was to increase the scope to suppress all ignorable errors within the walkers. Let me know if I have this wrong. Thanks!

dotNomad:
> I believe the general consensus was to increase the scope to suppress all ignorable errors within the walkers. Let me know if I have this wrong

Is there a reason we can't do this across multiple PRs? Or is the "Changes requested" to expand the issue?

jonkeane:
> I believe the general consensus was to increase the scope to suppress all ignorable errors within the walkers. Let me know if I have this wrong. Thanks!

I thought the opposite. I think we should ship what's here (so long as there aren't glaring issues, of course) and we will do more work in the future to be better about not erroring loudly when we shouldn't be.

> I have no qualms with merging this as a stepping stone though other than the CI failure on Windows. That looks like it needs some investigation.
https://github.com/posit-dev/publisher/pull/2335#pullrequestreview-2341616578

> I'm not opposed to this PR itself, but we need something broader to prevent errors like these from bubbling up in front of people and making them believe things are wrong. We know about symlinks today, but there will be some other issue / file type that causes these tomorrow.
https://github.com/posit-dev/publisher/pull/2335#issuecomment-2387061447

marcosnav:
I already have some more work around this, new changes are small enough to be pushed within this PR and will cover more areas as @sagerb suggests, early this Friday the changes will be up 👍 

Sorry for the lack of clarity and thanks for the patience 🙇

marcosnav:
New commit to prevent errors or scary warnings that block the user when a file presents a failure while inspecting directories. Still, logging all the failures.

sagerb (approved):
Thanks for the additional checks!

Issue #2326, opened 2024-09-30
Secrets and Environment collision


kgartland-rstudio:
If I have a config that includes both a Secret an Environment Variable with the same name, Secret wins. If I then go to redeploy the Environment variable will win because we clear out the Secret after the initial deployment.

This could lead to some confusion.  Can we add some validation to ensure Secrets and Environment cannot contain the same name?

Example config:

```
# Configuration file generated by Posit Publisher.
# Please review and modify as needed. See the documentation for more options:
# https://github.com/posit-dev/publisher/blob/main/docs/configuration.md
'$schema' = 'https://cdn.posit.co/publisher/schemas/posit-publishing-schema-v3.json'
type = 'python-fastapi'
entrypoint = 'simple.py'
validate = true
files = [
  '/simple.py',
  '/requirements.txt'
]
title = 'fastapi-simple'
secrets = ['secret1', 'secret2', 'secret3']

[environment]
secret3="environment"

[python]
version = '3.11.4'
package_file = 'requirements.txt'
package_manager = 'pip'
```

dotNomad:
I'm thinking a combination of two changes could be very helpful here:

1. When using the UI, change the add behavior to prevent adding a secret with the same name as an environment variable, or visa versa.

2. When the selected Configuration has Secret / Env Var collision show a warning and potentially disable the deploy button (look to other behavior to keep it consistent).

PR #2337, opened 2024-10-01
Check if an environment variable exists when validating added secret

dotNomad:
This adds a small bit of validation code to our `addSecret` flow that checks to see if an environment variable already exists with the input name, and prevents the user from inputting it.

This is added at the VS Code input level, and on the API endpoint.

<details>
  <summary>Preview</summary>
  
CleanShot 2024-10-01 at 15 09 42@2x

</details> 

## Intent

Part of #2326

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Automated tests were included for the Go code.

## Directions for Reviewers

1. Add a Secret via the UI
2. Attempt to add a Secret via the UI that has the same name as an environment variable already in the Configuration

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -964,7 +964,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
           return "Secret names cannot be empty.";
         }
         if (existingKeys.has(value)) {
-          return "There is already an environment variable with this name. Secrets and environment variable cannot share a name.";
+          return "There is already an environment variable with this name. Secrets and environment variable names must be unique.";
         }
         return;
       },
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -35,6 +35,7 @@ import {
   isPreContentRecordWithConfig,
   useApi,
   AllContentRecordTypes,
+  EnvironmentConfig,
 } from "src/api";
 import { useBus } from "src/bus";
 import { EventStream } from "src/events";
@@ -944,7 +945,16 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     }
   }
 
-  private inputSecretName = async () => {
+  private inputSecretName = async (
+    environment: EnvironmentConfig | undefined,
+  ) => {
+    const existingKeys = new Set();
+    if (environment) {
+      for (const secret in environment) {
+        existingKeys.add(secret);
+      }
+    }
+
     return await window.showInputBox({
       title: "Add a Secret",
       prompt: "Enter the name of the secret.",
@@ -952,6 +962,9 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       validateInput: (value: string) => {
         if (value.length === 0) {
           return "Secret names cannot be empty.";
+        }
+        if (existingKeys.has(value)) {
+          return "There is already an environment variable with this name. Secrets and environment variable cannot share a name.";
         }
         return;
       },
@@ -971,7 +984,9 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       return;
     }
 
-    const name = await this.inputSecretName();
+    const name = await this.inputSecretName(
+      activeConfig.configuration.environment,
+    );
     if (name === undefined) {
       // Cancelled by the user
       return;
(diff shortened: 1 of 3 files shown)


extensions/vscode/src/views/homeView.ts
```diff
@@ -953,6 +963,9 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
         if (value.length === 0) {
           return "Secret names cannot be empty.";
         }
+        if (existingKeys.has(value)) {
+          return "There is already an environment variable with this name. Secrets and environment variable cannot share a name.";
```
jonkeane, line 967:
```suggestion
          return "There is already an environment variable with this name. Secrets and environment variable names must be unique.";
```

Slightly more direct?

dotNomad: Adjusted. Great suggestion 👍

extensions/vscode/src/views/homeView.ts
```diff
@@ -944,7 +945,16 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+  private inputSecretName = async (
+    environment: EnvironmentConfig | undefined,
+  ) => {
+    const existingKeys = new Set();
+    if (environment) {
+      for (const secret in environment) {
+        existingKeys.add(secret);
+      }
+    }
+
```
sagerb, line 957: I'd suggest you create this set each time you are called to validate. That will allow a user to encounter the error, then correct it by removing the old secret, and then continue forward with the new secret definition.

I'd also suggest you perform this validation during the `finalValidation` callback rather than the one which is called on every keystroke. That should address any concerns raised with my suggestion above.

sagerb (changes requested):
Pointed out an improvement, which I think will enhance the experience. Let me know what you think. The largest improvement will be not showing the error on every keystroke when it is a duplicate, but the user is not yet done typing.

extensions/vscode/src/views/homeView.ts
dotNomad: > create this set each time you are called to validate

That won't quite do it. I got this working and I have to re-get the configuration entirely since it can be lost, have an error, or an entirely different `environment` since the input is using `ignoreFocusOut: true`. You can go into the configuration file and make any changes you want.

Like I said I got this working, but it doesn't feel like a good user or coding experience. There are way more checks in the code, and more avenues for user mistakes.

This feels like a really good reason to only use `ignoreFocusOut` when necessary. I'll add something to our team discussions to go over this.

> this validation during the finalValidation callback

There isn't a way to use our `finalValidation` without creating a `MultiStepInput`. That seems like a good follow-up to create that and use it here.

I do worry about the maintenance cost of that as oppose to using the `validateInput` style of VS Code, but do agree its a better UX.

dotNomad: Posting the code here to compare.

```typescript
    return await window.showInputBox({
      title: "Add a Secret",
      prompt: "Enter the name of the secret.",
      ignoreFocusOut: true,
      validateInput: async (value: string) => {
        const activeConfig = await this.state.getSelectedConfiguration();

        if (activeConfig === undefined) {
          return "The selected Configuration has been lost. Please select a Configuration and try again.";
        }
        if (isConfigurationError(activeConfig)) {
          return "The selected Configuration has an error. Please fix the error and try again.";
        }

        const environment = activeConfig.configuration.environment;
        const existingKeys = new Set();
        if (environment) {
          for (const secret in environment) {
            existingKeys.add(secret);
          }
        }

        if (value.length === 0) {
          return "Secret names cannot be empty.";
        }
        if (existingKeys.has(value)) {
          return "There is already an environment variable with this name. Secrets and environment variable names must be unique.";
        }
        return;
      },
    });
```

sagerb: Yeah, ok. I see your point. Thanks for exploring it.

sagerb (approved):

extensions/vscode/src/views/homeView.ts
dotNomad: I think there is a discussion to be had here, but we should leave it for after `1.2.0`

PR #2338, opened 2024-10-01
Show error and disable deployment when environment and secrets have duplicates

### LINKED ISSUE #2326: Secrets and Environment collision

dotNomad:
This PR adds a new error that disables deployment - similar to a missing credential error in style - when there are names shared in Environment and Secrets in the selected Configuration.

<details>
  <summary>Preview</summary>
  
CleanShot 2024-10-01 at 16 15 42@2x

CleanShot 2024-10-01 at 16 15 30@2x

</details> 

The duplicated names are present in the error to make it easier to resolve, along with a link to edit the configuration similar to other errors.

## Intent

Resolves #2326 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/DeployButton.vue
+++ b/extensions/vscode/webviews/homeView/src/components/DeployButton.vue
@@ -1,7 +1,12 @@
 <template>
   <vscode-button
     :data-automation="`deploy-button`"
-    :disabled="!haveResources || isConfigInError || home.publishInProgress"
+    :disabled="
+      !haveResources ||
+      isConfigInError ||
+      home.publishInProgress ||
+      home.duplicatedEnvironmentVariables.length
+    "
     @click="deploy"
   >
     Deploy Your Project
--- a/extensions/vscode/webviews/homeView/src/components/DeployButton.vue
+++ b/extensions/vscode/webviews/homeView/src/components/DeployButton.vue
@@ -1,12 +1,7 @@
 <template>
   <vscode-button
     :data-automation="`deploy-button`"
-    :disabled="
-      !haveResources ||
-      isConfigInError ||
-      home.publishInProgress ||
-      home.duplicatedEnvironmentVariables.length
-    "
+    :disabled="!haveResources || isConfigInError || home.publishInProgress"
     @click="deploy"
   >
     Deploy Your Project
--- a/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
+++ b/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
@@ -31,12 +31,13 @@
       <template v-if="home.duplicatedEnvironmentVariables.length">
         <p>
           <template v-if="home.duplicatedEnvironmentVariables.length === 1">
-            A duplicate name was
+            A variable was set as both a secret and environment variable. It
+            must only be set as one or the other.
           </template>
           <template v-if="home.duplicatedEnvironmentVariables.length > 1">
-            Duplicate names were
+            Variables were set as both secrets and environment variables. They
+            must only be set as one or the other.
           </template>
-          found in Secrets and Environment.
           <a
             class="webview-link"
             role="button"
--- a/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
+++ b/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
@@ -28,6 +28,28 @@
         />
       </div>
 
+      <template v-if="home.duplicatedEnvironmentVariables.length">
+        <p>
+          <template v-if="home.duplicatedEnvironmentVariables.length === 1">
+            A duplicate name was
+          </template>
+          <template v-if="home.duplicatedEnvironmentVariables.length > 1">
+            Duplicate names were
+          </template>
+          found in Secrets and Environment.
+          <a
+            class="webview-link"
+            role="button"
+            @click="
+              onEditConfiguration(home.selectedConfiguration!.configurationPath)
+            "
+            >Edit the Configuration</a
+          >.
+        </p>
+        <p>{{ home.duplicatedEnvironmentVariables.join(", ") }}</p>
+        <p></p>
+      </template>
+
       <p v-if="home.config.active.isEntryMissing">
         No Config Entry in Deployment record -
         {{ home.selectedContentRecord?.saveName }}.
--- a/extensions/vscode/webviews/homeView/src/stores/home.ts
+++ b/extensions/vscode/webviews/homeView/src/stores/home.ts
@@ -31,6 +31,33 @@ export const useHomeStore = defineStore("home", () => {
 
   const secrets = ref(new Map<string, string | undefined>());
 
+  const environment = computed((): Map<string, string> => {
+    const result = new Map<string, string>();
+    const config = selectedConfiguration.value;
+
+    if (config === undefined || isConfigurationError(config)) {
+      return result;
+    }
+
+    Object.entries(config.configuration.environment || {}).forEach(
+      ([name, value]) => {
+        result.set(name, value);
+      },
+    );
+
+    return result;
+  });
+
+  const duplicatedEnvironmentVariables = computed((): string[] => {
+    const result: string[] = [];
+    secrets.value.forEach((_, name) => {
+      if (environment.value.has(name)) {
+        result.push(name);
+      }
+    });
+    return result;
+  });
+
   const showDisabledOverlay = ref(false);
 
   const selectedContentRecord = ref<ContentRecord | PreContentRecord>();
@@ -375,6 +402,8 @@ export const useHomeStore = defineStore("home", () => {
     credentials,
     sortedCredentials,
     secrets,
+    environment,
+    duplicatedEnvironmentVariables,
     selectedContentRecord,
     selectedConfiguration,
     serverCredential,


extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -28,6 +28,28 @@
...
+          </template>
+          found in Secrets and Environment.
+          <a
+            class="webview-link"
+            role="button"
+            @click="
+              onEditConfiguration(home.selectedConfiguration!.configurationPath)
+            "
+            >Edit the Configuration</a
+          >.
```
jonkeane, line 47: What about saying something like "(A variable was|Variables were) set as both a secret and environment variable. (It|They) must only be in one or the other. Edit the Configuration." It's true that the names are the problem that's an issue, but at this point in the UX I think most people think of them more holistically and the names are below so it's clear that the names are the issue.

extensions/vscode/webviews/homeView/src/components/DeployButton.vue
```diff
@@ -1,7 +1,12 @@
 <template>
   <vscode-button
     :data-automation="`deploy-button`"
-    :disabled="!haveResources || isConfigInError || home.publishInProgress"
+    :disabled="
+      !haveResources ||
+      isConfigInError ||
+      home.publishInProgress ||
+      home.duplicatedEnvironmentVariables.length
+    "
```
jonkeane, line 9: In this case, disabling the deploy might be the right answer, but I am concerned about this list expanding. Are we sure all of those checking functions have 0 bugs and will never have a false positive?

sagerb: There are certainly error conditions which we know we cannot perform the operation (because we cannot determine the API parameters to make the call), and then some which we believe we have analyzed the state and determined it would not be good to deploy at this time. Perhaps in the latter case, we should offer an escape hatch to allow them to bypass bugs within our logic, rather than just lock them out. But to do this, we do have to decide how we want to display warnings and errors regarding the state.

sagerb:
@dotNomad The work you've done here definitely overlaps/conflicts with the work in https://github.com/posit-dev/publisher/pull/2331. We should discuss how we'll move forward and resolve this.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
dotNomad: Great suggestion. I really struggled coming up with text here and I think yours is a huge improvement. I'll make that change.

extensions/vscode/webviews/homeView/src/components/DeployButton.vue
dotNomad: > In this case, disabling the deploy might be the right answer, but I am concerned about this list expanding. Are we sure all of those checking functions have 0 bugs and will never have a false positive?

I echo your concern. Personally, with tools like this one, I would prefer to be able to "continue anyway" when given a warning that can be ignored. This is the first one that has been added to this list that technically can be deployed. The rest of the checks stop a completely invalid deploy like Bill mentioned.

I stuck with the previous pattern rather than introducing a new one, but I could remove the addition here and just show the warning instead.

Having good UX around these warnings (rather than errors that disable Deploy) becomes the harder question. If a user frequently wants to ignore a warning it would be nice if it was dismissible, but that would drastically increase the scope of this issue.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
dotNomad: Here is the UI with that text change with very minor adjustments. Let me know what you think.

<details>
  <summary>Images</summary>
  
CleanShot 2024-10-02 at 11 13 09
CleanShot 2024-10-02 at 11 13 21

</details>

extensions/vscode/webviews/homeView/src/components/DeployButton.vue
dotNomad: Had a quick chat with @jonkeane to figure out the action-ables from this chain:

- remove the disabling of the Deploy button for this case since it is still technically deployable it will just have odd behavior. We are still highlighting that with the warning message added in this PR
- create a follow-up task to analyze the states in which we disable the deploy button, and ideally remove them

The goal is to never disable the deploy button (if possible) and instead let informative errors come up during the deployment process. That way the user can always "continue anyway", and still deploy their content rather than be stuck.

For example, if a configuration is in error, let the user know, but let them still deploy to get the error we hit during deployment.

jonkeane: Thanks for that fantastic summary 🙏  💯

dotNomad: https://github.com/posit-dev/publisher/pull/2338/commits/e58aefe3b78114f15056588c6969891f234f891b has removed the disabling of the Deploy button for this PR and #2344 has been opened up as a follow-up

dotNomad:
> We should discuss how we'll move forward and resolve this.

We could merge this prior since it is a relatively simpler change, and componentize it with the other work.

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -31,6 +31,33 @@ export const useHomeStore = defineStore("home", () => {
...
+    }
+
+    Object.entries(config.configuration.environment || {}).forEach(
+      ([name, value]) => {
+        result.set(name, value);
+      },
+    );
+
+    return result;
+  });
```
sagerb, line 49: If you feel the environment vars are much better represented by a Map, what do you think about having the transport vs. internal representation being different here? 

If it is not hugely beneficial, then I'd prefer to be interacting with these directly as the type we have already defined. I can't think the inefficiency would be that much different, just different access techniques. Thoughts?

sagerb (commented):
Let me know what you think about my question. Thanks!

extensions/vscode/webviews/homeView/src/stores/home.ts
dotNomad: I used a Map due to its ease of use with `.has()`, `.get()`, and `.size`. We only use `.has()` in this PR, but the extendability is nice.

I tend to use Maps when applicable because of past experiences hitting the accidental keys issue (plus some other benefits which are less than applicable here): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps

I use a Map for secrets for the same reasons and use all the features of a map there. That does introduce the type of code below though since Maps do not have native support for serialization to JSON

https://github.com/posit-dev/publisher/blob/e58aefe3b78114f15056588c6969891f234f891b/extensions/vscode/webviews/homeView/src/components/DeployButton.vue#L50

> I'd prefer to be interacting with these directly as the type we have already defined

We are limited in defining types since JSON only allows objects and not more complex types like Maps. The conversion I do here is the spirit of the type (it maps strings to strings) rather than the strict type `Record<string, string>`.

It sounds like the trade off of having serialization to JSON is worth dealing with Objects directly to you?

sagerb: It's certainly a tradeoff. I'd rather we use JSON and the objects it supports as a transport mechanism, and then marshall/unmarshall into the data types which support the code the best. I know that this would typically introduce another layer, but I would prefer it to spreading the conversion around in all of the code that needs to use it. So, in my preference, the configuration object would have a map with the environment variables and secrets present in it.

Does that make sense?

dotNomad: I wrote out a long response not quite following, and then it finally clicked for me. Are you saying that you'd rather `selectedConfiguration.environment` be the Map?

`selectedConfiguration.secrets` can't necessary have the map because we need the representation in the file (the array of strings) and the state of the extension `secrets` (the Map of names to values) to be separated.

We can definitely convert the Configuration object into the most usable JavaScript objects. I think that is a pretty large refactor on the home store to take on right now though. The store in its current state is breaking up the configuration in a lot of places.

sagerb: Ok, thanks for thinking about this. I'm good with going forward with your current implementation. I'll approve.

sagerb (approved):
Concerns resolved.

kgartland-rstudio:
Verified the proper warning is shown if a Secret/Env Var collision occurs within the config. Deployment button remains active and deploys properly even if you forgo the warning.

Verified an error message is thrown when adding a Secret where an already-existing Environment Variable.

Tested on:
- [x] Positron on MacOS + Windows + Linux
- [x] VSCode MacOS + Windows + Linux

Screenshot 2024-10-07 at 11 14 39 AM
Screenshot 2024-10-07 at 11 16 08 AM

PR #2339, opened 2024-10-01
Only show select existing text if options exist

### LINKED ISSUE #2241: Confused "select new or existing" dialog
image

I think there are two different issues in here.

1. I have a project with a bunch of subdirectories. I had used an old version of Publisher to publish some of them by opening new windows in the subdirectories--this was back when Publisher didn't support subdirectories. For whatever reason, when I open this project now (in the top-level directory), none of the previously published things are found. I don't know whether you expect this to work or not; if not, that's fine, I'm an edge case not worth worrying about.
2. Assume this is the first time I'm publishing in a project--I've confirmed that it looks the same in a fresh project. The buttons and dialogs should be aware of the fact that there are 0 configs/deployments. How can I select an existing deployment? Why invite me to "pick one of the existing deployments below" when there are none?

dotNomad:
This PR corrects the details text for two of our `multiStepInputs` to only show the "pick one of the existing..." text when actual options exist to select. 

<details>
  <summary>Preview</summary>
  
CleanShot 2024-10-01 at 16 39 03@2x
CleanShot 2024-10-01 at 16 39 20@2x

</details> 

## Intent

Resolves #2241

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/src/multiStepInputs/selectNewOrExistingConfig.ts
+++ b/extensions/vscode/src/multiStepInputs/selectNewOrExistingConfig.ts
@@ -126,7 +126,9 @@ export async function selectNewOrExistingConfig(
       configFileListItems.push({
         iconPath: new ThemeIcon("plus"),
         label: createNewConfigurationLabel,
-        detail: "(or pick one of the existing deployments below)",
+        detail: configurations.length
+          ? "(or pick one of the existing configurations below)"
+          : undefined,
         picked: configurations.length ? false : true,
       });
 
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -1121,7 +1121,9 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       deploymentQuickPickList.push({
         iconPath: new ThemeIcon("plus"),
         label: createNewDeploymentLabel,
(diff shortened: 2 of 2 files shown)


extensions/vscode/src/multiStepInputs/selectNewOrExistingConfig.ts
```diff
@@ -126,7 +126,9 @@ export async function selectNewOrExistingConfig(
       configFileListItems.push({
         iconPath: new ThemeIcon("plus"),
         label: createNewConfigurationLabel,
-        detail: "(or pick one of the existing deployments below)",
+        detail: configurations.length
+          ? "(or pick one of the existing configurations below)"
```
dotNomad, line 130: This also corrects the text from `deployments` to `configurations` since configurations are listed. For context this is the multiStepInput when clicking "Select Active Configuration for Deployment" in the three dot context menu.

There shouldn't be a case where this detail is `undefined` since when a Deployment is missing its Configuration file and we click this context menu it immediately starts configuration creation. There isn't much harm in including the option though.

sagerb (approved):
LGTM

PR #2348, opened 2024-10-03
Allow deploy when the config is in error

### LINKED ISSUE #2345: Remove deploy disable state if configuration is in error

dotNomad:
This PR removes the disable state for the Deploy button when the configuration is in error.

There isn't anything preventing us from calling the Deploy API when this occurs. The warning is still present telling the user that their configuration is indeed in error, but now they can attempt a deploy and get an error message.

That looks something like this:

image

Our `ValidateTOMLFile` function will error when the Configuration doesn't match the schema, and has an error like the example above.

https://github.com/posit-dev/publisher/blob/7830e62c2bfe119d8d47c3dcc9dce19bacef5371/internal/schema/schema.go#L83

## Intent

Resolves #2345 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/DeployButton.vue
+++ b/extensions/vscode/webviews/homeView/src/components/DeployButton.vue
@@ -1,7 +1,7 @@
 <template>
   <vscode-button
     :data-automation="`deploy-button`"
-    :disabled="!haveResources || isConfigInError || home.publishInProgress"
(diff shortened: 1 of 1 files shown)


sagerb (approved):
LGTM

PR #2350, opened 2024-10-03
Add support for insecure TLS connections to Connect Servers

### LINKED ISSUE #2202: TLS verification off option for credentials
Add a Yes/No question to the create credential stepper. Add support to the credentials API and store. Pass it to the connect API client.

sagerb:
## Intent

This PR provides support for customer environments whose internal servers use self-signed SSL certificates, but for some reason, these certificates cannot be verified.

Resolves #2202

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

There are two agent APIs involved in the flows effected (`POST /test-credentials` and `POST /deployments/{target}`). The `test-credentials` API endpoint already supported an `insecure` option within it's body. I have replicated this into the `deployments/{target}` endpoint. The `insecure` option is now required for both APIs.

The credential object on the agent already supports an `insecure` attribute but it is not persisted. There was already code within the agent that implemented the insecure TLS functionality if set. Since we were looking for a global setting to enable `insecure TLS`, the API endpoint handlers each take the parameter passed in their body and update the memory copy of the account's insecure setting they are using to connect to the server. This allows this flag to be changed and immediately go into effect for the next test or deployment.

On the extension/UX side, a configuration setting for the Posit Publisher has been defined `Enable Insecure TLS`. You can access this setting within the settings UX for VSCode.
2024-10-03 at 1 21 PM

Access to the VSCode configuration setting is limited to the extension code (and is not available to the webview). It made most sense to have the current value available from the API client object, which is initialized by the extension code, but is available to the webview. It was not possible to simply add a method within the API client to retrieve the setting since both the extension and webview import the top-level `src/api` to access common types. While this can be changed in the future, I did not want to refactor the API to handle it within this PR. So instead, when the API client object is initialized, the extension passes in a callback which will return the current setting. 

As the appropriate API calls are made within the extension, they simply call the new method within the API client to retrieve the current value. Since there is no caching, the user should see the effect of a change immediately as long as the code retrieves this value at the last possible moment.

There are two scenarios in which you can see the effect of this PR.

The first is when creating a new credential. If validation of the server URL errors with a credential error, this error message is displayed with a link to the new configuration setting:
2024-10-03 at 1 34 PM

If the user clicks on the link within the error message, the multi-stepper stays open and the setting is made visible. They can alter this setting and then simply hit return within the multi-stepper to continue without TLS verification.
2024-10-03 at 1 36 PM

If verification is successful, then the next step for the multi-stepper is shown as expected.

Once a user has the credential established, then deployment attempts will use the current setting value from the configuration.

If a user changes this setting back to `disabled` after it was required to be `enabled` to create credentials, then they will see a deployment error. Currently it is shown as follows (although there is a triage item to be evaluated to improve this message/functionality).
2024-10-03 at 1 40 PM

## Automated Tests

Automated tests have been updated on the agent. The area in which the functionality has been implemented is within the extension framework, which will require integration tests to be written for it.

## Directions for Reviewers

To setup a connect server which fails TLS certificate validation:
- Within the connect repo's `ssl` directory, run `just certificates`
- It will create an `alwayssecure.rsc.crt` file which you can open and add it to your keychain.
- Start with setting this certificate to `Trust Always`
- Update your `etc/hosts` file to include a definition of `alwayssecure.rsc` to be `127.0.0.1`
- From the root directory of the connect repo, run `just start`
- Use a browser to connect to `https://alwayssecure.rsc:3443/`, and you should be able to access the Connect dashboard without issue.
- Change the certificate to `Never Trust` within the keychain
- You'll need to open a new browser instance or window, but then attempt to connect to the connect server once again. You should see the warnings or stop pages for an insecure setting. Chrome will allow you bypass it, but Safari stops you cold (at least with my system settings).

Now you can create a credential within the publisher extension and verify the new functionality. Be sure to:
1. Create a credential to a different server than `alwayssecure.rsc`, to confirm regression
2. Create a credential to `alwayssecure.rsc` when it is not trusted. Validate the error is encountered during the new Credential flow, and confirm the link brings up the settings page. Enable the option and then press return on the multi-stepper, and it should proceed forward to the API key input. Complete the credential creation.
3. With a credential in place, try deploying a project with the `insecure TLS` option still enabled. It should deploy successfully.
4. Change the `insecure TLS` option to be disabled and try deploying once again. It should fail with the error displayed.

NOTE: if you have any issues with recreating an insecure server, check with either Bill or Kevin (who have set this up). Bill was unable to use YETI to do this, so you will need to use the direct starting of a server within the connect repo.

## Checklist

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -33,6 +33,18 @@
   ],
   "main": "./dist/extension.js",
   "contributes": {
+    "configuration": [
+      {
+        "title": "Posit Publisher",
+        "properties": {
+          "positPublisher.enableInsecureTLS": {
+            "markdownDescription": "Enable/disable skipping TLS certificate validation for connections to servers.\n\nWhen **disabled** (default), TLS connections initiated by Posit Publisher will only be established if the SSL certificate associated with the target server can be validated with the configured certificate authority. \n\nWhen **enabled**, the SSL certificate validation will be skipped. Only **enable** this setting if you are accessing fully trusted servers as it could introduce security vulnerabilities since server identity is not being verified.",
+            "type": "boolean",
+            "default": false
+          }
+        }
+      }
+    ],
     "commands": [
       {
         "command": "posit.publisher.init-project",
--- a/extensions/vscode/src/api/client.ts
+++ b/extensions/vscode/src/api/client.ts
@@ -11,6 +11,8 @@ import { Secrets } from "./resources/Secrets";
 import { EntryPoints } from "./resources/Entrypoints";
 import * as Entities from "entities";
 
+export type GetInsecureSettingFn = () => boolean;
+
 class PublishingClientApi {
   private client;
 
@@ -22,13 +24,28 @@ class PublishingClientApi {
   secrets: Secrets;
   apiServiceIsUp: Promise<boolean>;
   entrypoints: EntryPoints;
+  // In order to allow this object to be used both within the VSCode extension
+  // code as well as the hosted webview, we need to isolate imports which are
+  // not allowed in both. In this case, vscode functionality is needed to
+  // access the extension's configuration setting. So we'll initialize this
+  // object with a callback which can provide the separation, while maintaining
+  // our ability to have this object accessible from both "domains".
+  getInsecureSetting: GetInsecureSettingFn;
 
-  constructor(apiBaseUrl: string, apiServiceIsUp: Promise<boolean>) {
+  constructor(
+    getInsecureSettingFn: GetInsecureSettingFn,
+    apiBaseUrl: string,
+    apiServiceIsUp: Promise<boolean>,
+  ) {
     this.client = axios.create({
       baseURL: apiBaseUrl,
     });
     this.client.interceptors.request.use((request) => {
       request.ts = performance.now();
+      // request.params = {
+      //   ...request.params,
+      //   insecure: this.getInsecureSetting(),
+      // };
       return request;
     });
 
@@ -55,6 +72,7 @@ class PublishingClientApi {
     this.packages = new Packages(this.client);
     this.secrets = new Secrets(this.client);
     this.entrypoints = new EntryPoints(this.client);
+    this.getInsecureSetting = getInsecureSettingFn;
   }
 
   logDuration(response: AxiosResponse<any, any>) {
@@ -77,10 +95,11 @@ let api: PublishingClientApi | undefined = undefined;
 // NOTE: this function must be called ahead of useApi()
 // so that the class is properly instantiated.
 export const initApi = (
+  insecureSettingFn: GetInsecureSettingFn,
   apiServiceIsUp: Promise<boolean>,
   apiBaseUrl: string = "/api",
 ) => {
-  api = new PublishingClientApi(apiBaseUrl, apiServiceIsUp);
+  api = new PublishingClientApi(insecureSettingFn, apiBaseUrl, apiServiceIsUp);
 };
 
 // NOTE: initApi(...) must be called ahead of the first time
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -2,7 +2,7 @@
 
 import * as path from "path";
 
-import { ExtensionContext } from "vscode";
+import { ExtensionContext, Uri } from "vscode";
 
 import { HOST } from "src";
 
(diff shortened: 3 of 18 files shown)


extensions/vscode/package.json
```diff
@@ -33,6 +33,18 @@
...
   "contributes": {
+    "configuration": [
+      {
+        "title": "Posit Publisher",
+        "properties": {
+          "positPublisher.enableInsecureTLS": {
+            "markdownDescription": "Enable/disable skipping TLS certificate validation for connections to servers.\n\nWhen **disabled** (default), TLS connections initiated by Posit Publisher will only be established if the SSL certificate associated with the target server can be validated with the configured certificate authority. \n\nWhen **enabled**, the SSL certificate validation will be skipped. Only **enable** this setting if you are accessing fully trusted servers as it could introduce security vulnerabilities since server identity is not being verified.",
+            "type": "boolean",
+            "default": false
+          }
```
jonkeane, line 44:
```suggestion
          "positPublisher.verifyCertificates": {
            "markdownDescription": "Verify TLS certificates for connections. Only **disable** this setting if you experience certificate verification errors and the Posit Connect server is using a self-signed, expired, or otherwise misconfigured certificate.",
            "type": "boolean",
            "default": true
          }
```

This is a bunch of flipping of the logic (which will need to happen elsewhere too) but attempting to reduce the number of booleans/negations one has to keep track of. I also attempted to match the (much terser) style of option descriptions I see in most other extensions and be as direct as possible.

extensions/vscode/src/api/client.ts
```diff
@@ -77,10 +95,11 @@ let api: PublishingClientApi | undefined = undefined;
 // NOTE: this function must be called ahead of useApi()
 // so that the class is properly instantiated.
 export const initApi = (
+  insecureSettingFn: GetInsecureSettingFn,
```
dotNomad, line 98: It is not clear to me why this is part of the API client.

Why not keep the parameters for the API calls that care about the insecure setting, and use a helper function in the extension to determine what to send to those functions like you are already doing?

```
workspace.getConfiguration("positPublisher").get<boolean>("enableInsecureTLS")
```

extensions/vscode/src/commands.ts
```diff
@@ -19,3 +19,8 @@ export const create = async (
 const getExecutableBinary = (context: ExtensionContext): string => {
   return path.join(context.extensionPath, "bin", "publisher");
 };
+
+const args = ["@ext:posit.publisher"];
+export const openConfigurationCommand = Uri.parse(
+  `command:workbench.action.openSettings?${encodeURIComponent(JSON.stringify(args))}`,
+);
```
dotNomad, line 26: Nice find to be able to do this with a link!

extensions/vscode/src/api/client.ts
sagerb: As discussed, I'll move the helper functionality out.

extensions/vscode/src/commands.ts
sagerb: Thanks!

extensions/vscode/package.json
sagerb: Will be changing as requested.

sagerb: After further investigation, the impact of refactoring "insecure" to "secure" throughout the agent would, IMO, be more changes to the backend than we want to take on at this time. It was not just a simple change to the API we were calling, but instead, to be consistent, the Account interface needed to be changed, which then impacted other operational code, other APIs, and many tests. 

Instead, @dotNomad and I refined the plan to redefine the configuration setting as requested, but consider it as a presentation layer. Under that presentation layer, the helper function that supported retrieval of the setting understood the logic flip. Everything else will remain consistent within the code from the back-end and front-end, using the existing definition of `insecure`.

jonkeane: > Instead, @dotNomad and I refined the plan to redefine the configuration setting as requested, but consider it as a presentation layer. Under that presentation layer, the helper function that supported retrieval of the setting understood the logic flip. Everything else will remain consistent within the code from the back-end and front-end, using the existing definition of insecure.

If I'm following this correctly (the code is the ultimate definition of this of course, LMK when that is here and I'll look to confirm that matches the mental model I have from this description), that sounds good. 

Consistency is nice, but we don't have the luxury of infinite time of doing a lot of work to make things perfectly consistent when that's difficult. So absolutely 💯 of deciding a pragmatic placer where the logic flipping boundary. The most important part is what we surface to users — that must be as clear as possible. If underneath that in our own code have to flip and remember that, that's ok.

sagerb:
Requested changes pushed up in the latest commit.

dotNomad (approved):
Thanks for those additional adjustments. This is look great!

PR #2353, opened 2024-10-04
Show some details for a TOML parsing error

### LINKED ISSUE #2330: Display of Error when configuration file has an invalid TOML error

sagerb:
## Intent

With this PR, configuration files which are unable to be parsed by the agent due to a TOML parsing error, will now indicate that the file was unable to be parsed and what line the error has been detected.

Resolves #2330 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The ConfigurationError object contains an AgentError. I have redefined an AgentError to be either an `AgentErrorTypeUnknown` or a `AgentErrorInvalidTOML`, with type guards defined for both.

When the active configuration has a TOML parsing error, the `EvenEasierDeploy` component will now display a brief statement as seen below. It includes an anchor to edit the configuration and the cursor will be positioned at the line and character location reported within the TOML parsing error.

2024-10-04 at 12 32 PM

## Directions for Reviewers

1. Confirm that if there is no error, the UX displays as expected prior to these changes.
2. Edit the active configuration, making the file invalid in some way, and save the file. Verify that the new error is displayed. Close the configuration file window and click on the `edit configuration` link. Verify that the correct configuration file is opened and that the cursor is positioned on the line indicated in the error message.
3. Correct the active configuration and include overlapping variable definitions for secrets and environment variables. Confirm that the proper warning message is displayed. Close the configuration file and then click the `edit configuration` link and confirm that the correct file is opened, with the cursor located at the top of the file.

## Checklist

### DIFF
--- a/extensions/vscode/src/api/types/error.ts
+++ b/extensions/vscode/src/api/types/error.ts
@@ -26,8 +26,8 @@ export type AgentErrorInvalidTOML = AgentErrorBase & {
   data: {
     problem: string;
     file: string;
-    line: string;
-    column: string;
+    line: number;
+    column: number;
   };
 };
 
--- a/extensions/vscode/src/api/types/error.ts
+++ b/extensions/vscode/src/api/types/error.ts
@@ -1,10 +1,38 @@
 // Copyright (C) 2023 by Posit Software, PBC.
 
-export type AgentError = {
+type AgentErrorBase = {
   code: string;
   msg: string;
   operation: string;
+};
+
+export type AgentError = AgentErrorTypeUnknown | AgentErrorInvalidTOML;
+
+export type AgentErrorTypeUnknown = AgentErrorBase & {
   data: {
     [key: string]: unknown;
   };
+};
+
+// This represents the Agent Errors we haven't been able to
+// narrow down to a specific type.
+export const isAgentErrorTypeUnknown = (
+  e: AgentError,
+): e is AgentErrorTypeUnknown => {
+  return !isAgentErrorInvalidTOML(e);
+};
+
+export type AgentErrorInvalidTOML = AgentErrorBase & {
+  data: {
+    problem: string;
+    file: string;
+    line: string;
+    column: string;
+  };
+};
+
+export const isAgentErrorInvalidTOML = (
+  e: AgentError,
+): e is AgentErrorInvalidTOML => {
+  return e.code === "invalidTOML" || e.code === "unknownTOMLKey";
 };
--- a/extensions/vscode/src/commands.ts
+++ b/extensions/vscode/src/commands.ts
@@ -43,7 +43,8 @@ export const openFileInEditor = async (
     }
   }
 
+  // we'll only indicate a location, if one was passed to us.
   await window.showTextDocument(Uri.file(path), {
-    selection: new Range(start, end),
+    selection: options && options.selection ? new Range(start, end) : undefined,
   });
 };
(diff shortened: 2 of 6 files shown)


extensions/vscode/src/commands.ts
```diff
@@ -19,3 +20,30 @@ export const create = async (
...
+        options.selection.end.line,
+        options.selection.end.character,
+      );
+    } else {
+      end = start;
+    }
+  }
+
+  await window.showTextDocument(Uri.file(path), {
+    selection: new Range(start, end),
```
dotNomad, line 47: Always including the `selection` has a bit of an odd behavior when this is called without `options: { selection }` and the file is open. It moves the cursor up to `0:0` where previously the file just remained open.

Perhaps we can only call with `selection` if we were passed a selection?

dotNomad (commented):
This is look pretty great! Getting the opening of the configuration at the error position is a really nice touch 👍 

I did think that the default behavior of opening the Configuration (like through the edit button in the top toolbar) was a bit strange.

Noticed there are a few conflicts to take care of before merging as well. If you'd like me to take another look after those are addressed just let me know 👍

extensions/vscode/src/commands.ts
sagerb: Sounds like a great idea. Done!

dotNomad (approved):
Thanks for that change this looks great 😎

extensions/vscode/src/api/types/error.ts
```diff
@@ -1,10 +1,38 @@
...
+export type AgentErrorInvalidTOML = AgentErrorBase & {
+  data: {
+    problem: string;
+    file: string;
+    line: number;
+    column: number;
+  };
+};
+
+export const isAgentErrorInvalidTOML = (
```
marcosnav, line 34: With these new AgentError typing helpers we now have 3 files handling incoming agent errors:
1. `utils/errorTypes.ts` for incoming agent errors via HTTP responses (AxiosErrors)
2. `eventErrors.ts` for incoming agent errors via SSE - EventStreamMessages
3. This one

It'd be great if we can consolidate or minimize areas handling errors in different ways, I don't have a suggestion right away, but it might be worth sharing ideas to prevent confusion in the future.

Is `home.selectedConfiguration.error` an event stream error?

marcosnav (commented):
Looking good so far! I like this new approach, just one concern about us creating new ways to handle errors.

extensions/vscode/src/api/types/error.ts
sagerb: No, it is present within the configuration object returned from the configuration query API. Those objects are either `Configurations` or `ConfigurationErrors`.  `ConfigurationErrors` have an `AgentError`, which is present.

The base `AgentError` object is defined within the `error.ts` file. It is the core concept. It feels like to me that we have the concept of types of `AgentErrors`, in which we have some types which include transport-specific info.

I totally agree that we need to find a good way to consolidate these. Maybe we can hop on a call real quick?

dotNomad: After a quick discussion we agreed that a bit of error code consolidation would be great looking to our more complicated error handling (like this one) and trying to unify how we handle errors.

We both agreed that would be a good clean up task for a milestone in the future. After we work with our error patterns a bit more, and know which to unify on.

PR #2354, opened 2024-10-04
Set R and Python Config defaults

### LINKED ISSUE #2328: Opening Folder after scanning for R dependencies?
Within the `shinyapp` sample-content project, I loaded a valid deployment but then removed the `package-file` and `package-manager` entries. I then deleted the project's `renv.lock` file.

This simulated me manually creating an R config, without all of fields (these two are not required).

The extension properly showed that I had no requirements file and offered to scan. 

I then hit scan on the R project and was presented with this:
Image

It seems that the action after creating a `renv.lock` file was to open the file `shiny-app` which is a directory. Because there is no entry within the config for the package file, the agent is probably taking the project-dir and appending an empty string to it, before attempting to open.

I believe the agent should be updating config to have both of these entries updated when we initiated scan.
```
package_file = 'renv.lock'
package_manager = 'renv'
```

NOTE: `renv.lock` was created correctly.

dotNomad:
This PR sets the the defaults for the `R` and `Python` structs for the Config after we fetch the details from the file.

Our schema defaults `package_manager` and `package_file` for both sections, but when doing a `GET /api/configurations/$name` we would get back empty strings for those values with a configuration file that looked like:

```
...
[r]
version = '4.3.1'

[python]
version = '3.8.5'
```

This PR avoids us from defaulting in the frontend and moving the onus to the backend to tell us how it interprets the Configuration

## Intent

Resolves #2328

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The error from #2328 was caused by:

https://github.com/posit-dev/publisher/blob/67706f071bff37b69e5e94bbb6ab4e521d25b31b/extensions/vscode/src/views/homeView.ts#L685

Assuming that if `package_file` was unset it was `undefined` when in fact it is `""` causing it to continue in the flow and hit an overwrite prompt for the wrong file URI, then opening the wrong file URI. 

The backend handles this well since it is using the default value of `renv.lock` but the frontend isn't being communicated that is the case via the Configuration.

The idea here was to correct that by inserting the defaults.

We investigated if there was a way to default in the struct tags, but there is not.

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -680,11 +680,14 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       return;
     }
 
-    const relPathPackageFile =
-      activeConfiguration.configuration.python?.packageFile;
-    if (relPathPackageFile === undefined) {
+    // Project is not configured for Python so cannot scan
+    // Scan button is not visible when this is the case
+    if (!activeConfiguration.configuration.python) {
       return;
     }
+
+    const relPathPackageFile =
+      activeConfiguration.configuration.python.packageFile;
 
     const fileUri = Uri.joinPath(
       this.root.uri,
@@ -747,11 +750,13 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       return;
     }
 
-    const relPathPackageFile =
-      activeConfiguration?.configuration.r?.packageFile;
-    if (relPathPackageFile === undefined) {
+    // Project is not configured for R so cannot scan
+    // Scan button is not visible when this is the case
+    if (!activeConfiguration.configuration.r) {
       return;
     }
+
+    const relPathPackageFile = activeConfiguration.configuration.r.packageFile;
 
     const fileUri = Uri.joinPath(
       this.root.uri,
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -69,6 +69,7 @@ func FromFile(path util.AbsolutePath) (*Config, error) {
 	if err != nil {
 		return nil, err
 	}
+	cfg.FillDefaults()
 	cfg.Comments, err = readLeadingComments(path)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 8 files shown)


dotNomad:
`validate` and `hasParameters` always seem to have the correct default values even if the TOML file doesn't contain them. I'm unsure why those are defaulted, but not the `package_file` and `package_manager` values. Any assistance on understanding that would be super helpful.

mmarchetti:
> I'm unsure why those are defaulted

`hasParameters` default is false, which is go's "zero value" for booleans so it's set that way on construction. `validate` defaults to true, and we set that on construction as well; `FromFile` will overwrite it with the value in the file, if present.

We can't use the same trick for these fields, because they live in optional subsections that can be null. The implementation here seems like it will work. An alternative would be to default these fields during the inspection process, but that leaves open the possibility of a user removing those lines from the file, putting us right back here.

dotNomad:
> `hasParameters` default is false, which is go's "zero value"

Ah! That is the bit I was missing. Thanks!

dotNomad:
Thanks for the extra detail @mmarchetti and @marcosnav. I'll move this along by adding tests and some corrections to the frontend types.

dotNomad:
Alright this should be ready for review!

extensions/vscode/src/views/homeView.ts
```diff
@@ -680,12 +680,15 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       return;
     }
 
-    const relPathPackageFile =
-      activeConfiguration.configuration.python?.packageFile;
-    if (relPathPackageFile === undefined) {
+    // Project is not configured for Python so cannot scan
+    // Scan button is not visible when this is the case
+    if (!activeConfiguration.configuration.python) {
```
marcosnav, line 685: Double checking here, I'm not familiar with configuration defaults but is there the possibility for `activeConfiguration.configuration.python` to be an empty object at any point giving a false positive?

extensions/vscode/src/views/homeView.ts
```diff
@@ -747,12 +750,14 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       return;
     }
 
-    const relPathPackageFile =
-      activeConfiguration?.configuration.r?.packageFile;
-    if (relPathPackageFile === undefined) {
+    // Project is not configured for R so cannot scan
+    // Scan button is not visible when this is the case
+    if (!activeConfiguration.configuration.r) {
```
marcosnav, line 755: same here, could `r` be an empty object bringing false positives?

internal/config/config.go
```diff
@@ -108,6 +109,28 @@ func (cfg *Config) WriteFile(path util.AbsolutePath) error {
 	return cfg.Write(f)
 }
 
+func (cfg *Config) FillDefaults() error {
```
marcosnav, line 112: No need to return an error on this func

marcosnav (approved):
LGTM, with a few questions that might be worth to double check

extensions/vscode/src/views/homeView.ts
dotNomad: I don't believe so. The schema is invalid if the Python configuration doesn't have a version, and when it only has that the `packageFile` and `packageManager` become the default value. So it cannot be an empty object without also being a configuration error.

The same goes for the R section

PR #2355, opened 2024-10-05
R Version Detection: Loop over lines rather than just first line of 'R --version' output

### LINKED ISSUE #2343: R Version not being detected
We had a report from Trevor Nederlof (SE) hitting this error when deploying R content on Workbench:

https://positpbc.slack.com/archives/C05D7NZD52S/p1727877579376109

> time=2024-10-02T13:56:32.858Z level=ERROR msg="could not run R executable &#39;/usr/local/bin/R&#39;: couldn&#39;t parse R version from output: WARNING: ignoring environment value of R_HOME" method=POST url="/api/inspect?dir=app&entrypoint=app.R" error="could not run R executable '/usr/local/bin/R': couldn't parse R version from output: WARNING: ignoring environment value of R_HOME"

When he runs:
`/usr/local/bin/R --version`
It returns:
```
WARNING: ignoring environment value of R_HOME
R version 4.4.0 (2024-04-24) -- "Puppy Cup"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under the terms of the
GNU General Public License versions 2 or 3.
For more information about these matters see
https://www.gnu.org/licenses/.
```
His value of `$R_HOME` is ```/opt/R/4.4.1/lib/R```

I suspect that warning is thrown when the versions of $R_HOME and R on PATH don't match.

FWIW, when I run the same command on the same Workbench instance it works as expected:
```
$ /usr/local/bin/R --version
R version 4.4.1 (2024-06-14) -- "Race for Your Life"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under the terms of the
GNU General Public License versions 2 or 3.
For more information about these matters see
https://www.gnu.org/licenses/.
```
My R_HOME:
```
$ echo $R_HOME
/opt/R/4.4.1/lib/R
```

We must be fetching the first line from the output of `R --version` to detect R but if that warning is thrown, we can't find it.  Is there a better way to get the R Version?

sagerb:
…irst line.

## Intent

Handle the case where the user has defined `R_HOME` in their environment. 

NOTE: The objective here is to determine the version of R based on the one that executes off of the path, NOT the one which may be at the `R_HOME` location.

Resolves #2343 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The existing code was only running regEx match against the first line in the output.

Fix simply loops over every line of the output and uses the first match.

Each line is logged as examined, which continues what was happening before (and helped us diagnose this so quickly).

## Automated Tests

Unit tests updated.

## Directions for Reviewers

**Confirm the fix for the bug**
- You'll need to define the `R_HOME` environment variable in your base user profile (I inserted into my `~.zshenv` file.
- start a new terminal and examine the output of `R --version` to confirm that the output includes the warning that R_HOME is defined.
- Make sure no instances are running of VSCode before you start one up.
- Open an R project - I used shinyapp. 
- Delete the `renv.lock` file if you have one.
- Use the publisher's functionality to scan for R dependencies.
- Confirm the version in the `renv.lock` file is the same version as expected.
- Examine the publisher logs in the extension's output log. Search for `Parsing line for R version` to determine if the code encountered the test case where the first line examined included a warning, and if the correct version was extracted from one of the following lines. 

** Confirm regression **
- Remove the environment variable from your environment. 
- Shutdown all instances of VSCode
- Create a new terminal instance and confirm `R_HOME` is no longer present within the environment variables (`env | grep R_HOME`)
- Start up an instance of VSCode
- Repeat the steps to re-open the R project used previously, delete `renv.lock` file, and use the extension to scan for R dependencies.
- Confirm that `renv.lock` is created with the expected version of R
- Confirm within the Publisher's log (output window), that the command output did NOT include the warning previously seen when the environment variable was active.

## Checklist

### DIFF
--- a/internal/inspect/r.go
+++ b/internal/inspect/r.go
@@ -176,15 +176,19 @@ func (i *defaultRInspector) getRVersion(rExecutable string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	line := strings.SplitN(string(append(output, stderr...)), "\n", 2)[0]
-	m := rVersionRE.FindStringSubmatch(line)
-	if len(m) < 2 {
-		return "", fmt.Errorf("couldn't parse R version from output: %s", line)
+	lines := strings.SplitN(string(append(output, stderr...)), "\n", -1)
+	for _, l := range lines {
+		i.log.Info("Parsing line for R version", "l", l)
+		m := rVersionRE.FindStringSubmatch(l)
+		if len(m) < 2 {
+			continue
+		}
+		version := m[1]
+		i.log.Info("Detected R version", "version", version)
+		rVersionCache[rExecutable] = version
+		return version, nil
 	}
-	version := m[1]
-	i.log.Info("Detected R version", "version", version)
-	rVersionCache[rExecutable] = version
-	return version, nil
+	return "", fmt.Errorf("couldn't parse R version from command output (%s --version)", rExecutable)
 }
 
 var renvLockRE = regexp.MustCompile(`^\[1\] "(.*)"`)
--- a/internal/inspect/r_test.go
+++ b/internal/inspect/r_test.go
@@ -46,7 +46,14 @@ func (s *RSuite) TestNewRInspector() {
 	s.Equal(log, inspector.log)
 }
 
-const rOutput = `R version 4.3.0 (2023-04-21) -- "Already Tomorrow"
+type OutputTestData struct {
+	output, expectedVersion string
+}
+
+func getOutputTestData() []OutputTestData {
+	data := []OutputTestData{
+		// typical output from command `r --version`
+		{`R version 4.3.0 (2023-04-21) -- "Already Tomorrow"
 Copyright (C) 2023 The R Foundation for Statistical Computing
 Platform: x86_64-apple-darwin20 (64-bit)
 
@@ -55,9 +62,9 @@ You are welcome to redistribute it under the terms of the
 GNU General Public License versions 2 or 3.
 For more information about these matters see
 https://www.gnu.org/licenses/.
-`
-
-const rOutputWithWarnings = `WARNING: ignoring environment value of R_HOME
+`, "4.3.0"},
+		// output when there is a warning
+		{`WARNING: ignoring environment value of R_HOME
 R version 4.3.3 (2024-02-29) -- "Angel Food Cake"
 Copyright (C) 2024 The R Foundation for Statistical Computing
 Platform: x86_64-apple-darwin20 (64-bit)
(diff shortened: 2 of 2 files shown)


internal/inspect/r_test.go
```diff
@@ -57,6 +57,17 @@ For more information about these matters see
 https://www.gnu.org/licenses/.
 `
 
+const rOutputWithWarnings = `WARNING: ignoring environment value of R_HOME
+R version 4.3.3 (2024-02-29) -- "Angel Food Cake"
```
jonkeane, line 61: I know that this is an exact example that we saw in real life, but if we're iterating over lines and want to ensure that we do, in fact, loop over them we should add a few more lines up above this too (and that's a good reminder to future-us that this is not about finding the version in the first or second lines only).

internal/inspect/r_test.go
```diff
@@ -73,6 +84,22 @@ func (s *RSuite) TestGetRVersionFromExecutable() {
...
+	i := NewRInspector(s.cwd, rPath.Path, log)
+	inspector := i.(*defaultRInspector)
+
+	executor := executortest.NewMockExecutor()
+	executor.On("RunCommand", rPath.String(), []string{"--version"}, mock.Anything, mock.Anything).Return([]byte(rOutputWithWarnings), nil, nil)
+	inspector.executor = executor
+	version, err := inspector.getRVersion(rPath.String())
+	s.NoError(err)
+	s.Equal("4.3.3", version)
+}
```
jonkeane, line 101: Another test case that is worth at least considering, and possibly adding is what happens if the output has multiple strings that match:

```
R version 4.4.0 (2024-04-24) -- "Puppy Cup"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: aarch64-apple-darwin20

R version 3.6.0 (2019-04-19) -- "blah blah"
```

In principle this should _never happen_ but now that we're iterating overlines I'm curious what would happen and if we want to proactively state + test it's the first (or the last, though I think first is probably best) instance?

sagerb:
Additional requested tests added in latest commit (also switched to looping over test data).

internal/inspect/r.go
```diff
@@ -176,15 +176,19 @@ func (i *defaultRInspector) getRVersion(rExecutable string) (string, error) {
...
-	if len(m) < 2 {
-		return "", fmt.Errorf("couldn't parse R version from output: %s", line)
-	}
-	version := m[1]
-	i.log.Info("Detected R version", "version", version)
-	rVersionCache[rExecutable] = version
-	return version, nil
+	lines := strings.SplitN(string(append(output, stderr...)), "\n", -1)
+	for _, l := range lines {
+		i.log.Info("Parsing line for R version", "l", l)
```
marcosnav, line 181: nit: This log might be better as `Debug` instead of `Info`

marcosnav (approved):
💯

internal/inspect/r.go
sagerb: It looks like the existing pattern for this type of tracing within the `r.go` and `python.go` inspection code is `Info`. So I stayed with that to be consistent.

PR #2359, opened 2024-10-07
Add more specificity to git ignore files pattern documentation

dotNomad:
Improved documentation for the `files` attribute in our Configuration files, being a little bit more specific about it using the gitignore syntax.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [x] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -37,7 +37,9 @@ Description for this content. It may span multiple lines and be up to 4000 chara
 
 #### files
 
-Project-relative paths of the files to be included in the deployment. Wildcards are accepted, using [`.gitignore` syntax](https://git-scm.com/docs/gitignore).
+Project-relative paths of the files to be included in the deployment.
+Uses [`.gitignore` syntax](https://git-scm.com/docs/gitignore) for file include
+patterns, like wildcards.
 
 #### has_parameters
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,6 +41,23 @@ Project-relative paths of the files to be included in the deployment.
 Uses [`.gitignore` syntax](https://git-scm.com/docs/gitignore) for file include
 patterns, like wildcards.
 
+Here are some examples given an example project structure:
+
+```
+project-root/
+|
+├── dir/
+│   ├── util.py
+│   └─  helper.py
+│
+└── app.py
+```
+
+- To include all Python files in your project, you can use a wildcard pattern.
+  `*.py` would include all `.py` files in any directory.
+- `/dir/util.py` would include the specific `util.py` file in the `dir`
+  directory under the root of the project to be published.
+
 #### has_parameters
 
 `true` if this is a report that accepts parameters.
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -37,7 +37,9 @@ Description for this content. It may span multiple lines and be up to 4000 chara
 
 #### files
 
-Project-relative paths of the files to be included in the deployment. Wildcards are accepted, using [`.gitignore` syntax](https://git-scm.com/docs/gitignore).
+Project-relative paths of the files to be included in the deployment.
+Uses [`.gitignore` syntax](https://git-scm.com/docs/gitignore) for file include
+patterns, like wildcards.
```
jonkeane, line 42:
```suggestion
Uses [`.gitignore` syntax](https://git-scm.com/docs/gitignore) for file include
patterns, like wildcards. For example, `/some_dir/util.py` references a `util.py` file under the directory `some_dir` that is just under the root of the project to be published.
```

We might need to expand / explain more what "root of the project to be published" means (or link to the relevant section in our docs). But giving a concrete example of the syntax most likely to be confusing would be great.

sagerb: Perhaps we should also consider including an example when the forward slash is not used, and the pattern will match any file located in the project subdirectories?

dotNomad: I expanded this a bit to include an example project structure, and a wildcard example. Let me know what you think

dotNomad: Maybe something like the below?

"- `app.py` will match any file, in any directory, named `app.py`"

docs/configuration.md
```diff
@@ -37,7 +37,26 @@ Description for this content. It may span multiple lines and be up to 4000 chara
...
+│   ├── util.py
+│   └─  helper.py
+│
+└── app.py
+```
+
+- To include all Python files in your project, you can use a wildcard pattern.
+  `*.py` would include all `.py` files in any directory.
+- `/dir/util.py` would include the specific `util.py` file in the `dir`
+  directory under the root of the project to be published.
```
jonkeane, line 59:
```suggestion
- `/dir/util.py` would include the specific `util.py` file in the `dir`
  directory under the root of the project to be published.
- To include the file `app.py` use `/app.py`
```

One more example that shows that it's not only about directories that you need the leading `/`

jonkeane (approved):
I'm fine with this as is, have one more suggestion. Also opening the link to gitscm and 😮‍💨  that's some documentation!

dotNomad:
> I'm fine with this as is, have one more suggestion.

Thanks for the suggestion. Added in https://github.com/posit-dev/publisher/pull/2359/commits/6494e87936dc03b57cfaf694ef7323fe543e1cf9, and then I'll get this merged once it passes CI

> Also opening the link to gitscm and 😮‍💨 that's some documentation!

It is... 🤔  thorough

PR #2360, opened 2024-10-07
Be silent on renv.lock file errors until trying to deploy

### LINKED ISSUE #2327: Error 500 reported if renv.lock is empty file
To reproduce, open an R Project deployment and delete the contents of the renv.lock file.

Image

marcosnav:
## Intent

This PR aims to stop showing or blocking the user on `renv.lock` file errors. Only showing the error when proceeding to deploy.

Fixes #2327 

**Screenshot of the new friendlier message**
Screen Shot 2024-10-07 at 14 08 37

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Having a more structured and identifiable error to handle better the messaging when R packages cannot be scanned from the existing `renv.lock` file. Now logging out the errors as needed but not blocking the user until the `renv.lock` file scanning is absolutely required.

## Automated Tests

Updated and added unit tests as needed.

## Directions for Reviewers

Having an empty `renv.lock`, you shouldn't see an `renv.lock` related error message until trying to deploy.

## Checklist

### DIFF
--- a/extensions/vscode/src/eventErrors.ts
+++ b/extensions/vscode/src/eventErrors.ts
@@ -14,23 +14,37 @@ export function isCodedEventErrorMessage(
   return msg.errCode !== undefined;
 }
 
-type deploymentEvtErr = {
+type baseEvtErr = {
   dashboardUrl: string;
   localId: string;
   message: string;
   error: string;
 };
 
+type lockfileReadingEvtErr = baseEvtErr & {
+  lockfile: string;
+};
+
 export const isEvtErrDeploymentFailed = (
   emsg: EventStreamMessageErrorCoded,
-): emsg is EventStreamMessageErrorCoded<deploymentEvtErr> => {
+): emsg is EventStreamMessageErrorCoded<baseEvtErr> => {
   return emsg.errCode === "deployFailed";
+};
+
+export const isEvtErrRenvLockPackagesReadingFailed = (
+  emsg: EventStreamMessageErrorCoded,
+): emsg is EventStreamMessageErrorCoded<lockfileReadingEvtErr> => {
+  return emsg.errCode === "renvlockPackagesReadingError";
 };
 
 export const handleEventCodedError = (
   emsg: EventStreamMessageErrorCoded,
 ): string => {
   if (isEvtErrDeploymentFailed(emsg)) {
+    return emsg.data.message;
+  }
+
+  if (isEvtErrRenvLockPackagesReadingFailed(emsg)) {
     return emsg.data.message;
   }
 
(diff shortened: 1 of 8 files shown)


extensions/vscode/src/eventErrors.ts
```diff
@@ -14,26 +14,40 @@ export function isCodedEventErrorMessage(
...
+
 export const handleEventCodedError = (
   emsg: EventStreamMessageErrorCoded,
 ): string => {
   if (isEvtErrDeploymentFailed(emsg)) {
     return emsg.data.message;
   }
 
+  if (isEvtErrRenvLockPackagesReadingFailed(emsg)) {
+    return emsg.data.message;
```
dotNomad, line 48: Minor nit: the error we show in the notification and the sidebar would be more user friendly if the path wasn't an absolute file path, but a relative path based on their workspace root.

It makes sense for the error in the deployment record to be an absolute file path since it doesn't know where the workspace is in VS Code (and it could change between work sessions).

To accomplish this we can do something like:

```suggestion
    return `Could not scan R packages from R package file: ${relativePath(Uri.parse(emsg.data.lockfile))}`;
```

This doesn't change the sidebar though. That would require a special case in `EvenEasierDeploy`.

dotNomad (commented):
Noting this down that I did occasionally get a deploying notification and the "Could not scan R packages" notification at the same time during testing. 

Looks like there might be a race condition where the deployment is showing as in progress (and can be cancelled) even if it fails. It may be related to just how fast this fails. The Deploy button never stayed disabled, just the notification stayed up. Perhaps we are a bit slow in creating that notification compared to how quickly the failure occurs.

I don't think that should hold up the PR, but perhaps a follow-up issue should be created. What do you think @marcosnav?

---

Noted one UX nitpick down, approving and can re-look at the PR upon request if any changes are made.

dotNomad (approved):
Oops, meant to hit "Approve"

extensions/vscode/src/eventErrors.ts
marcosnav: Ah, but it seems only the extension code knows the current VSCode workspace tree. I now better understand your point.

dotNomad:
> Both things are worth improving but might be worth to address in follow-up issues since are not considered blockers

Fantastic choice. Thanks for looking into them to see if they were easy things we could do as part of the PR. Would you mind making the follow-up issues with the information you've gleaned?

PR #2364, opened 2024-10-08
use 1.93.0 in vscode tests

kgartland-rstudio:
## Intent
Pin vscode to 1.93.0 for tests for now. It seems webdriver.io has broken with the 1.94.1 release today.  This will be needed to unblock CI.

I'll keep reaching out to webdriver to see what we can do.


sagerb (approved):
Hopefully, we won't be pinned for too long.

Thanks! LGTM.

PR #2378, opened 2024-10-09
Better messaging for missing `requirements.txt` or `renv.lock`

### LINKED ISSUE #2177: nit: When renv.lock file is not available an uninformative error is thrown
Since we now include `renv.lock` by default for all R content in the `files` array, an error is thrown that doesn't explain _why_ we're looking for a renv.lock file.  I think this would be confusing, especially to users who don't use renv regularly.  My first instinct would be to simply remove it from the config. 

> Deployment failed: open c:\Users\kevin\connect-content\bundles\deploy-err\renv.lock: The system cannot find the file specified.

I wonder if we can detect that there is no renv.lock file in the target directory and disabling the Deploy button with a clear error.

marcosnav:
## Intent

This PR focuses on 3 things:

**1. Better messaging for missing renv.lock**
Screen Shot 2024-10-11 at 13 44 46

**2. Better messaging for missing requirements.txt**
Screen Shot 2024-10-11 at 13 44 04

**3. Bit better messaging - with identifiable error code - when deployed content did not run**
_In this case it is because of empty requirements.txt_
Screen Shot 2024-10-09 at 13 27 02

The missing `renv.lock` or `requirements.txt` errors are also shown in `Posit Publisher Log`
Screen Shot 2024-10-09 at 13 14 58

Extras: Tests for `ConnectClient#CheckCapabilities`

Fixes #2177 

## Approach

Wording, error codes and tests

## Automated Tests

Yes there is

## Directions for Reviewers

The improved messages should show up when:
- [ ] Trying to deploy an R project without `renv.lock`
- [ ] Trying to deploy a Python project without `requirements.txt`
- [ ] Deploying a Python project with an **empty** `requirements.txt`

## Checklist


marcosnav:
@dotNomad @sagerb @jonkeane Notice that the behavior between an empty `renv.lock` and `requirements.txt` is different. For `requirements.txt` the deploy is done and bundle is published but content is not able to run. 

This seems ok to me given that R and Python publishing process is different - renv.lock is required at early stages to generate a manifest - but still I'd like to know if there is something you should think we should take action on.

internal/clients/connect/capabilities.go
```diff
@@ -26,11 +26,14 @@ type allSettings struct {
...
 const requirementsFileMissing = `
-can't find the package file (%s) in the project directory.
-Create the file, listing the packages your project depends on.
-Or scan your project dependencies using scan button in
-the Python Packages section of the UI and review the
-generated file`
+Missing package file %s. The file must exist and be included in the deployment.
+Create the file listing your project dependencies.
+Or do an automatic scan with the help of the Python Packages section
+of the Publisher extension and review the generated file`
```
dotNomad, line 32: I'm being nitpicky and trying to be minimal so feel free to ignore this suggestion.
- Removed "must exist" since "must be included" assumes existence 
- Added an ending `.`

```
Missing package file %s. The file must be included in the deployment.
Create the file listing your project dependencies
or do an automatic scan with the help of the Python Packages section
of the Publisher extension and review the generated file.`
```

I think we could go even more minimal since the sidebar gets _busy_ with a lot of text

<details>
  <summary>Sidebar Photo</summary>
  
  CleanShot 2024-10-10 at 14 28 20@2x

</details> 

```
Missing package file %s. The file must be included in the deployment.`
```

internal/publish/r_package_descriptions.go
```diff
@@ -19,6 +21,12 @@ type lockfileErrDetails struct {
 	Lockfile string
 }
 
+const lockfileMissing = `Missing R lockfile %s. 
+The file must exist and be included in the deployment.
+Create the file listing your project dependencies.
+Or do an automatic scan with the help of the R Packages section
+of the Publisher extension and review the generated file`
```
dotNomad, line 28: Whatever changes you end up making to the Python side based on my feedback I'd also make them here.

dotNomad (commented):
> Automated Tests
> Yes there is

I always find this hilarious 😆 

---

I believe I caught a bug with this PR. If you don't have a `requirements.txt` in your file system you get the nice error, but if you have one, but don't have it included in the Configuration you get an `exit status 255` which isn't great. It seems like it should be the same error.

The R version of this shows "Could not scan R packages from lockfile: renv.lock" which is not quite as bad, but does feel like it should have the same error.

I'll leave this review as a comment in case that isn't a bug, and should be handled as a follow-up PR.

internal/clients/connect/capabilities.go
jonkeane: I like the more minimal comment here. What about using `dependency file %s` instead of `package file %s`? From the author of the thing being published's perspective, they might have / be deploying what they think of as a package so `package` is a bit ambiguous if it's that the publisher is claiming they are missing one or their source files or what.


A few other edits:

```
Missing dependency file %s. This file must be included in the deployment.
Create the file listing your project dependencies
or do an automatic scan with the help of the Python Packages section below.
```

Even fewer words. I think the review step is something that is important (and we should include in documentation, etc.) but for how little space we have, adding recommended next steps after fixing seems a bit like a lot. I also used `below` so that we had fewer long words (I believe this is true now that all of those sections are under the webview, but correct me if I'm wrong. I also think unless we really need to we probably shouldn't reference the name of the extension like this. To me it reads as "you need to go to this other extension called Posit Publisher to do this" because it's called out like that.

```
Missing dependency file %s. This file must be included in the deployment.
```

marcosnav: I like the super minimal approach, I'll make that change 👍

marcosnav:
> I believe I caught a bug with this PR. If you don't have a requirements.txt in your file system you get the nice error, but if you have one, but don't have it included in the Configuration you get an exit status 255 which isn't great. It seems like it should be the same error.

@dotNomad Thanks for finding this! Pushed some more work addressing this 👍

dotNomad (approved):
Thanks for fixing that! Tested all the different cases with the package file and this worked great 🎉

PR #2380, opened 2024-10-10
Include posit files in new deployments

### LINKED ISSUE #2370: Include configuration (and deployment?) file by default when deploying?
We used to include this by default when we included everything, but we don't anymore. We probably should include these by default when deploying (but allow folks to _not_ send them if they really do not want to). If we include them, so long as someone pulls a bundle from connect they can start working with it (and redeploy to the same place because of that).

Are there any reasons not to?

dotNomad:
This is a rough draft of how we could include the posit files when creating new deployments.

## Intent

Resolves #2370

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

While exploring how we could do this I implemented a really simple version of this.

When creating a new deployment we first create a Configuration, passing the name we would like it to have, then create a Deployment, also with a name passed.

The configuration is passed from the inspection API which doesn't include the `.posit` files sense they aren't created at that point.

We could insert the configuration file into the files array before sending it up to be created, but the deployment still hasn't been made.

To avoid changing the APIs I did the simplest thing, included two more API calls to include the files after they are created.

We may not want to go with this method, and instead move some of this logic into the API endpoints.

Additionally this PR does not mess with the "Select Active Configuration for Deployment" feature which ideally would include the new deployment in the configuration file. It is unclear how that should behave if a Configuration by multiple Deployments.


kgartland-rstudio:
I noticed this failed on the config check in the [ui tests](https://github.com/posit-dev/publisher/actions/runs/11265087801/job/31326495488).  I have a fix locally, let me know if you want me to push it up to your branch.

dotNomad:
> I noticed this failed on the config check in the [ui tests](https://github.com/posit-dev/publisher/actions/runs/11265087801/job/31326495488). I have a fix locally, let me know if you want me to push it up to your branch.

No that is alright. This is very much a draft to prompt discussion right now. If we decide to go down this path we can update the tests.

sagerb (approved):
Reasonable approach. We can update the APIs later, but that won't be a simple exercise due to the cyclical nature of the dependencies.

kgartland-rstudio (approved):
Verified on all OSes, updated tests so they'll validate the new files in the config.

Issue #2301, opened 2024-09-23
Import: When deploying to a mismatching content type, metadata is updated but deployment fails


kgartland-rstudio:
As a test I tried to deploy a bokeh app targeting an already-deployed fastapi app on the dogfood server.

1. The Deployment Status briefly shows the deployment succeeded and then flips to failure:

https://github.com/user-attachments/assets/ad257b4a-76d2-448c-8e55-32d9e91a5818

2. The title (and other metadata included in the configuration file) for the content on Connect was updated even though the deployment failed.

I would expect this to fail earlier so we don't update the content on Connect.  Can we detect what target content type is before we even start the deployment? I think we're making the api calls on Connect to update the metadata before the call to do the deployment.  Maybe these should be in the reverse order?

STR:
* Deploy a fastapi app to Connect
* Now deploy a bokeh app using the new `update that previous deployment` link and point to the previous app URL
* I'm allowed to deploy the app
* The Deployment shows Success briefly but then displays the Failure
* The target app's metadata was updated even though the deployment failed.

Expected:
I would expect...
* that we detect a mismatching content type prior to the deployment
OR
* the deployment fails, never shows a success message and the metadata to remain unupdated

kgartland-rstudio:
Probably related:
In the standard redeployment flow, again the metadata gets updated when the deployments fail.

STR:
1. deploy the fastapi example successfully
2. update the simple.py so that it will fail
3. change the title to see if the metadata updates
4. deploy

The title updates even though the deployment fails.

sagerb:
> 2\. update the simple.py so that it will fail

@kgartland-rstudio I believe the scenario you're describing here differs from this issue. I believe that even if the redeployed content did not run successfully, the title and other metadata required to be in place before running would have been updated.

Am I following your comment correctly or perhaps missing something?

kgartland-rstudio:
> I believe that even if the redeployed content did not run successfully, the title and other metadata required to be in place before running would have been updated.

The title and other metadata from the original deploy, yes those would be up and running prior to the redeployment, but if you update the title in your config and update the app so it's now broken, the title will still get updated even though the deployment actually failed.

I'm not sure if that's a bug or not but it seems unexpected.

sagerb:
>even though the deployment actually failed.

@kgartland-rstudio I see it as two different things, but perhaps I'm not viewing it as a user would. 

[...]

PR #2381, opened 2024-10-10
Change Locked Content Error to be compatible with publisher event stream

### LINKED ISSUE #2375: Locked Content Error during deployment not represented in Publisher log view
The publisher log view summarizes the processing involved during deployments. All errors should be displayed within the log at the operation in which they were encountered.

Unfortunately, the implementation of checking for locked content does not support this.

Image

Moving this back into the previous preflight checks (currently in (bad name) internal/publish/check_configuration.go) and reporting operation status/error logs will allow the operation logging and event phases to be maintained.

sagerb:
## Intent

Resolves #2375 - Locked Content Error during deployment not represented in Publisher log view

Also, as discussed, refactors this and a few other errors to be more consistent with other preflight errors.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

The preflight checks for the content on the Connect server were not returning operation errors and were not interacting with the log used for the publisher log window (and associated event stream). This has been corrected.

To increase consistency, the preflight checks have been refactored with some renaming and movement. To summarize:
- rename publish.checkConfiguration to publish.preFlightChecks
- update connectClient.CheckCapabilities to have content target checks (ValidateDeploymentTarget) - which were moved from publish
- update unit tests

After these changes, the publisher log shows an error at the "check capabilities" stage as well as showing the error at the end of the stage summary log messages.
2024-10-09 at 10 04 PM

## Automated Tests

Agent unit tests have been updated as part of the refactoring. 

## Directions for Reviewers

I've tried to add useful comments to the file diffs. I understand that refactorings often make the worst diffs to follow... if you have any questions, please reach out.

Validate that an attempt to deploy to a locked piece of content fails in the expected error and is included in the publisher log. 
Also validate a new deployment works as expected (it will bypass the existing content prechecks).

## Checklist

### DIFF
--- a/internal/clients/connect/capabilities.go
+++ b/internal/clients/connect/capabilities.go
@@ -44,7 +44,13 @@ func checkRequirementsFile(base util.AbsolutePath, requirementsFilename string) 
 	return nil
 }
 
-func (c *ConnectClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, log logging.Logger) error {
+func (c *ConnectClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, contentID *types.ContentID, log logging.Logger) error {
+	if contentID != nil && *contentID != "" {
+		err := c.ValidateDeploymentTarget(*contentID, log)
+		if err != nil {
+			return err
+		}
+	}
 	if cfg.Python != nil {
 		err := checkRequirementsFile(base, cfg.Python.PackageFile)
 		if err != nil {
--- a/internal/clients/connect/client.go
+++ b/internal/clients/connect/client.go
@@ -30,5 +30,5 @@ type APIClient interface {
 	DeployBundle(types.ContentID, types.BundleID, logging.Logger) (types.TaskID, error)
 	WaitForTask(taskID types.TaskID, log logging.Logger) error
 	ValidateDeployment(types.ContentID, logging.Logger) error
-	CheckCapabilities(util.AbsolutePath, *config.Config, logging.Logger) error
+	CheckCapabilities(util.AbsolutePath, *config.Config, *types.ContentID, logging.Logger) error
 }
--- a/internal/clients/connect/client_connect.go
+++ b/internal/clients/connect/client_connect.go
@@ -486,3 +486,46 @@ func (c *ConnectClient) ValidateDeployment(contentID types.ContentID, log loggin
 	}
 	return err
 }
+
+// Handle preflight agent error details
+func (c *ConnectClient) preflightAgentError(agenterr *types.AgentError, contentID types.ContentID) *types.AgentError {
+	agenterr.Code = events.DeploymentFailedCode
+
+	if _, isNotFound := http_client.IsHTTPAgentErrorStatusOf(agenterr, http.StatusNotFound); isNotFound {
+		agenterr.Message = fmt.Sprintf("Cannot deploy content: ID %s - Content cannot be found", contentID)
+	} else if _, isForbidden := http_client.IsHTTPAgentErrorStatusOf(agenterr, http.StatusForbidden); isForbidden {
+		agenterr.Message = fmt.Sprintf("Cannot deploy content: ID %s - You may need to request collaborator permissions or verify the credentials in use", contentID)
+	} else {
+		agenterr.Message = fmt.Sprintf("Cannot deploy content: ID %s - Unknown error: %s", contentID, agenterr.Error())
+	}
+	return agenterr
+}
+
+func (c *ConnectClient) ValidateDeploymentTarget(contentID types.ContentID, log logging.Logger) error {
+	log.Info("Verifying existing content", "content_id", contentID)
+
+	content := &ConnectContent{}
+	err := c.ContentDetails(contentID, content, log)
+
+	// Handle possible errors from pinging to the content item
+	if aerr, isAgentErr := types.IsAgentError(err); isAgentErr {
+		return c.preflightAgentError(aerr, contentID)
+	}
+	if err != nil {
+		msg := fmt.Sprintf("Deployment target cannot be reached. Halting deployment. (Content ID = %s)", contentID)
+		return types.NewAgentError(events.DeploymentFailedCode,
+			errors.New(msg),
+			nil)
+	}
+
+	// Verify content is not locked
+	lockedErr := content.LockedError()
+	if lockedErr != nil {
+		msg := fmt.Sprintf("Content is locked, cannot deploy to it (content ID = %s)", contentID)
+		return types.NewAgentError(events.DeploymentFailedCode,
+			errors.New(msg),
+			nil)
+	}
+
+	return nil
+}
--- a/internal/clients/connect/client_connect_test.go
+++ b/internal/clients/connect/client_connect_test.go
@@ -5,6 +5,7 @@ package connect
 import (
 	"errors"
 	"io/fs"
+	"net/http"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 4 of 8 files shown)


internal/clients/connect/capabilities.go
```diff
@@ -44,7 +44,13 @@ func checkRequirementsFile(base util.AbsolutePath, requirementsFilename string)
...
 }
 
-func (c *ConnectClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, log logging.Logger) error {
+func (c *ConnectClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, contentID *types.ContentID, log logging.Logger) error {
+	if contentID != nil && *contentID != "" {
+		err := c.ValidateDeploymentTarget(*contentID, log)
+		if err != nil {
+			return err
+		}
+	}
```
sagerb, line 53: If we are updating an existing deployment, we'll run those validation tests (now located within the ConnectClient package)

internal/clients/connect/client.go
```diff
@@ -30,5 +30,5 @@ type APIClient interface {
 	DeployBundle(types.ContentID, types.BundleID, logging.Logger) (types.TaskID, error)
 	WaitForTask(taskID types.TaskID, log logging.Logger) error
 	ValidateDeployment(types.ContentID, logging.Logger) error
-	CheckCapabilities(util.AbsolutePath, *config.Config, logging.Logger) error
+	CheckCapabilities(util.AbsolutePath, *config.Config, *types.ContentID, logging.Logger) error
```
sagerb, line 33: If you didn't notice it in the previous file, I've had to add a pointer to an optional contentID into this interface.

internal/clients/connect/client_connect.go
```diff
@@ -486,3 +486,46 @@ func (c *ConnectClient) ValidateDeployment(contentID types.ContentID, log loggin
 	}
 	return err
 }
+
+// Handle preflight agent error details
```
sagerb, line 490: the `preflightAgentError` code is a port from the previous implementation within the publish.go file.

internal/clients/connect/client_connect.go
```diff
@@ -486,3 +486,46 @@ func (c *ConnectClient) ValidateDeployment(contentID types.ContentID, log loggin
...
+	if _, isNotFound := http_client.IsHTTPAgentErrorStatusOf(agenterr, http.StatusNotFound); isNotFound {
+		agenterr.Message = fmt.Sprintf("Cannot deploy content: ID %s - Content cannot be found", contentID)
+	} else if _, isForbidden := http_client.IsHTTPAgentErrorStatusOf(agenterr, http.StatusForbidden); isForbidden {
+		agenterr.Message = fmt.Sprintf("Cannot deploy content: ID %s - You may need to request collaborator permissions or verify the credentials in use", contentID)
+	} else {
+		agenterr.Message = fmt.Sprintf("Cannot deploy content: ID %s - Unknown error: %s", contentID, agenterr.Error())
+	}
+	return agenterr
+}
+
```
sagerb, line 503: The new method below, while renamed, is pretty much still the same implementation that was moved from the publish.go file.

internal/clients/connect/client_connect_test.go
```diff
@@ -590,3 +591,131 @@ func (s *ConnectClientSuite) TestContentDetails() {
 	s.ErrorIs(err, expectedErr)
 	httpClient.AssertExpectations(s.T())
 }
+
```
sagerb, line 594: Unit tests for all of the validate errors, below.

internal/clients/connect/client_connect_test.go
```diff
@@ -590,3 +591,131 @@ func (s *ConnectClientSuite) TestContentDetails() {
...
+		Title:              "",
+		GUID:               "",
+		Description:        "",
+		AccessType:         "",
+		ServiceAccountName: "",
+		DefaultImageName:   "",
+		Locked:             true,
+	}
+	httpClient := &http_client.MockHTTPClient{}
+
```
sagerb, line 698: below, I was able to use the mock library's ability to run a function to implement the actual update into the ConnectContent struct. This allowed us to properly unit test the locked content test.

internal/clients/connect/mock_client.go
```diff
@@ -82,7 +82,12 @@ func (m *MockClient) ValidateDeployment(id types.ContentID, log logging.Logger)
 	return args.Error(0)
 }
 
-func (m *MockClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, log logging.Logger) error {
-	args := m.Called(base, cfg, log)
+func (m *MockClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, contentID *types.ContentID, log logging.Logger) error {
```
sagerb, line 85: had to mimic the interface change with the additional param of a ContentID.

internal/clients/connect/mock_client.go
```diff
@@ -82,7 +82,12 @@ func (m *MockClient) ValidateDeployment(id types.ContentID, log logging.Logger)
...
+	args := m.Called(base, cfg, contentID, log)
+	return args.Error(0)
+}
+
```
sagerb, line 89: new mock implemented as it was added to the interface we're testing.

internal/publish/preflight_checks.go
```diff
@@ -17,7 +16,7 @@ type checkConfigurationSuccessData struct{}
 
 var errTypeChanged = errors.New("configuration type cannot be changed once deployed; create a new destination to use a different type")
 
-func (p *defaultPublisher) checkConfiguration(client connect.APIClient) error {
+func (p *defaultPublisher) preFlightChecks(client connect.APIClient) error {
```
sagerb, line 19: the functionality started out just checking the configuration, but we had always referred to it as the preflight checks. This renaming should help remove some of the confusion we've recently had.

internal/publish/preflight_checks.go
```diff
@@ -30,17 +29,12 @@ func (p *defaultPublisher) checkConfiguration(client connect.APIClient) error {
 	}
 	log.Info("Publishing with credentials", "username", user.Username, "email", user.Email)
 
+	var existingContentID *types.ContentID
 	if p.Target != nil {
```
sagerb, line 33: Removed the code below as it was simply checking if the last content type deployed (from the content record) is the same as the "new" value within the configuration file. Since we really want to be checking against the Connect server's state for this deployment, this is being removed. It will be replaced in a PR coming up for issue #2301.

internal/publish/publish.go
```diff
@@ -267,49 +265,6 @@ func (p *defaultPublisher) createDeploymentRecord(
 	return p.writeDeploymentRecord()
 }
```
sagerb, line 267: Below you'll see the original code which was moved into `client_connect`

internal/publish/publish.go
```diff
@@ -267,49 +265,6 @@ func (p *defaultPublisher) createDeploymentRecord(
...
-		p.log.Debug("Insufficient permissions. Content deployment cannot proceed", "content_id", contentID)
-		aerr.Message = fmt.Sprintf("Cannot deploy content: ID %s - You may need to request collaborator permissions or verify the credentials in use", contentID)
-	} else {
-		p.log.Debug("Unknown error while deploying", "content_id", contentID)
-		aerr.Message = fmt.Sprintf("Cannot deploy content: ID %s - Unknown error: %s", contentID, agenterr.Error())
-	}
-
-	return agenterr
-}
-
```
sagerb, line 287: Below you'll see the original code which was moved into client_connect: preflightAgentError method.

internal/publish/publish.go
```diff
@@ -329,7 +284,7 @@ func (p *defaultPublisher) publishWithClient(
 		return err
 	}
 
-	err = p.checkConfiguration(client)
+	err = p.preFlightChecks(client)
```
sagerb, line 287: As discussed, this was renamed to reduce confusion and increase maintainability.

internal/publish/publish_test.go
```diff
@@ -198,54 +195,21 @@ func (s *PublishSuite) TestPublishWithClientRedeployErrors() {
 	target := deployment.New()
 	target.ID = "myContentID"
 
+	// Fail the preflight existing content checks. Further testing of these will
+	// occur in the connect_client package
+
```
sagerb, line 200: The thought with these changes is that the real testing has moved into the `client` implementation (`client_connect`). What remains here is enough testing to know if the capabilities check fails with an error, then this method `publishWithClient` will fail with an error.

internal/publish/publish_test.go
```diff
@@ -309,7 +273,7 @@ func (s *PublishSuite) publishWithClient(
 		client.On("CreateDeployment", mock.Anything, mock.Anything).Return(myContentID, errsMock.createErr)
 	}
 	client.On("TestAuthentication", mock.Anything).Return(&connect.User{}, errsMock.authErr)
-	client.On("CheckCapabilities", mock.Anything, mock.Anything, mock.Anything).Return(errsMock.capErr)
+	client.On("CheckCapabilities", mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(errsMock.capErr)
```
sagerb, line 276: impact of new parameter...

internal/publish/preflight_checks.go
```diff
@@ -30,17 +29,12 @@ func (p *defaultPublisher) checkConfiguration(client connect.APIClient) error {
...
-
-		if previousType != "" && previousType != config.ContentTypeUnknown && currentType != previousType {
-			log.Debug("Content type mismatch between previous and new target", "previous_type", previousType, "current_type", currentType)
-			return types.OperationError(op, errTypeChanged)
-		}
+		existingContentID = &p.Target.ID
 	}
-	err = client.CheckCapabilities(p.Dir, p.Config, log)
+
+	err = client.CheckCapabilities(p.Dir, p.Config, existingContentID, log)
```
marcosnav, line 37: Having to use a pointer for the ContentID feels a bit off, WDYT about passing the `Target` to the `CheckCapabilities`? then, you would only need one check, the one you have here above moved over to `ChackCapabilities`.
e.g:
```go
err = client.CheckCapabilities(p.Dir, p.Config, p.Target, log)

...
// CheckCapabilities
func (c *ConnectClient) CheckCapabilities(base util.AbsolutePath, cfg *config.Config, target *deployment.Deployment, log logging.Logger) error {
    if target != nil {
        err := c.ValidateDeploymentTarget(target.ID, log)
        if err != nil {
            return err
        }
}
```

marcosnav (approved):
LGTM, nice approach on organizing these checks 💯 

Just one small suggestion

internal/publish/preflight_checks.go
sagerb: @marcosnav It's a decent idea, but unfortunately, the inclusion of a `deployment` import brings us to a circular dependency that doesn't look quick to unwind. OK if we leave this for later?

>package github.com/posit-dev/publisher/cmd/publisher
        imports github.com/posit-dev/publisher/cmd/publisher/commands
        imports github.com/posit-dev/publisher/internal/clients/connect
        imports github.com/posit-dev/publisher/internal/deployment
        imports github.com/posit-dev/publisher/internal/inspect/dependencies/renv
        imports github.com/posit-dev/publisher/internal/bundles
        imports github.com/posit-dev/publisher/internal/clients/connect: import cycle not allowed

marcosnav: Ah, lovely circular dependencies 😬 !

Yes, I'm okay with leaving this as is, hitting circular dependencies issues is food for thought, though. Very likely we are sharing similar responsibilities between different instances, and/or not using interfaces enough. Worth revisiting at some point.

sagerb: Totally agree, it is a very big warning sign. We'll revisit this no doubt and get it straightened out.

PR #2392, opened 2024-10-10
Add content type preflight check for already deployed content

### LINKED ISSUE #2301: Import: When deploying to a mismatching content type, metadata is updated but deployment fails

sagerb:
## Intent

Resolves #2301

This PR adds a appMode pre-flight check, so a content item selected by mistake with the wrong appMode doesn't get updated.

## Type of Change

- - [x] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Additional pre-flight check added. It now shows in the Publisher log and the error displays consistently.

2024-10-10 at 1 03 PM

## Automated Tests

Unit tests updated.

## Directions for Reviewers

To reproduce the issue, (and not have to deal with selectively updating your config manually), you should create a new deployment and then associate it with a different appMode type of content already deployed.

Please also verify that normal re-deployment works without an issue.

## Checklist

### DIFF
--- a/extensions/vscode/src/eventErrors.ts
+++ b/extensions/vscode/src/eventErrors.ts
@@ -48,6 +48,5 @@ export const handleEventCodedError = (
     return emsg.data.message;
   }
 
-  const unknownErrMsg = emsg.data.error || emsg.data.message;
-  return `${unknownErrMsg}`;
+  return emsg.data.error || emsg.data.message;
 };
--- a/extensions/vscode/src/eventErrors.ts
+++ b/extensions/vscode/src/eventErrors.ts
@@ -49,5 +49,5 @@ export const handleEventCodedError = (
   }
 
   const unknownErrMsg = emsg.data.error || emsg.data.message;
-  return `Unknown error: ${unknownErrMsg}`;
+  return `${unknownErrMsg}`;
 };
--- a/internal/clients/connect/client_connect.go
+++ b/internal/clients/connect/client_connect.go
@@ -530,10 +530,10 @@ func (c *ConnectClient) ValidateDeploymentTarget(contentID types.ContentID, cfg 
 
 	// Verify content type has not changed
 	log.Info("Verifying content type is the same")
-	existingAppMode := content.AppMode
-	configAppMode := AppModeFromType(cfg.Type)
-	if configAppMode != existingAppMode {
-		msg := fmt.Sprintf("Content was previously deployed as '%s' but your configuration is set to '%s'. A new deployment must be needed.", existingAppMode.Description(), configAppMode.Description())
+	existingType := ContentTypeFromAppMode(content.AppMode)
+	configType := cfg.Type
+	if existingType != configType {
+		msg := fmt.Sprintf("Content was previously deployed as '%s' but your configuration is set to '%s'.", existingType, configType)
 		return types.NewAgentError(events.AppModeNotModifiableCode, errors.New(msg), nil)
 	}
 
--- a/internal/clients/connect/client_connect.go
+++ b/internal/clients/connect/client_connect.go
@@ -501,7 +501,7 @@ func (c *ConnectClient) preflightAgentError(agenterr *types.AgentError, contentI
 	return agenterr
 }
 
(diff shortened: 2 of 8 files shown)


internal/clients/connect/client_connect.go
```diff
@@ -527,5 +528,14 @@ func (c *ConnectClient) ValidateDeploymentTarget(contentID types.ContentID, log
 			nil)
 	}
 
+	// Verify content type has not changed
+	log.Info("Verifying content type is the same")
+	existingAppMode := content.AppMode
+	configAppMode := AppModeFromType(cfg.Type)
+	if configAppMode != existingAppMode {
+		msg := fmt.Sprintf("Content was previously deployed as '%s' but your configuration is set to '%s'. A new deployment must be needed.", existingAppMode.Description(), configAppMode.Description())
```
dotNomad, line 536: I'm a bit worried about using the description version of these rather than pointing to the type verbatim. I think having the type will help users identify exactly what is going wrong.

> Content was previously deployed as a FastAPI application with `type='python-fastapi'`, but the configuration is set to `type='python-flask'`

I would drop "A new deployment must be needed." and leave it to the user to decide to make a correction or make a  new deployment.

dotNomad (approved):
Logic looks great, I just had a thought on the error message itself.

extensions/vscode/src/eventErrors.ts
```diff
@@ -49,5 +49,5 @@ export const handleEventCodedError = (
   }
 
   const unknownErrMsg = emsg.data.error || emsg.data.message;
-  return `Unknown error: ${unknownErrMsg}`;
+  return `${unknownErrMsg}`;
```
marcosnav, line 52: Without the introductory text, now this can be `return emsg.data.error || emsg.data.message;`

marcosnav (approved):
LGTM

Just one comment added previously to simplify returning the unknown error event messages

internal/clients/connect/client_connect.go
sagerb: I considered this. I found that we have different values for config.types vs. AppModes, so it became a question of which one should you consider. I leaned towards showing the config.types (knowing it is different than the AppMode descriptions or values), but thought this might be a better compromise. Any thoughts?

extensions/vscode/src/eventErrors.ts
sagerb: Great observation. I'll make the change.

internal/clients/connect/client_connect.go
sagerb: After thinking further on this, I'm just going to go ahead and make the change you suggested and we can always change it in the future.

sagerb:
Updated screenshot after PR recommendations:

2024-10-10 at 4 08 PM

kgartland-rstudio:
When this fails, I would expect there to be an option to try to try again with a different URL.

Scenario:
Grab a URL to a piece of content from the server that you want to target
Accidentally chose the wrong URL
Got the proper failure 🎉 
Now I get the right URL and want to deploy to that content, but I can't now 😠 

Screenshot 2024-10-16 at 12 46 54 PM

PR #2393, opened 2024-10-10
Messaging details for deployed content that runs with error

### LINKED ISSUE #2206: Content failed to start message could be better

marcosnav:
## Intent

This PR applies some presentational details for content that for some reason fails to start, this is only for content that has runtime.

- Showing a warning sign now, instead of an "x" icon that mistakenly suggested to be clickable
- The button reads as "View Content Logs" and directs to Connect content logs page.

Screen Shot 2024-10-11 at 14 34 30

Fixes #2206 

## Approach

Just some presentational changes and checks for errored deployments.

## Automated Tests

Added some unit tests around the Agent Errors type guards

## Directions for Reviewers

- [ ] Deploy an R or Python content item that has some code with errors in it - forcing it to fail at runtime. The message should be better now.

## Checklist

### DIFF
--- a/extensions/vscode/src/api/types/error.ts
+++ b/extensions/vscode/src/api/types/error.ts
@@ -1,12 +1,17 @@
 // Copyright (C) 2023 by Posit Software, PBC.
 
+import { ErrorCode } from "../../utils/errorTypes";
+
 type AgentErrorBase = {
-  code: string;
+  code: ErrorCode;
   msg: string;
   operation: string;
 };
 
-export type AgentError = AgentErrorTypeUnknown | AgentErrorInvalidTOML;
+export type AgentError =
+  | AgentErrorTypeUnknown
+  | AgentErrorInvalidTOML
+  | AgentErrorContentNotRunning;
 
 export type AgentErrorTypeUnknown = AgentErrorBase & {
   data: {
@@ -19,7 +24,9 @@ export type AgentErrorTypeUnknown = AgentErrorBase & {
 export const isAgentErrorTypeUnknown = (
   e: AgentError,
 ): e is AgentErrorTypeUnknown => {
-  return !isAgentErrorInvalidTOML(e);
+  return (
+    !isAgentErrorInvalidTOML(e) && !isAgentErrorDeployedContentNotRunning(e)
+  );
 };
 
 export type AgentErrorInvalidTOML = AgentErrorBase & {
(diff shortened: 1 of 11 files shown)


extensions/vscode/src/api/types/error.ts
```diff
@@ -1,12 +1,17 @@
 // Copyright (C) 2023 by Posit Software, PBC.
 
+import { ErrorCode } from "../../utils/errorTypes";
+
 type AgentErrorBase = {
-  code: string;
+  code: ErrorCode;
```
dotNomad, line 6: Can we move this type under the API module since it is errors that are coming from our API? Feel free to ignore, just a cleanup suggestion.

dotNomad (approved):
Minor nit about text styling, otherwise looks great 😎 

> deployed content does not seem to be running. See the logs in Connect for details

Some errors we capitalize the beginning, and others we don't. Same goes for the period ending.

I'd push for sentence case and period endings to be consistent.

sagerb:
Sorry that I didn't provide this input sooner but I regret that we've lost the View Content button. That button is still applicable in the case where the content didn't run, in that you can sometimes fine error information displayed when viewing that content.

An alternative, which I'd like to propose, would be to have your error message include an anchor that would link to showing the logs, and we keep the View Content button as it was. This requires some string parsing/replacement, but we do have an example in our implementation for editing a config file. Check out the usage of the `TextStringWithAnchor` component within the `EvenEasierDeploy` component.

extensions/vscode/src/api/types/error.ts
marcosnav: Agreed 100% with you, there are some more changes to do on error types and at some point we will consolidate these, hopefully in the near term

marcosnav:
@dotNomad Thanks for the punctuation observation. In my last commit, I went ahead and created a small normalization method to prevent this from happening when the message is derived from an `error` instance in the Go code. Since linting rules for go discourage creating errors with sentence case and period endings, this for error chaining practices.

dotNomad:
> Since linting rules for go discourage creating errors with sentence case and period endings, this for error chaining practices.

Fantastic. Thanks for adding that!

PR #2394, opened 2024-10-11
Display a user-friendly message when deploying, but config has a TOML…

### LINKED ISSUE #2374: Unfriendly error when deploying with invalid TOML
Since we can now deploy content with an invalid configuration file, unfriendly errors are thrown.

Example:
I have an error in my configuration file and the proper warning is displayed in the Home View.

Image

However when I try to deploy with that invalid configuration file I get this error:

> Failed to deploy . An error has occurred at homeView, deploy, Status=400, StatusText=Bad Request, Code=ERR_BAD_REQUEST, Msg=Bad Request: /Users/kgartland/work/publishing-client/test/sample-content/fastapi-simple/.posit/publish/fastapi-simple-0ARR.toml:18:7: expected = after a key, but the document ends there , URL=http://localhost:53562/api/post/deployments/deployment-NU2B

Can we throw a cleaner error here?

sagerb:
… errror

## Intent

Resolves #2374 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Changed agent to return a JSON error which then was detected at the UX and was able to build an error string similar to what we show in our configuration warning.

There was a circular dependency that I had to unwind, which moved internal/services/api/api_errors.go to types}/api_errors.go. This affected the majority of the files within this PR. The important files relevant to the functionality is in `internal/services/api/post_deployment.go`

Screenshot of error
2024-10-11 at 11 44 AM

## Automated Tests

Unit tests updated.

## Directions for Reviewers

A few questions for the reviewers:
- I am planning on updating the error message to contain a button to edit the configuration. Do you think that needs to be included in this blocking PR or can it come later?
- You should notice this failure does not update the content-record for the last deployment. For this type of error, which we were previously blocking the deployment for, do you think it should?

## Checklist


internal/types/api_errors.go
```diff
@@ -1,13 +1,18 @@
...
 // Copyright (C) 2024 by Posit Software, PBC.
 
 import (
+	"encoding/json"
 	"net/http"
-
-	"github.com/posit-dev/publisher/internal/types"
 )
 
+func jsonResult(w http.ResponseWriter, status int, result any) {
```
dotNomad, line 10: Did we have to remake this from `api_helpers` due to the circular dependency?

extensions/vscode/src/views/homeView.ts
```diff
@@ -268,7 +268,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       deployProject(response.data.localId, this.stream);
     } catch (error: unknown) {
       const summary = getSummaryStringFromError("homeView, deploy", error);
-      window.showInformationMessage(`Failed to deploy . ${summary}`);
+      window.showErrorMessage(`Failed to deploy . ${summary}`);
```
dotNomad, line 271: Not sure how this extra space snuck in, but lets fix it here.
```suggestion
      window.showErrorMessage(`Failed to deploy. ${summary}`);
```

dotNomad (approved):
This worked great for me!

> I am planning on updating the error message to contain a button to edit the configuration. Do you think that needs to be included in this blocking PR or can it come later?

I think this can come later as a follow-up since we already have the "Edit the Configuration" link in the sidebar.

> You should notice this failure does not update the content-record for the last deployment. For this type of error, which we were previously blocking the deployment for, do you think it should?

Is this the only time we don't update the content record? I don't think we really can since the Configuration isn't valid, but was curious.

marcosnav (approved):
LGTM.

Same observations as Jordan. I think this work is excellent for 1.2. The button to edit configuration and also updating the content record for TOML errors are not required or should not block the release

internal/types/api_errors.go
sagerb: yes, that is correct. I didn't see a good way of moving the original JsonResult to somewhere else. Any suggestions?

sagerb:
The latest commit addressed a few other ways a TOML parsing error was reported through the agent.
(To handle additional locations where TOML may fail, I added new errors tomlValidationError and tomlUnknownError.) Some errors will be shown with the line number, others will not (when not available). These will be shown when you click deploy.

List of invalid TOML syntaxes used for testing:
- `Title` (line number shown)
- `Title =` (line number shown)
- `Title = 123` (line number shown)
- `[Title]` (no line number)
- `title = ['abc']` (no line number)
- `abc` (line number shown)
- `title = 'abc'
title = 123` (two lines... no line number)
- `type = 'python-fastapi22'` (no line number)

kgartland-rstudio:
Looks good.  Tested the suggested scenarios on MacOS, Windows and Linux AMD+ARM.

kgartland-rstudio (approved):

PR #2401, opened 2024-10-17
Add credential server URL discovery

### LINKED ISSUE #2001: Credentials: URL Validation
Adding `https://rsc.radixu.com/adsdfafasdf` passes URL validation when it shouldn't.

Adding `https://rsc.radixu.com/connect` passes URL validation but fails API Key validation with...
Screenshot 2024-07-19 at 11 47 12 AM

 If we're supporting the `/connect` path, this should pass.  If we aren't supporting the `/connect` path, I would expect this to fail at the URL validation step with an explanation that `/connect` shouldn't be included in the server url.

### LINKED ISSUE #2371: "Enter the Public URL of the Posit Connect Server" is a little too wordy
One location: https://github.com/posit-dev/publisher/blob/7f8677442ea18c17863808d1ffcd45abc53da847/extensions/vscode/src/multiStepInputs/newCredential.ts#L87-L88

Consider changing to "Enter the URL of the Posit Connect server" not sure that 'public' is helping here. I also wonder if this whole sequence should have "Enter the..." (do other steppers say the same thing? or do they just start with the noun phrase?)

sagerb:
## Intent

When prompting for a server URL, discover the actual server API address needed by iteratively removing path segments and retesting API key authentication.

PR also includes word change for server URL prompts.

Resolves #2001
Resolves #2371 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Given the URL string for the server (provided by the publisher), attempt to validate the combination of server address and API key by removing path segments from the end of the server URL, until we either are able to make a successful API key validated request or we run out of possibilities.

A URL utility method was created to return the possibilities. It strips off any query params and URL fragments before producing an array of server URLs to try. 

The multi-steppers that create credentials (new credential and new deployment) have been updated to use the validated URL returned from the test credentials API, (instead of the URL provided by the user), when calling the credential creation API.

## Automated Tests

Unit tests have been added and updated to cover the functionality.

## Directions for Reviewers

There are two multi-steppers impacted by this change - creating a new deployment with a new credential as well as creating a new credential (from the credential view).

Validation needs to test the entry of a server URL with
- invalid or unreachable URL to a Connect server. 
- URL to another type of server, should error
- valid server URL, but the URL contains additional path segments and query params (copy/paste from the dashboard URL works nicely)
- valid server URL as previously required.

You can save yourself some time by going through the multi-steppers, entering the URL to be tested, then entering the API key, and if it moves to the naming credential step, you can then back up and you should see the updated/discovered value shown in the server URL step. This only occurs after the API key validation occurs (if you try backing up just after entering the URL at the API key entry, you will see the original entered URL).

Be sure to validate that you can deploy to the servers after creating credentials with the exact URL or one which needs to be automatically adjusted through the new discovery process.

## Checklist

### DIFF
--- a/extensions/vscode/src/utils/errors.ts
+++ b/extensions/vscode/src/utils/errors.ts
@@ -34,14 +34,18 @@ export const getCodeStringFromError = (error: unknown): string | undefined => {
 };
 
 export const getMessageFromError = (error: unknown): string => {
-  if (axios.isAxiosError(error)) {
-    return error.response?.data || error.message;
-  }
-  if (isAgentError(error)) {
-    return error.msg;
-  }
-  if (error instanceof Error) {
-    return error.message;
+  try {
+    if (axios.isAxiosError(error)) {
+      return error.response?.data || error.message;
+    }
+    if (isAgentError(error)) {
+      return error.msg;
+    }
+    if (error instanceof Error) {
+      return error.message;
+    }
+  } catch {
+    // errors suppressed
   }
   return String(error);
 };
@@ -57,8 +61,9 @@ export const getAPIURLFromError = (error: unknown) => {
   return undefined;
 };
 
-// When the error is a known JSON agent error it returns it's message.
-// Otherwise, a tracing message is returned to help diagnose.
+// This method builds a diagnostic message which is output to the
+// VSCode console (output/window) to help diagnose, but then returns the
+// base error string from the error.
 export const getSummaryStringFromError = (location: string, error: unknown) => {
   let msg = `An error has occurred at ${location}`;
 
@@ -97,7 +102,8 @@ export const getSummaryStringFromError = (location: string, error: unknown) => {
   } else {
     msg += `, Error=${error}`;
   }
-  return msg;
+  console.error(msg);
+  return getMessageFromError(error);
 };
 
 export const getSummaryFromError = (error: unknown) => {
--- a/extensions/vscode/src/utils/errors.ts
+++ b/extensions/vscode/src/utils/errors.ts
@@ -2,6 +2,7 @@
 
 import axios from "axios";
 import { isAxiosErrorWithJson, resolveAgentJsonErrorMsg } from "./errorTypes";
+import { isAgentError } from "src/api/types/error";
 
 export type ErrorMessage = string[];
 export type ErrorMessages = ErrorMessage[];
@@ -26,12 +27,18 @@ export const getCodeStringFromError = (error: unknown): string | undefined => {
   if (axios.isAxiosError(error)) {
     return error.code;
   }
+  if (isAgentError(error)) {
+    return error.code;
+  }
   return undefined;
 };
 
 export const getMessageFromError = (error: unknown): string => {
   if (axios.isAxiosError(error)) {
     return error.response?.data || error.message;
+  }
+  if (isAgentError(error)) {
+    return error.msg;
   }
   if (error instanceof Error) {
     return error.message;
@@ -57,6 +64,17 @@ export const getSummaryStringFromError = (location: string, error: unknown) => {
 
   if (isAxiosErrorWithJson(error)) {
     return resolveAgentJsonErrorMsg(error);
+  }
+
+  if (isAgentError(error)) {
+    msg = error.msg;
+    if (error.code) {
+      msg += `, Code=${error.code}`;
+    }
+    if (error.operation) {
+      msg += `, Operation=${error.operation}`;
+    }
+    return msg;
   }
 
   const summary = getSummaryFromError(error);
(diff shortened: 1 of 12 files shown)


internal/services/api/post_test_credentials.go
```diff
@@ -37,26 +39,57 @@ func PostTestCredentialsHandlerFunc(log logging.Logger) http.HandlerFunc {
...
+				AuthType:   accounts.AuthTypeAPIKey,
+				URL:        urlToBeTested,
+				Insecure:   b.Insecure,
+				ApiKey:     b.ApiKey,
+			}
+
+			timeout := time.Duration(max(b.Timeout, 30) * 1e9)
+			client, err := clientFactory(acct, timeout, nil, log)
+			if err != nil {
+				InternalError(w, req, log, err)
```
marcosnav, line 69: I know this is the previous `InternalError` usage moved into here, but it'd be nice to start shying away on returning 500 errors when possible

extensions/vscode/src/utils/errors.ts
```diff
@@ -59,6 +66,17 @@ export const getSummaryStringFromError = (location: string, error: unknown) => {
...
+  if (isAgentError(error)) {
+    msg = error.msg;
+    if (error.code) {
+      msg += `, Code=${error.code}`;
+    }
+    if (error.operation) {
+      msg += `, Operation=${error.operation}`;
+    }
+    return msg;
+  }
```
marcosnav, line 78: For my understanding, the changes introduced to `AgentError` in TS files are meant for debugging purposes and do not represent a change in functionality, am I right?

Just to confirm we are not adding up on having cryptic messages showing up to users

internal/services/api/post_test_credentials.go
```diff
@@ -37,26 +39,57 @@ func PostTestCredentialsHandlerFunc(log logging.Logger) http.HandlerFunc {
...
+			if lastTestError == nil {
+				// If we succeeded, pass back what URL succeeded
+				response := &PostTestCredentialsResponseBody{
+					User:  user,
+					Error: nil,
+					URL:   urlToBeTested,
+				}
+				w.Header().Set("content-type", "application/json")
+				w.WriteHeader(http.StatusOK)
+				json.NewEncoder(w).Encode(response)
```
marcosnav, line 83: It should be possible to use `JsonResult` here

kgartland-rstudio:
This is great, it fixes issues with server discovery that each deployment tool has had. Great idea to iteratively strip the URL back.

Found one issue though:
If using a URL that matches an existing host but includes a path, we don't catch the conflict.

Example:
* Create a set of creds for http://localhost:3939
* Now create a new set of creds for http://localhost:3939/connect/
* Instead of throwing an error in the multi-stepper, I'm allowed to proceed to the API Key field.
* After adding the API Key I hit an uncaught 409 error in the notification pop-up in the lower right screen:
Screenshot 2024-10-17 at 9 25 43 AM


Expected:
Catch the conflict and throw the error after adding the server like we do when not including paths:
Screenshot 2024-10-17 at 9 00 45 AM

sagerb:
>Found one issue though:

@kgartland-rstudio I pushed up a fix for the issue you found. Can you pull it down and confirm that I've addressed it? Thanks!

kgartland-rstudio:
> fixes issues with serve

Verified fix!

kgartland-rstudio (approved):
Validated new capability with [Dogfood](https://rsc.radixu.com/connect/#/content/listing), [Palm Server](https://pub.palm.ptd.posit.it/connect/#/content/listing?q=is:published), and Yeti to ensure we cover various proxies and paths.

internal/services/api/post_test_credentials.go
sagerb: agreed, but let's address it in a separate PR as we eliminate the others. I'd like to keep this release-blocking PR as targeted as possible.

extensions/vscode/src/utils/errors.ts
sagerb: We've had some additional changes within this area, to remove the cryptic side of the unexpected error reporting which relies upon these updates. The additional functionality you saw above was to improve pulling out any error message we might have within the AgentError responses. This is certainly an area we'll be coming back to, but for now, for this release and PR, we decided to remove the diagnostic information, so that the error messages are brief and further diagnostic info is now directed to the VSCode output pane for Window (by using console.error()).

internal/services/api/post_test_credentials.go
sagerb: Sorry, but I am not understanding why we would want to do that. This is a normal 200 response for this API call and my updates simply added another data field to the response.

Issue #2365, opened 2024-10-08
Add support for retrieving environment variables from Connect


sagerb:
This will include extending `connect_client` within the agent and supporting a new API to be accessed from UX.

Retrieve the list of environment variable names defined for this content item w/
> GET /v1/content/{guid}/environment

source: [API documentation](https://docs.posit.co/connect/api/#get-/v1/content/-guid-/environment)

will return a list of environment variable names, without values.
```
[
"DATABASE_URL",
"DEBUG"
]
```

dotNomad:
Idea for the wording on the warning above the deploy button:

> A Secret is required for this deployment, but does not have a value set.
> SECRET_NAME

dotNomad:
Some additional suggestions are to use `cloud-upload` for secrets that have values set to send up, and show the `***` value if the secret has a value (on the server or set to send up)

![Image](https://github.com/user-attachments/assets/f75853ff-d523-465a-b040-bfa26234d466)

dotNomad:
Additional improvement from the thread. Using `check` for Secrets already satisfied on the server.

![Image](https://github.com/user-attachments/assets/b6526bea-a377-481b-a341-e101bb0e402d)

PR #2412, opened 2024-10-28
Add deployment environment API endpoint

dotNomad:
This PR adds a new endpoint for our binary `GET /api/deployments/$NAME/environment` which will either:
- Fail in a myriad of different ways
- Give an array of strings with the names of the environment variables on Connect (see [the API Docs](https://docs.posit.co/connect/api/#get-/v1/content/-guid-/environment))

I kept all of the errors bubbling up since the frontend can handle this by ignoring them for the most part and leaving better Connect Content error handling to other, more pointed features.

The plan is to capture these errors and assume that the environment is `[]` if something goes awry. 

## Intent

Part of #2365 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach was to fetch everything that is necessary to call the Connect API and fail on the different cases that can happen before calling the API.

Then if we call the API pass through any errors.

## Automated Tests

There are automated tests for the new endpoint.

## Directions for Reviewers

The API can be tested directly using `just run`. There are tests for the various error states:
- If the deployment record with the given name doesn't exist -> `404`
- If the deployment record exists, but is in error and cannot be decoded -> `400`
- If the deployment record exists, but hasn't been deployed -> `400`
- If the deployment record exists, but there is no credential for the given Server URL -> `400`
- Pass through errors from the Connect API
  - For example if authentication fails, or the content has been deleted
- A few `InternalErrors` for unexpected cases


internal/services/api/api_service.go
```diff
@@ -166,6 +166,10 @@ func RouterHandlerFunc(base util.AbsolutePath, lister accounts.AccountList, log
 	r.Handle(ToPath("deployments", "{name}"), DeleteDeploymentHandlerFunc(base, log)).
 		Methods(http.MethodDelete)
 
+	// GET /api/deployments/$NAME/environment
+	r.Handle(ToPath("deployments", "{name}", "environment"), GetDeploymentEnvironmentHandlerFunc(base, log, lister)).
```
dotNomad, line 170: Unsure about the name here. Maybe we should be more direct about this being the Connect Content environment?

This is on the `deployment` resource which implies this is the content, and not the configuration's environment.

internal/services/api/get_deployment_env.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			// If the deployment file doesn't exist, return a 404
+			if errors.Is(err, fs.ErrNotExist) {
+				http.NotFound(w, req)
+				return
+			}
+			// If the deployment file is in error return a 400
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte(fmt.Sprintf("deployment %s is invalid: %s", name, err)))
+			return
```
marcosnav, line 41: Many of these error responses can be sent as specific agent errors, like in `post_deployment.go`. I'm guessing those changes are scoped to a future PR, when bridging the frontend, but wanted to confirm.

internal/services/api/api_service.go
marcosnav: I think it makes sense as you have it 👍

marcosnav (approved):
LGTM

PR #2414, opened 2024-10-31
Improve Secrets view by reflecting server Environment Variable states

### LINKED ISSUE #2365: Add support for retrieving environment variables from Connect

dotNomad:
This PR introduces a new feature for the Secrets view showing more detail based on the state of the content on Connect using [the Get environment variables API endpoint](https://docs.posit.co/connect/api/#get-/v1/content/-guid-/environment).

image

Now the Secret will be de-emphasized if it already has a value is on the server, and emphasized when an action needs to be taken or is currently being taken (for example setting a value) by the user.

Additional the codicons are improved showing:
- a warning when the Configuration has a Secret declared that is not on the server yet
- showing a cloud upload icon when a value is ready to be sent up
- showing a check mark when the server already the Environment Variable set

The tooltips are also improved for the 4 different states the Secrets can have.

## Intent

Resolves #2365

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach is similar to our other features, this one being a bit more unique since it potentially calls the Connect API.

When the Configuration changes or the Deployment changes we make a request to the binaries API at `GET api/deployments/$NAME/environment` which handles whether it should error or if it has all of the details to call Connect.

If Connect is called we either silently error if the API errors (for example from a missing piece of Content from deletion or auth errors) or we send the web view the array of Environment Variable names.

From there the names are put into the webview Pinia store to compare with the secrets the user has locally set or unset.

One thing to note about the approach is that the `updateServerEnvironment` method in the `homeView` fails silently. The idea here is that the environment updating flow may not be the best place to surface more general errors like the content missing or auth errors. Instead the errors from our API or the Connect API are captured and the web view is told there are no Environment Variables.

Most cases where our binaries API can fail this doesn't impact anything. Those cases are:
- the content record file doesn't exist
- the content record file is in error
- the deployment has not been deployed
- the credential is missing

Once a failure happens on Connect that is more note-able, but it felt out of the scope of this ticket to surface those errors.

#### Why request the server environment on Configuration changes?

The content record could have a new secret added.

#### Why request the server environment on Content Record changes?

The content record changes when a deployment is made, so we update the server environment afterwords to ensure that what is displayed is correct.

### DIFF
--- a/extensions/vscode/src/views/homeView.ts
+++ b/extensions/vscode/src/views/homeView.ts
@@ -68,7 +68,11 @@ import { DeploymentQuickPick } from "src/types/quickPicks";
 import { selectNewOrExistingConfig } from "src/multiStepInputs/selectNewOrExistingConfig";
 import { RPackage, RVersionConfig } from "src/api/types/packages";
 import { calculateTitle } from "src/utils/titles";
-import { ConfigWatcherManager, WatcherManager } from "src/watchers";
+import {
+  ConfigWatcherManager,
+  ContentRecordWatcherManager,
+  WatcherManager,
+} from "src/watchers";
 import { Commands, Contexts, DebounceDelaysMS, Views } from "src/constants";
 import { showProgress } from "src/utils/progress";
 import { newCredential } from "src/multiStepInputs/newCredential";
@@ -102,6 +106,7 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     this.initCompleteResolver = resolve;
   });
 
+  private contentRecordWatchers: ContentRecordWatcherManager | undefined;
   private configWatchers: ConfigWatcherManager | undefined;
 
   constructor(
@@ -134,6 +139,19 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
       useBus().trigger(
         "activeContentRecordChanged",
         await this.state.getSelectedContentRecord(),
+      );
+    });
+
+    useBus().on("activeContentRecordChanged", (contentRecord) => {
+      this.contentRecordWatchers?.dispose();
+
+      this.contentRecordWatchers = new ContentRecordWatcherManager(
+        contentRecord,
+      );
+
+      this.contentRecordWatchers.contentRecord?.onDidChange(
+        this.getContentRecordEnvironment,
+        this,
       );
     });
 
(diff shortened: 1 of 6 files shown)


extensions/vscode/src/watchers.ts
```diff
@@ -82,6 +83,31 @@ export class WatcherManager implements Disposable {
   }
 }
 
+/**
+ *  Manages file watchers for a specific Content Record
+ */
+export class ContentRecordWatcherManager implements Disposable {
```
dotNomad, line 89: This is very similar to our `ConfigWatcherManager` and I figured it may expand in the future to contain more than `.contentRecord` so kept the structure the same.

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -33,6 +33,7 @@ export const useHomeStore = defineStore("home", () => {
     return credentials.value.sort((a, b) => a.name.localeCompare(b.name));
   });
 
+  const serverSecrets = ref<Set<string>>(new Set());
```
dotNomad, line 36: I used a Set here to avoid iterating through the array each time, and Environment Variables are unique by name anyway.

extensions/vscode/src/views/homeView.ts
```diff
@@ -137,10 +142,24 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+        this.updateServerEnvironment,
+        this,
+      );
+    });
+
     useBus().on(
       "activeConfigChanged",
       (cfg: Configuration | ConfigurationError | undefined) => {
         this.sendRefreshedFilesLists();
+        this.updateServerEnvironment();
```
dotNomad, line 162: I did not end up getting to the only-update-on-config's-server-attribute feature. I figure I can re-visit this, but if we don't want to take on that technical debt I can address it in this PR.

marcosnav: I'm ok with this as is and wouldn't even consider it tech debt unless it had enough impact on performance. IMO this fulfills the goal of secrets syncing nicely without trying to include a more complex solution with extra checks that could expand edge cases.

dotNomad: Sounds great! I'm glad I didn't delve into that rabbit hole then

marcosnav (approved):
LGTM

PR #2415, opened 2024-10-31
Stop hidding details for renv lockfile reading errors

marcosnav:
## Intent

PR introducing some changes to stop hiding valuable error details when reading renv lockfile.

Instead of always displaying the message, `"Could not scan R packages from lockfile: renv.lock"`, we now allow some more details to be shown to the user related to the error in question.

**Screenshot of mismatching package version when reading renv.lock**
Screen Shot 2024-10-31 at 22 04 13

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Created a couple known agent errors to share more details with the user. Also, we are now providing some more error specifics for unrecognized renv.lock reading errors.

## Automated Tests

Created and updated unit tests as needed.

## Directions for Reviewers

More details about reading the renv.lock file can be seen when:
- [ ] The version on the renv.lock file and the library doesn't match (can be tested manually changing the version of a package in the lockfile)
- [ ] The version installed locally for a package is a dev version, newer than what is present in the lockfile

## Checklist

### DIFF
--- a/internal/inspect/dependencies/renv/manifest_packages.go
+++ b/internal/inspect/dependencies/renv/manifest_packages.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/posit-dev/publisher/internal/bundles"
 	"github.com/posit-dev/publisher/internal/logging"
+	"github.com/posit-dev/publisher/internal/types"
 	"github.com/posit-dev/publisher/internal/util"
 	"github.com/posit-dev/publisher/internal/util/dcf"
 )
@@ -155,8 +156,24 @@ func readPackageDescription(name PackageName, libPaths []util.AbsolutePath) (dcf
 	return nil, fmt.Errorf("%s: %w", name, errPackageNotFound)
 }
 
-var errLockfileLibraryMismatch = errors.New("package versions in library and lockfile are out of sync; use renv::restore() or renv::snapshot() to synchronize")
-var errMissingPackageSource = errors.New("can't re-install packages installed from source; all packages must be installed from a reproducible location such as a repository")
+var lockfileLibraryMismatchMsg = "package %s: versions in lockfile '%s' and library '%s' are out of sync. Use renv::restore() or renv::snapshot() to synchronize"
+var errMissingPackageSourceMsg = "cannot re-install packages installed from source; all packages must be installed from a reproducible location such as a repository. Package %s, Version %s"
+
+type renvReadErrDetails struct {
+	Lockfile        string
+	Package         PackageName
+	LockfileVersion string
+	LibraryVersion  string
+}
+
+func mkRenvReadErrDetails(lockfile string, pkg PackageName, lockVersion, libVersion string) renvReadErrDetails {
+	return renvReadErrDetails{
+		Lockfile:        lockfile,
+		Package:         pkg,
+		LockfileVersion: lockVersion,
+		LibraryVersion:  libVersion,
+	}
+}
 
 func (m *defaultPackageMapper) GetManifestPackages(
 	base util.AbsolutePath,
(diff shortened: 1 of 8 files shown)


internal/inspect/dependencies/renv/manifest_packages.go
```diff
@@ -206,12 +223,20 @@ func (m *defaultPackageMapper) GetManifestPackages(
...
-				pkg.Package, pkg.Version, description["Version"], errLockfileLibraryMismatch)
+			agentErr := types.NewAgentError(
+				types.ErrorRenvPackageVersionMismatch,
+				fmt.Errorf(lockfileLibraryMismatchMsg, pkg.Package, pkg.Version, description["Version"]),
+				renvErrDetails)
+			return nil, agentErr
 		}
 		if manifestPkg.Source == "" {
-			return nil, fmt.Errorf("package %s, version %s: %w", pkg.Package, pkg.Version, errMissingPackageSource)
+			agentErr := types.NewAgentError(
```
dotNomad, line 235: The idea behind these two agent errors is that we can provide some additional information that `renv` itself wouldn't give us right?

I just want to be sure that `renv` errors alone wouldn't capture these cases.

dotNomad (approved):
This looks great, I just had a question about the two agent errors that we create vs just raising any errors that occur during `GetManifestPackages`

internal/inspect/dependencies/renv/manifest_packages.go
marcosnav: Oh, I think there is some confusion. It seems like at some point we started building the idea that `renv` was being used at some point to read the `renv.lock`, but that is not the case.

We are reading the `renv.lock` file as JSON just with `go` code and structs to create the manifest for the deployment, and in the process of finding the lock file and reading it, we may encounter some errors which were the ones being hidden by a generic message.

The couple agent errors here are just proactive messages to:
- Prevent packages version mismatch for the deployment
- Be sure that the content will be able to run by not including local-only packages

We do call `renv` to generate a new `renv.lock` file, but this PR's fix is scoped only to when trying to read such file and error details were being discarded.

jonkeane: > Oh, I think there is some confusion. It seems like at some point we started building the idea that renv was being used at some point to read the renv.lock, but that is not the case.

OOOOOOOH I thought these were errors coming directly from `renv` this is my bad. Ok, now that I get that these are errors we are generating, this makes more sense. A few thoughts (though these become _much larger):

> * Prevent packages version mismatch for the deployment

Is this something we actually want to enforce? Also are we doing this check on our own with our own go code? Should we be? If this is a valuable feature (which I'm not convinced it is!!!), is this something we could defer to `renv` to manage? If we are imposing environment management (via `renv`) on folks, we should be leaning into and using as much of `renv` as possible to do this and not re-inventing and re-creating slight variations on those features.

> * Be sure that the content will be able to run by not including local-only packages

This is a good guard — I'm curious if there's a way to have `renv` do this already? There might not be, but maybe that's something we should add and upstream? This isn't the worst thing to re-create ourselves since the check is light / relatively well known / possibly slightly bespoke to Connect's use case *but* we only get the benefit of not calling out to `renv` if this is the only place we actually call out, and I don't think that's the case.

marcosnav: > Is this something we actually want to enforce? Also are we doing this check on our own with our own go code? Should we be? If this is a valuable feature (which I'm not convinced it is!!!)

Yes, this is done within the go code, I don't know if there is previous history with a use case that required this check to be in. I'll ask Mike. I agree that we might have a check that's not needed.

> This is a good guard — I'm curious if there's a way to have renv do this already? There might not be, but maybe that's something we should add and upstream? 

Looking for options I noticed that maybe `renv::status` might be a better solution here, even for the packages version mismatch. I'll play around a bit with that command and report here what I find.
https://rstudio.github.io/renv/reference/status.html

mmarchetti: The Go code is reading the `renv.lock` and then the `DESCRIPTION` file for each package in order to populate the manifest. If the version in the `renv.lock` doesn't match the installed package, what we read from `DESCRIPTION` won't be correct. It comes down to: what is the single source of truth for the package set we're deploying? We've said (so far) that it is renv.lock. If we had a better way to get the description for a given package/version, this check could go away and we could be independent of the current R library.

marcosnav:
Merged as is for now. We discussed in a team meeting the possibility of leveraging more on `renv` for checks, but we won't address that now. We talked about being diligent and keeping eyes open for more scenarios like these, where we shouldn't re-create solutions that third party tooling already solve.

PR #2418, opened 2024-11-06
Add Gradio support

edavidaja:
## Intent

Add support for deploying gradio applications.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

Added.

## Directions for Reviewers

Obtain a [sample gradio application](https://www.gradio.app/playground) and deploy it to Connect.

## Checklist


dotNomad:
Note that we will want to wait to merge this until the Connect release with Gradio support is released.

dotNomad (commented):
I was able to successfully deploy the hello world example however a few things:

On the Connect side I got errors, and it didn't behave as expected. I have a theory this is because of Connect, and not necessarily Publisher, but I'll want to verify that.

Additionally I think there is a question about if we want to add this to our documentation in the state this is currently in without inspection when selecting a gradio Python file

That isn't a criticism of your work @edavidaja but a question on how to handle this PR and when to merge it / announce it.

jonkeane:
> without inspection when selecting a gradio Python file

Presumably this is either something we add here or follow on with a PR? Or am I missing something else?

dotNomad:
> > without inspection when selecting a gradio Python file
> 
> Presumably this is either something we add here or follow on with a PR? Or am I missing something else?

No you aren't missing anything. I just think we need to make a decision on:
- do we think that updating the docs and communicating the gradio type to exists without our inspection helping is acceptable
- if not do we pick up the work to add that inspection before this merges / open another PR

marcosnav:
> do we think that updating the docs and communicating the gradio type to exists without our inspection helping is acceptable

IMO, It does not feel complete without the inspection part.

Another thing to consider, how does the publisher behave when publishing a `gradio` type app to a Connect instance that does not support it?. I'm assuming this is the first time the `publisher` adds a new content type that many Connect instances might not support?

dotNomad:
> Another thing to consider, how does the publisher behave when publishing a `gradio` type app to a Connect instance that does not support it?

This is great question. Ideally what it _should_ do is propagate the error we get back from Connect which should be meaningful about not supporting that content type.

We are getting into a bit of a larger design discussion here, but from what it sounds like we should be able to merge this, but perhaps hold off on including it in the changelog until the Gradio support is complete. We could create an epic on the board to track that.

jonkeane:
> We are getting into a bit of a larger design discussion here, but from what it sounds like we should be able to merge this, but perhaps hold off on including it in the changelog until the Gradio support is complete. We could create an epic on the board to track that.

This sounds good. Yes, in this case we definitely do not want to advertise this too heavily until it's officially added to Connect and that is released.

sagerb:
>Another thing to consider, how does the publisher behave when publishing a gradio type app to a Connect instance that does not support it?

Our approach has been so far, (following the goal of failing as fast as possible), to do the check within our prechecks, rather than wait for a server failure (which often comes back pretty raw - this one would come back as an invalid content type?).

We are already detecting if functionality is available or not within the functionality of `internal/clients/connect/capabilities.go`, and it would seem proper to add a check within the `CheckCapabilities` call. 

For example, we are already checking within that code if the target server supports a Python deployment. I see a strong parallel with Gradio here.

Does a Connect server that supports Gradio offer any way of querying its support via an API?

dotNomad:
The code looks good, but I am still seeing errors in Connect when I deploy the Hello World gradio example. @edavidaja  lets chat on Monday to figure out what is going on there, and we will get this PR buttoned up.

sagerb:
FYI, I'm working on an issue that will require a schema change (moving to v4 - new file). Since this PR includes an addition to the allowed content types, we'll need to coordinate the changes or we may lose a change. issue #2468.

dotNomad (approved):
Thanks for taking the time to diagnose this with me @edavidaja 

For those following along it was an older version of Gradio specified in the `requirements.txt`. Un-bounding that and getting a newer version fixed the errors I was seeing on the Connect side.

Approving and merging 🎉

PR #2430, opened 2024-11-13
Compare AppMode in pre-check not ContentType

dotNomad:
This PR changes our behavior in our pre-check comparing the App Mode on Connect and the Content Type we are trying to deploy.

Previously we compared the Content Type from our Config with the AppMode converted to a Content Type. With the introduction of the `quarto` and `quarto-static` types both leading to the same AppMode this comparison didn't quite work.

This changes the check to compare AppModes specifically since that is what truly can or cannot change, and handles our deprecated `quarto` type much better.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling

### DIFF
--- a/internal/clients/connect/client_connect.go
+++ b/internal/clients/connect/client_connect.go
@@ -539,11 +539,10 @@ func (c *ConnectClient) ValidateDeploymentTarget(contentID types.ContentID, cfg 
 	}
 
 	// Verify content type has not changed
-	log.Info("Verifying content type is the same")
-	existingType := ContentTypeFromAppMode(content.AppMode)
-	configType := cfg.Type
-	if existingType != configType {
-		msg := fmt.Sprintf("Content was previously deployed as '%s' but your configuration is set to '%s'.", existingType, configType)
+	log.Info("Verifying app mode is the same")
+	configAppType := AppModeFromType(cfg.Type)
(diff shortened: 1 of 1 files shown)


internal/clients/connect/client_connect.go
```diff
@@ -539,11 +539,10 @@ func (c *ConnectClient) ValidateDeploymentTarget(contentID types.ContentID, cfg
...
 
 	// Verify content type has not changed
-	log.Info("Verifying content type is the same")
-	existingType := ContentTypeFromAppMode(content.AppMode)
-	configType := cfg.Type
-	if existingType != configType {
-		msg := fmt.Sprintf("Content was previously deployed as '%s' but your configuration is set to '%s'.", existingType, configType)
+	log.Info("Verifying app mode is the same")
+	configAppType := AppModeFromType(cfg.Type)
+	if content.AppMode != configAppType && content.AppMode != UnknownMode {
```
dotNomad, line 544: The unknown check is being brought over from `main`. A small conflict should happen when we start bringing this all in.

marcosnav (approved):
💯

dotNomad:
> 💯

Thanks for the quick review, feel free to merge into your PR when ready.

PR #2434, opened 2024-11-14
Add notes on updating dependencies to `CONTRIBUTING`

dotNomad:
Updating our `CONTRIBUTING` document for the repo to include our flow around updating our dependencies.


CONTRIBUTING.md
```diff
@@ -187,3 +187,16 @@ This command will trigger the [Release GitHub Action](https://github.com/rstudio
...
+After a release has gone out we take the opportunity to update dependencies.
+
+This includes our JavaScript packages, Go version/packages, and tooling
+dependencies.
+
+Any significantly difficult dependency updates should be have an issue created
+to track the work and can be triaged alongside our other issues.
+
+Any updates to dependencies should be done in a separate PR to the release PR,
+and merged after the release has been tagged.
```
jonkeane, line 202: It might be sufficient to have the conversation here in the PR, but this doesn't preclude bumping dependencies whenever (and we'll get dependabot nudges anyway), just that we want to use releases as a reminder to do a sweep. Right?

dotNomad: > this doesn't preclude bumping dependencies whenever (and we'll get dependabot nudges anyway), just that we want to use releases as a reminder to do a sweep. Right?

Right, we can bump dependencies whenever, we don't _need_ to do it only after a release. Having a consistent process as that reminder helps to avoid the "we will do it later" mentality that is so easy to adopt.

That plus the issue creation outlined for the more difficult updates.

jonkeane: > Perhaps I should update that last line to be less specific about "Any updates to dependencies" to convey the way we are thinking about this?

That would be nice!

marcosnav (approved):
LGTM

PR #2435, opened 2024-11-14
Add emphasis to entrypoint TreeItem in Files

### LINKED ISSUE #2239: Emphasize the entrypoint in the Project Files tree
Relies on the files API correctly returning `isEntrypoint`.

---

There are a few ways we could do this.

Bolding is one option:

Image

but VS Code hardly ever bolds text.

Another is to use `list.activeSelectionForeground` and potentially `list.activeSelectionBackground` as well:

Image

the explorer does something similar so this may look like the file is open, or selected in some other way.

VS Code typically uses color over font-weight or other tools to emphasize text so maybe another color could be used. Ideally it would be a VS Code color that themes could change (to avoid light vs dark mode issues)

We could also do something as simple as including it in the details of the tree item:

Image

There are a lot of options.

dotNomad:
Adds the ability to emphasize our Tree Item components with a new `listStyle` prop option, then uses that to emphasize the entrypoint.

## Intent

Resolves #2239

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach was to use a CSS variable provided by VS Code, in this case `--vscode-list-activeSelectionForeground`. This is the same color as the active file when looking at the Explorer.

## User Impact

This provides the user a bit of glanceability to the understand what the "main"/entrypoint file is when looking at the Project Files view. It reinforces the most important file being deployed.

This is particularly helpful if the folder has multiple deployments with multiple entrypoints.

This option of using the same color as the active file in the Explorer view may be a bit confusing since our Project Files looks a bit like the Explorer pane, and has the "open file" icon button. This is opened as a draft to discuss this.

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
@@ -8,7 +8,13 @@
       file.reason?.source === 'built-in' ||
       file.reason?.source === 'permissions'
     "
-    :list-style="isFileIncluded(file) ? 'default' : 'deemphasized'"
+    :list-style="
+      isEntrypoint(file)
+        ? 'emphasized'
+        : isFileIncluded(file)
+          ? 'default'
+          : 'deemphasized'
+    "
     :tooltip="
       isFileIncluded(file)
         ? includedFileTooltip(file)
@@ -61,7 +67,10 @@ import { useHostConduitService } from "src/HostConduitService";
 import PostDecor from "src/components/tree/PostDecor.vue";
 import { ActionButton } from "src/components/ActionToolbar.vue";
 
-import { ContentRecordFile, FileMatchSource } from "../../../../../../src/api";
+import {
+  ContentRecordFile,
+  isConfigurationError,
+} from "../../../../../../src/api";
 import { WebviewToHostMessageType } from "../../../../../../src/types/messages/webviewToHostMessages";
 
 interface Props {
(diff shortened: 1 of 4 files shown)


extensions/vscode/webviews/homeView/src/components/views/projectFiles/TreeProjectFiles.vue
```diff
@@ -76,6 +85,14 @@ const props = withDefaults(defineProps<Props>(), {
 const home = useHomeStore();
 const { sendMsg } = useHostConduitService();
 
+const isEntrypoint = (file: ContentRecordFile): boolean => {
+  const config = home.selectedConfiguration;
+  if (config != undefined && !isConfigurationError(config)) {
+    return file.id === config.configuration.entrypoint;
+  }
+  return false;
+};
```
jonkeane, line 94: Oh, nice. It's nice that this code is closer to the actual display as well — keeps the logic closer to the thing that uses it.

I wonder why we thought we required this to come from the backend in the first place?

dotNomad: > I wonder why we thought we required this to come from the backend in the first place?

I don't quite recall, but I think we have a tendency to move logic to the backend if we do (or would) repeat it a lot on the display side. This may have been one of those we made the judgement too early calls and we didn't actually need `isEntrypoint` all that much.

marcosnav:
### Validation screenshot (entrypoint highlighted)
Screen Shot 2024-11-21 at 6 18 48

PR #2441, opened 2024-11-15
Address ESLint errors from recommended plugins

### LINKED ISSUE #1293: Use ESLint flat configurations
[ESLint v9.0.0 (which is currently in beta.1)](https://eslint.org/blog/2024/02/eslint-v9.0.0-beta.1-released/) will make the eslint.config.js configuration format the default and deprecate the current configuration method. In ESLint 10 our current configuration will no longer be supported.

Note that we cannot use ESLint v9 directly because the https://github.com/typescript-eslint/typescript-eslint/issues/8211 just yet, but can use the configuration format since it is supported in v8 (as long as our extends support it too).

See #1030 for more comments about this as things were figured out.

### References

- [ESLint 9.0 migration doc](https://eslint.org/docs/next/use/migrate-to-9.0.0)

dotNomad:
This PR is a follow-up to #2440 that addresses all of the new rules that are failing.

## Intent

Resolves #1293

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

Rather than change every `any` I figured we could make a follow-up issue to remove the `// eslint-disable-next-line @typescript-eslint/no-explicit-any` added here since those are a bit complex.

The work I did do here that wasn't auto-fixed was removed the async promise executors. To see why this is recommended I suggest reading the `no-async-promise-executor` rule docs: https://eslint.org/docs/latest/rules/no-async-promise-executor

## Directions for Reviewers

Test the changed code in the following flows:
- New Deployment flows
- Selecting a New or Existing Config
- The home view's `handleFileInitiatedDeploymentSelection` which fires when the entrypoint button is pressed

Most everything else is disabling `any` rules and changing to `const` where applicable.

### DIFF
--- a/extensions/vscode/src/actions/showAssociateGUID.ts
+++ b/extensions/vscode/src/actions/showAssociateGUID.ts
@@ -8,7 +8,7 @@ import { Views } from "src/constants";
 import { useApi } from "src/api";
 
 export async function showAssociateGUID(state: PublisherState) {
-  let urlOrGuid = "";
+  const urlOrGuid = "";
   const result = await window.showInputBox({
     title: "Enter the URL of the Existing Content Item on the Server",
     prompt: "Please provide the content's URL from Connect",
--- a/extensions/vscode/src/api/client.ts
+++ b/extensions/vscode/src/api/client.ts
@@ -57,7 +57,7 @@ class PublishingClientApi {
     this.entrypoints = new EntryPoints(this.client);
   }
 
-  logDuration(response: AxiosResponse<any, any>) {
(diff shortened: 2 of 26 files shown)


marcosnav (approved):
Nice! It catched a lot of things!

dotNomad:
> Nice! It catched a lot of things!

Yeah I'm really happy to get these recommended rules in to catch things like this!

PR #2447, opened 2024-11-19
Splitting a bit `homeView` and new tests for `PublisherState`

marcosnav:
## Intent

This PR aims to promote having a better test coverage in the project.
1. Pulls out some methods from `homeView` to `PublisherState` as the responsibility for those operations is better suited there.
2. Removes the dependency `@hypersphere/omnibus` and the existing bus instance. Usage of the bus wasn't justifying having it since calls were pretty localized and, on the contrary, complicated creating tests.
3. Pulled in `fishery` and created some factories to create instances of data for tests.
4. Included some new mock factory methods for `vscode` instances of `ExtensionContext`.
5. Adding new tests for `PublisherState` methods making use of all the previous points above.

## User Impact

No functional impact for the end user.

## Automated Tests

This is all about it.

## Directions for Reviewers

No specific directions. Open to feedback.

## Checklist


extensions/vscode/src/state.test.ts
```diff
@@ -0,0 +1,376 @@
...
+
+  test.todo("refreshContentRecords", () => {});
+
+  test.todo("refreshConfigurations", () => {});
+
+  test.todo("validConfigs", () => {});
+
+  test.todo("configsInError", () => {});
+
+  test.todo("refreshCredentials", () => {});
```
marcosnav, line 375: I'll keep working on these, to be landing in future PRs, but wanted to push and share the current progress.

extensions/vscode/src/state.ts
```diff
@@ -191,14 +203,27 @@ export class PublisherState implements Disposable {
   }
 
   async refreshContentRecords() {
-    const api = await useApi();
-    const response = await api.contentRecords.getAll(".", { recursive: true });
+    try {
+      await showProgress("Refreshing Deployments", Views.HomeView, async () => {
```
dotNomad, line 207: I'm not adamantly against it, but I do find it strange that the Home View doesn't control its own `showProgress` and instead the responsibility gets passed here.

Perhaps we could do something more generic on the `homeView` side?

extensions/vscode/src/types/shared.ts
```diff
@@ -18,11 +18,11 @@ export type PublishProcessParams = DeploymentSelector & {
...
-export type SelectionState =
-  | (DeploymentSelector & {
-      version: "v1";
-    })
-  | null;
+export type DeploymentSelectorState = DeploymentSelector & {
+  version: "v1";
+};
+
+export type SelectionState = DeploymentSelectorState | null;
```
dotNomad, line 25: Much clear 👍

extensions/vscode/src/state.test.ts
dotNomad, line 1: Great to see the tests getting started here!

dotNomad (commented):
Really like to see the removal of the bus for just function calls since everything is in Home View, and the tests 🎉 

I just had a single concern around `showProgress` moving to `state`

sagerb:
> Really like to see the removal of the bus for just function calls since everything is in Home View

Yes, this makes sense since we have moved away from the multiple classes, which required the bus for notifications.

sagerb (commented):
These changes look good and I've added my request to split up the unit tests a bit. I don't think that will be hard, you've pretty much got comments in place to identify the sub-tests. I think it is worthwhile at this point to use best practices on the granularity of the tests, rather than kicking that down the road.

Due to the diff, it was difficult to confirm that all of the functionality that used to be called from each of the bus event handlers was still maintained after your refactoring, but I spot-checked several, and they looked consistent.

extensions/vscode/src/state.ts
marcosnav: I understand that it could feel strange that the Home View doesn't control the show progress, but at the same time that might be good? 

I think it could be nice that every place using `PublisherState` does not have to remember to wrap methods in `showProgress`.

On the other hand, while taking a look to other methods, we don't use `showProgress` consistently, e.g: `getSelectedConfiguration` does not use it, I'm tempted to remove it at all on these methods.

It is nice to have visual feedback that something is happening but these operations do not seem to take enough time for the progress indicator to show up, unless I'm missing some background history on this. What do you folks think if we remove `showProgress` at all from these methods?

dotNomad: > I understand that it could feel strange that the Home View doesn't control the show progress, but at the same time that might be good?
I think it could be nice that every place using PublisherState does not have to remember to wrap methods in showProgress.

Yeah you are right. I always struggle a bit to place behavior like this 😓 

> What do you folks think if we remove `showProgress` at all from these methods?

These operations are very quick, but they can get slower if a user has a lot of them. I'm sure that will be very rare, if it happens at all, but in a large workspace it could. If it does occur it would be nice to have a progress bar.

PR #2466, opened 2024-12-06
Improved support for RMarkdown sites

### LINKED ISSUE #1643: Support R Markdown Site. Entrypoint being identified as a single R Markdown Doc.
Example: https://github.com/rstudio/connect-content/tree/main/bundles/rmd-site

This example is getting initialized as an `rmd` and Connect is treating it as an `R Markdown Document`.  It should be an `rmd-site` so Connect will treat it as an `R Markdown Site`.

Since it's an `R Markdown Document` the links are broken in Connect.

Here's the example deployed with publisher client:
https://rsc.radixu.com/connect/#/apps/7562e43e-af3d-4d67-9d72-4a34df032684/access/6545

And the same example published with the IDE:
https://rsc.radixu.com/connect/#/apps/e7c076ae-12ce-4027-b03e-db50446b2ae7/info/6547

marcosnav:
## Intent

**Update**
The direction of this PR changed, it isn't automatically including the files and assets that could be required by the project. This change of direction after discussing with the team and surfacing previous feedback from user testing in which users might prefer to add things manually, instead of including files that could represent a security risk, (e.g: pushing secrets unintentionally) 

---

**Don't be alarmed:**
From the 53 files shown in this PR 98% are artifacts used for testing, no need to code review those under `.../inspect/detectors/testdata/*`.

This PR introduces some changes to improve the experience of deploying RMD sites.

Fixes #1643 

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Connect requires the `manifest.json` file to specify `content_category` as `site` when the document is going to be a website, the common cases for this are projects using `_site.yml` configurations or `_bookdown.yml` projects.

The approach in this PR when there is a `_site.yml` or `_bookdown.yml` file present:
1.  Pick all the existing files and directories located at the level of the chosen entrypoint. This way, the user has a much better head start.
2. The generated `manifest.json` for the deployment bundle includes the category of `site`.

## User Impact

Deployment of RMD sites should have a much better out of the box experience.

## Automated Tests

Added and updated tests for the generated configuration and it's files, and for the update of the manifest file for the content category.

## Directions for Reviewers

**Update**
As mentioned above, with the change of direction in this PR, assets and files will need to be added manually after creating the deployment.

Aside from possible local renv requirements, the following examples can be used to try this out, sites should be easily deployed without much intervention after creating the deployment with Publisher:
- https://github.com/rstudio/connect-content/tree/main/bundles/rmd-site 
    - [ ] Links should work
- Bookdown demo: https://github.com/rstudio/bookdown-demo
    - [ ] Links and assets should look good
    - [ ] **Important note:** By default the bookdown demo has a configuration block to output PDF, handling `xelatex` system dependencies is not worth the work to test this, plain GitBook or HTML is enough. Updating the `_output.yml` to the block of content shown below does the trick.

```yml
bookdown::gitbook:
  css: style.css
  config:
    toc:
      before: |
        <li><a href="./">A Minimal Book Example</a></li>
      after: |
        <li><a href="https://github.com/rstudio/bookdown" target="blank">Published with bookdown</a></li>
    edit: https://github.com/rstudio/bookdown-demo/edit/master/%s
```

## Checklist


internal/bundles/manifest.go
```diff
@@ -202,6 +203,14 @@ func (manifest *Manifest) AddFile(path string, fileMD5 []byte) {
 	manifest.Files[path] = ManifestFile{
 		Checksum: hex.EncodeToString(fileMD5),
 	}
+
+	// Update manifest content category if a file being added is a site configuration
+	for _, ymlFile := range util.KnownSiteYmlConfigFiles {
+		if strings.HasSuffix(path, ymlFile) {
```
sagerb, line 209: I believe there is a potential problem here. While this will match what you intend, such as `/usr/bill/project/demo/_site.yml`, it would also match what I believe you don't want in that `/usr/bill/project/demo/abc123_mystuff_site.yml`, since the search is simply using `HasSuffix` rather than getting the filename from the path and then comparing that file name with the known site YML Config Files.

Perhaps you could consider using `internal/util/fspath.go`, and comparing the Base() against the known filenames? i'm not sure if you'll need to make that a lowercase test or not.

Or perhaps I'm missing something here?

internal/inspect/detectors/rmarkdown.go
```diff
@@ -32,6 +34,46 @@ func NewRMarkdownDetector(log logging.Logger) *RMarkdownDetector {
...
+	// StandardExclusions is also used when querying the configuration defined files.
+	// Then it returns hard stops and disables such files from the project tree.
+	// In this particular case, it also requires some extra exclusions,
+	// since this list is used while generating the configuration,
+	// it is desired to keep out these files at the beginning from the original generated configuration.
+	exclusionsExt := append(matcher.StandardExclusions,
+		"!.gitignore",
+		"!.github/",
+		"!renv/",
+	)
```
sagerb, line 52: I'm not following why you want to add these specifically for `rmarkdown` projects, vs all projects? Could you expand on why you didn't want to add these into the standard exclusion list within internal/bundles/matcher/walker.go?

sagerb (commented):
Code looks great, but I raised a few questions. I haven't tried to verify the functionality yet, that is my next step (and I'll update this comment with the result).

sagerb:
I was able to verify the functionality for the `rmd_site`:
- https://github.com/rstudio/connect-content/tree/main/bundles/rmd-site
  - [x] Links should work

I'm having problems getting the bookdown demo to run. Having problems installing packages within the directories of that repo, almost as if my GITHUB_PAT was invalid. I even went and generated a new one, set it up in my environment and made sure the R sessions were started from terminals with that change to my `.zshenv` file active. But still no success.

I'm also guessing a lot as to what would be the entrypoint for many of these content types... In the case of `rmd_site` I picked `index.Rmd`, and then picked the rmd content type. I believe that is correct?

I'm unsure if my experience with bookdown is a user error or an issue with my configuration. I was able to render the bookdown site, with the changes you specified for `_output.yml`, but trying to get renv happy with package dependencies set, is basically where I think I got hung up. Any help you can provide in that area, per how you set it up will allow me to verify that functionality as well faster. Thanks!

internal/inspect/detectors/rmarkdown.go
marcosnav: Sure thing, when the agent sends information for every file to the UI, `StandardExclusions` is used to add a piece of info to let the UI know that certain file(s) are not allowed to be included in the deployment whatsoever, disabling those in the project tree.

In this new case, we are using it to prevent adding items when a more global search of files and directories takes place for RMD sites, but users can still interact with such files if needed. We are leveraging on the `StandardExclustions` list and extending it.

E.g:
If we add `renv` dir directly to `StandardExclusions`, `activate.R` or any other script that the user could need would not be allowed, that's why here want it to be as an optional-initial exclusion.

internal/bundles/manifest.go
marcosnav: Adjustment done, got rid of the `HasSuffix` method, we can compare this filename directly since the possible site configuration files we look for must be present at the project root, we can ignore multi step paths. Thanks for noticing this!

internal/inspect/detectors/rmarkdown.go
sagerb: >In this new case, we are using it to prevent adding items when a more global search of files and directories takes place for RMD sites, but users can still interact with such files if needed. We are leveraging on the StandardExclustions list and extending it.

I think these two should be added to the `StandardExclusions`, if you feel they are important to exclude, as they apply to every other content type.

We already have in the `StandardExclusion` list the following for `renv`:
```
	// renv library cannot be included; Connect doesn't need it
	// and it's probably the wrong platform anyway.
	"!renv/library",
	"!renv/sandbox",
	"!renv/staging",
```
So I'm not sure why we wouldn't simply allow the other files within `renv` to be included if they are present. If they are there, it is probably because they serve a purpose, as RMDs can include scripts that require `renv` to be used, right? So I'd remove the `!renv` exclusion for `rmarkdown`, if I'm understanding correctly.

sagerb (commented):
Code changes looked good, thanks!

Further questions on the exclusions, but we're close... :-)

internal/inspect/detectors/rmarkdown.go
marcosnav: > I think these two should be added to the StandardExclusions, if you feel they are important to exclude, as they apply to every other content type.

Not that these are important to exclude, but we don't want them to be added in the beginning, the user should be allowed to add those if needed, might not be common, but we should not deny the user to check them in. Including those to `StandardExclusions` would not allow to pick them.

Similar for the `renv` directory, IMO we don't want it checked by default, honestly I'm not sure how much stuff can be within the `renv` directory but I think it is more likely that the user wants to have it out by default and allow then  pick manually what's needed from it.

I could be wrong though, it is just that at the moment makes more sense to me to be cautious with whats in the `renv` dir and let the user be explicit on what's needed from it.

sagerb: >Not that these are important to exclude, but we don't want them to be added in the beginning, the user should be allowed to add those if needed, might not be common, but we should not deny the user to check them in. Including those to StandardExclusions would not allow to pick them.

Yes, totally agree. So I guess I'm back to why do you feel that we need to have this alternate behavior for only the RMD projects? (My original question seems to stand: I'm not following why you want to add these specifically for rmarkdown projects, vs all projects?) None of the other walkers exclude these, but it seems like they should or shouldn't, as a single behavior, not one custom to each walker.

Let's move this to a chat next chance we can get, so we can resolve this question. Let me know when you're available. Thanks!

dotNomad: > None of the other walkers exclude these, but it seems like they should or shouldn't, as a single behavior, not one custom to each walker.

I believe our `StandardExclusions` for our file walker handles the Files API as well. Adding a generic `!renv` would mean that the directory wouldn't appear in the Project Files any longer (without other changes). The same goes for `.github/` and `.gitignore` here.

CleanShot 2024-12-10 at 19 49 36@2x
CleanShot 2024-12-10 at 19 49 48@2x

The difference can be seen between Project Files and the VS Code explorer. If I'm wrong in this analysis let me know, but I think the more unique exclusions in this file matcher make sense to me as this is the only time we are adding more than very specific files (entrypoint, package file) to the generated config.

dotNomad (commented):
The RMD site worked beautifully for me. I did have to select a `.Rmd` file for the inspection to work. Selecting `_site.yml` created an unknown type configuration with no included files. That is acceptable, perhaps a follow-up would be to allow that file to work for inspection just to cover all our bases.

Once I figured out some weirdness with `renv` on my system installing `bookdown` I was able to deploy with no issues. Similar to the comment above about using `_site.yml` as the entrypoint, using `_bookdown.yml` gives `type: 'unknown'` and could be a nice follow-up issue. Using a `Rmd` works perfectly.

I'll need to update the CHANGELOG in #2472 as reading the description gave me a different impression on how this worked. I'll get that updated before I mark it as ready to review.

---

Including all but a few files for sites makes sense to me, but understandably we got some push back when we initially had `files = ['*']` in the default configuration due to security concerns / accidentally deploying certain files.

The logic here includes a `.env` file for example which could contain environment variables in a pretty common use case.

We should consider this carefully, and perhaps tighten up the file inclusion to only match certain file extensions for example.

The generated `files` is the only bit of this PR I'm concerned with. The thorough testing you have done here makes it rock solid deployment-wise well done 👍

marcosnav:
> We should consider this carefully, and perhaps tighten up the file inclusion to only match certain file extensions.

Looking forward to chat about this in our standup, meanwhile...

> Selecting _site.yml created an unknown type configuration with no included files.

That's a good point! Specially when it is possible to select `_quarto.yml` for quarto websites, it is a small change so I'll add the check to proceed with a `site` type when the chosen entrypoint is a site configuration file.

marcosnav:
Pushed a new commit so that a configuration for R Markdown Site projects is created when picking `_site.yml` or `_bookdown.yml` files.

marcosnav:
@dotNomad @sagerb PR ready for another look, removed the logic that added files and directories and scoped the files discovery to only the entrypoint and the site configuration `.yml`.

dotNomad (approved):
Fantastic PR. Thank you very much @marcosnav for the changes, discussions, and the thorough testing. I tested with the two examples (`rmd-site` and the Bookdown demo) and both worked beautifully after the few changes you made.

Issue #1493, opened 2024-05-01
Don't allow entrypoint file to be excluded from Project


kgartland-rstudio:
You can exclude the entrypoint file from the Configuration File in the UI under Project Files.  Since the entrypoint is required, can we block that from being excluded? 

I know we can't stop them from manually editing the Configuration File but it seems like we can guard against them breaking themselves from doing it in the extension.

---

- [ ] ~Add a pre-flight check to ensure the entrypoint is included~ Superseded by a separate issue #2566
- [ ] Add a warning in the sidebar when the entrypoint is not included in the selected configuration
- [ ] Disallow the Files API from removing the entrypoint
- [x] Disable the checkbox in the Project Files view (handled in #2487)

PR #2487, opened 2024-12-12
Disable entrypoint checkbox in Project Files view

dotNomad:
This prevents the user from unchecking the entrypoint checkbox (and therefor removing the entrypoint from the `files` attribute in their configuration).

The configuration can still be edited by hand, but the sidebar no longer allows getting into this bad state.

## Intent

Part of #1493 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## User Impact

The user will no longer be able to uncheck the entrypoint's checkbox.

## Directions for Reviewers

Ensure that the entrypoint, no matter the details in the configuration, can be unchecked.

Checking the entrypoint checkbox should still be available if the configuration is manually edited and the entrypoint needs to be added.

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/tree/TreeItemCheckbox.vue
+++ b/extensions/vscode/webviews/homeView/src/components/tree/TreeItemCheckbox.vue
@@ -35,6 +35,7 @@
         :checked="checked"
         :disabled="disabled"
         class="tree-item-checkbox"
+        :class="{ 'opacity-100': disableOpacity }"
         @click="checked ? $emit('uncheck') : $emit('check')"
       >
         <span class="tree-item-title">{{ title }}</span>
@@ -77,6 +78,7 @@ interface Props {
   indentLevel?: number;
   expandable?: boolean;
   virtualized?: boolean;
+  disableOpacity?: boolean;
 }
 
 const props = withDefaults(defineProps<Props>(), {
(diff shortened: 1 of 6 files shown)


sagerb (commented):
Code looks great. I'll test shortly.

sagerb (commented):
Code functions well. I tested it with multiple existing configurations.

I do wonder if the entrypoint file should be visually indicated a little more than the slight opacity change which you implemented. Perhaps with another visual icon, asterick, something? 

I do think, regardless of the visual note above, that you should go ahead and add to the tooltip that the entrypoint cannot be removed from the deployment file list.

dotNomad:
> I do wonder if the entrypoint file should be visually indicated a little more than the slight opacity change which you implemented. Perhaps with another visual icon, asterick, something?

The entrypoint file should be emphasized with the color, and the opacity change is to avoid the disabled state making it semi-transparent (and less emphasized). We don't have any other icons or anything in that Project Files view so I'm not quite sure what exactly to do to further visually emphasize it. 🤔 

> I do think, regardless of the visual note above, that you should go ahead and add to the tooltip that the entrypoint cannot be removed from the deployment file list.

Great point. I'll get that in real quick ⌨️ 🏃

dotNomad:
Got the tooltip change in :)

sagerb:
Thanks for making the changes. @marcosnav I haven't verified the functionality after Jordan made the last changes. Can you test and merge?

marcosnav (approved):
### Confirmed latest changes

Screen Shot 2024-12-23 at 10 56 42

PR #2502, opened 2024-12-19
Enable coverage for home view Vitest

### LINKED ISSUE #2488: Add code coverage CLI checks for home view
Utilizing Vitest we can setup code coverage reporting, and CLI checks to fail when we do not hit thresholds.

dotNomad:
This PR enables `coverage` for our home view Vitest suite. Coverage can be seen when running vitest (`npm test`).

Vitest has a handy feature [`coverage.thresholds.autoUpdate`](https://vitest.dev/config/#coverage-thresholds-autoupdate) which will update the `vite.config.ts` with improved threshold numbers when `vitest` runs.

This also sets up failures in CI if:
- Coverage thresholds are not met
- Coverage thresholds in the config need to be updated

Both of those CI flows have been tested in this PR.

## Intent

Resolves #2488

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

The approach was to enable coverage for Vitest, and add CI checks to ensure that our current thresholds are met.

### DIFF
--- a/.github/workflows/home-view-unit-test.yaml
+++ b/.github/workflows/home-view-unit-test.yaml
@@ -15,3 +15,5 @@ jobs:
           cache-dependency-path: "**/package-lock.json"
       - run: npm install
       - run: npm test
+      - name: Check Vitest thresholds
+        run: git diff --name-only --exit-code vite.config.ts
--- a/extensions/vscode/webviews/homeView/.gitignore
+++ b/extensions/vscode/webviews/homeView/.gitignore
@@ -1,2 +1,4 @@
+coverage
+
 node_modules
 .vscode/*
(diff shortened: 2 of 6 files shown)


dotNomad:
Let's add something to the CONTRIBUTING doc about workflows around the coverage thresholds for the home view.

sagerb (approved):
Code changes make sense to me. As discussed, I'm fine with experimenting with this in place. I immediately wonder about the current values being used as the minimums, rather than picking a number we feel good about. But we will adjust! :-)

Issue #2492, opened 2024-12-17
When a credential is corrupted in keychain, allow the user to reset the stored data


marcosnav:
Issue stemmed up from Slack conversation:
https://positpbc.slack.com/archives/C06B0UZKT34/p1734380016863249

Investigations point out that the Gnome Keyring provides some data, although the credentials record cannot be found by the user, the corrupted credential error is instantiated only if there is a record to go through.

dotNomad:
Verified 🎉

PR #2504, opened 2024-12-20
Reset corrupted credentials storage when applicable

### LINKED ISSUE #2491: File credentials - failure to read file not clear for the user
Issue stemmed up from this Slack thread:
https://positpbc.slack.com/archives/C05K993PVEV/p1734368462380679

Although the error comes from having a file in an old format, if we have an issue with the new format, we shouldn’t fail the activation or at least surface the error to the user in a better way to take action. (User won't be able to deploy if credentials do not work, there might be questions that will pop up from this work)

marcosnav:
## Intent

This PR introduces a new reset step for credentials storage, both keyring and file based, when the data gets into a bad or unrecognizable state.

A warning will be displayed to let the user know that very likely it'll be necessary to recreate the credentials.

**Warning message screenshot**
Screen Shot 2024-12-24 at 1 44 35

**Output logged warning, notice the `credentials_service` log field to track the strategy being cleared up**
Screen Shot 2024-12-24 at 1 45 02
---
Screen Shot 2024-12-24 at 2 30 48

Fixes #2491 #2492

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

If we detect a malformed state of the credentials data, we proceed to reset it.

## User Impact

There were a couple reported cases internally in which the Publisher extension wouldn't finish loading and didn't provide much details other than some hints in the output logs.

This due to a corrupted or old and unsupported credentials data.

With this fix, the unusable credentials data will be reset and the user will be able to continue using Publisher.

## Automated Tests

Updated and added tests accordingly.

## Directions for Reviewers

Here is an example of malformed data that can be used to update the keychain record. This when keyring is supported, Desktop version of Positron or VSCode.
```
go-keyring-base64:eyI1ZjRmMjI4Ni00MjUzLTRjMTYtYWU1OS0yMDU5MjgyMmUyOWEiOnsiZ3VpZCI6IjVmNGYyMjg2LTQyNTMtNGMxNi1hZTU5LTIwNTkyODIyZTI5YSIsInZlcnNpb24iOjAsImRhdGEiOiJub3QtYS1qc29uIn19
```

For testing workbench, update the `.connect-credentials` file with an unrecognizable schema, e.g:
```
unrecognizable_key = "this will trigger a reset"
```

The previous credentials file should be backed up and can be found at `.credentials-file-{DATE}`

## Checklist


extensions/vscode/src/utils/errorTypes.ts
```diff
@@ -162,6 +163,15 @@ export type ErrInvalidConfigFiles = MkErrorDataType<
...
   mkErrorTypeGuard<ErrInvalidConfigFiles>("invalidConfigFile");
 
+// Invalid configuration file(s)
+export type ErrCredentialsReset = MkErrorDataType<"credentialCorruptedReset">;
+export const isErrCredentialsReset = mkErrorTypeGuard<ErrInvalidConfigFiles>(
+  "credentialCorruptedReset",
+);
+export const errCredentialsResetMessage = () => {
+  return "Stored credentials for Posit Publisher found in an unrecognizable state. A reset was required in order to proceed.";
+};
```
sagerb, line 173: The current wording might leave the user trying to figure out how to reset the credentials even though we have already done it for them. It's long and could use some further refinement, but how about something like this?
>Unable to use existing stored credentials for Posit Publisher (reason: unrecognizable state). The existing credentials have been deleted. Recreate them using the Credentials section or the Command Pallet's "Posit Publisher: New Credential" command.

sagerb:
Could you expand a little bit on your statement:
>A warning will be shown if the IDE is already open, if a new Positron or VSCode session is open, we won't show a warning, since the early stages of initialization of the extension do not communicate status with the extension.

Isn't the reason why we won't show a warning is because the storage would have already been reset?

dotNomad (commented):
> Here is an example of malformed data that can be used to update the keychain record. This when keyring is supported, Desktop version of Positron or VSCode.

Is there a good way for me to reproduce malformed data in my keychain record? 

> For testing workbench, update the `.connect-credentials` file with an unrecognizable schema

Testing on Workbench this works, but I am not getting the notification of the reset when reloading the page / accessing the session for the first time. If I set `.connect-credentials` to something invalid and refresh the home view I see the notification, but we will need to look at this flow to ensure that notification is shown.

That may be what you were talking about here:

> A warning will be shown if the IDE is already open, if a new Positron or VSCode session is open, we won't show a warning, since the early stages of initialization of the extension do not communicate status with the extension.

But like @sagerb I think I need some more details on this, and what we could possibly do to get around it.

extensions/vscode/src/utils/errorTypes.ts
dotNomad: I agree it should be very clear that something has already been done. Another suggestion for the re-word:

> Unrecognizable credentials for Posit Publisher were found and removed. Credentials may need to be recreated.

I went with "removed" instead of "deleted" to be a bit softer but still clear that they are no longer on the system. I left the recreation message at the end since it will be collapsed most of the time, and since they have already made credentials (we know this because they needed a reset) did not include instructions.

internal/credentials/credentials.go
```diff
@@ -86,21 +87,58 @@ type CredentialsService interface {
...
-		return krService, nil
+		err := enforceKeyringIntegrity(krService, log)
+		return krService, err
 	}
 
 	log.Debug("Fallback to file managed credentials service due to unavailable system keyring")
 	fcService, err := NewFileCredentialsService(log)
 	if err != nil {
-		return nil, types.NewAgentError(types.ErrorCredentialServiceUnavailable, err, nil)
+		return fcService, enforceFilecredsIntegrity(fcService, err, log)
```
dotNomad, line 106: I was a bit confused how the credentials were being reset right when the extension activated, but if I'm understanding correctly the enforcement happens before the APIs are even called since it occurs during the setup of the Credentials Service.

Perhaps we can do something with a different flow to ensure that notifications are always present?

Some potential ideas (these are really quick so take with a large grain of salt):
- have a `POST /api/credentials/reset` endpoint that we can call when we get certain errors
- only call the enforcement when the APIs are actually called

extensions/vscode/src/utils/errorTypes.ts
marcosnav: Thanks for the feedback, I agree. That's a much better warning message.

marcosnav:
@sagerb @dotNomad PR ready for another look. Changed a bit the way of resetting credentials, now it is always controlled via an endpoint and the warning message is consistent so the user will always be informed when we clear up the credentials data.

> Is there a good way for me to reproduce malformed data in my keychain record?

The only way I've found is updating it with bad data directly on keychain. I tried replicating it using only the Publisher but couldn't find a scenario that saves data in a bad state, so my assumption is that only an external influence on the system outside Publisher leads to a bad state.

jonkeane:
Thanks for jumping on this, Marcos!

I know that I'm coming to this late, but would there be any way for us to skip over invalid credentials when we parse these? We might be stuck here since we us a single file and we wouldn't be able to write back to that file a new, correct credential. But in an ideal world we wouldn't be required to wipe anything, just let the user know that there are no (valid) credentials, they should make a new one to continue. That would also make this more robust to other unknown failures. But it's possible the architecture we have with a single credential file and writing to that file prevents us from doing that.

A slightly separate issue (though it would be side-stepped if we could do the above) is that it's not super user-friendly to delete a file from the filesystem without any sort of confirmation or backup of what was deleted. I know that this circumstance should be pretty rare (basically only people at Posit who used earlier versions of the publisher), but it's not something we should do generally. We should either add a confirmation that tells someone we are about to delete that file and that if there's anything important in it they should check it, or we should create a backup version of it before we delete it (and we should tell folks we did that).

internal/credentials/file.go
```diff
@@ -207,6 +207,12 @@ func (c *fileCredentialsService) Delete(guid string) error {
 	return nil
 }
 
+func (c *fileCredentialsService) Reset() error {
+	c.log.Warn("Corrupted credentials data found. A reset was applied to stored data to be able to proceed.", "credentials_service", "file")
```
jonkeane, line 211: We will have a different behavior here anyway (see my comment on the PR), so this suggestion is a little moot, but I'm going to make it anyway since precisions and conciseness are skills and seeing edits is one way to learn that

```suggestion
	c.log.Warn("Corrupted credentials data found. The stored data was reset.", "credentials_service", "file")
```

marcosnav:
> I know that I'm coming to this late, but would there be any way for us to skip over invalid credentials when we parse these?

You are right that the work required to skip invalid credentials would require non-trivial effort. Since there are many possibilities for a file to not comply with TOML or with the expected schema.

> ...is that it's not super user-friendly to delete a file from the filesystem without any sort of confirmation or backup of what was deleted.

Yes, I was thinking that re-creating credentials in Publisher is an easy thing, however Connect does not display values for already created API Keys, so having to do that again is not great.

I like the idea of backing up the file, it is totally possible to `cp ~/.connect-credentials ~/.connect-credentials-{date}.backup` before wiping out the data, then include that detail on the warning message shown to the user. WDYT?

jonkeane:
> I like the idea of backing up the file, it is totally possible to cp ~/.connect-credentials ~/.connect-credentials-{date}.backup before wiping out the data, then include that detail on the warning message shown to the user. WDYT?

Sounds good. We'll want to craft the message to be clear + actionable (ideally also lead someone to that file easily), but that _should_ help folks not get stuck.

> Yes, I was thinking that re-creating credentials in Publisher is an easy thing, however Connect does not display values for already created API Keys, so having to do that again is not great.

Yeah, totally. In many cases creating a new API key is NBD, but if for whatever reason (the person doesn't remember how, they were given this key by someone else, they don't remember the URL(s) of the connect server(s) they had setup, etc.) it would be particularly painful.

marcosnav:
Updated PR, now backing up the previous credentials file and displaying to the user the backup file path.

**The message**
Screen Shot 2024-12-31 at 1 28 20

**Additional log with file backup path**
Screen Shot 2024-12-31 at 1 18 33

internal/services/api/reset_credentials.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/posit-dev/publisher/internal/credentials"
+	"github.com/posit-dev/publisher/internal/logging"
+	"github.com/posit-dev/publisher/internal/types"
+)
+
+// We can ignore errors related to malformed data, we are resetting it ain't we?
+func errIsNotLoadError(err error) bool {
+	return err != nil && !errors.Is(err, &credentials.LoadError{}) && !errors.Is(err, &credentials.CorruptedError{})
+}
```
dotNomad, line 18: We have this function in two spots - `provider_credentials` and here, can we combine them?

dotNomad (commented):
Overall this is looking and working much better. I'm seeing the reset every time now and it only happens when the extension interacts with credentials which is a large improvement in my opinion.

I tested on Workbench and it worked right away for the file based credential storage.

I made a couple of notes for code improvements, and since you are out @marcosnav I will take a look at the changes to get this merged in.

extensions/vscode/src/state.ts
```diff
@@ -283,11 +287,30 @@ export class PublisherState implements Disposable {
...
     }
   }
 
+  // Calls to reset all credentials data stored.
+  // Meant to be a last resort when we get loading data or corrupted data errors.
+  async resetCredentials() {
+    try {
+      const api = await useApi();
+      const response = await api.credentials.reset();
+      this.credentials = [];
```
sagerb, line 305: While it is logical to think that reset would simply result in empty credentials, I think it would be better to follow our existing patterns where we rely on the business logic to be implemented within the agent API. So in this case, we'd change the API result to return the list of credentials, which would be empty. That would be the value that would be updated within the extension.

extensions/vscode/src/utils/errorTypes.ts
```diff
@@ -185,5 +199,11 @@ export function resolveAgentJsonErrorMsg(err: axiosErrorWithJson) {
     return errPythonExecNotFoundErrorMessage(err);
   }
 
+  // Ignore errors coming from credentials being reset,
+  // a warning is shown when PublisherState is updated.
+  if (isErrPythonExecNotFoundError(err)) {
+    return errPythonExecNotFoundErrorMessage(err);
+  }
```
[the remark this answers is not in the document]
sagerb: @dotNomad This has to be a copy/paste error. I see no reason for the duplicated code. It should be removed.

internal/accounts/provider_credentials.go
```diff
@@ -11,9 +13,15 @@ type CredentialsProvider struct {
 	cs credentials.CredentialsService
 }
 
+// We can ignore errors related to malformed data on the initial loader
+// API consumers handle resetting malformed data when needed.
+func errIsNotLoadError(err error) bool {
```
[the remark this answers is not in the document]
sagerb: agreed.

internal/credentials/credentialstest/credentialstest.go
```diff
@@ -0,0 +1,42 @@
+package credentialstest
```
sagerb, line 1: Shouldn't this file be named `credentials_test.go` to match our pattern (and the unit test greps)?

internal/credentials/file.go
```diff
@@ -207,6 +209,50 @@ func (c *fileCredentialsService) Delete(guid string) error {
 	return nil
 }
 
+// Resets the credentials file
+// it is a last resort in case the data turns out to be irrecognizable
+// Returns the backup path of the original credentials file
+func (c *fileCredentialsService) Reset() (string, error) {
+	copyFilename, err := c.backupFile()
+	if err != nil {
+		return "", err
```
sagerb, line 218: Have we looked at what the user experience will be if we are unable to backup the file (with the raw errors able to be returned from the file operations within `.backupFile()`)? I would think we might want to capture these a bit and convert them into a nicer message for the user.

internal/services/api/reset_credentials_test.go
```diff
@@ -0,0 +1,127 @@
...
+	s.credservice.On("Reset").Return("", nil)
+
+	rec := httptest.NewRecorder()
+	h := ResetCredentialsHandlerFunc(s.log, s.credsFactory)
+	h(rec, req)
+
+	s.Equal(http.StatusOK, rec.Result().StatusCode)
+}
+
+func (s *ResetCredsSuite) TestReset_UnknownError() {
```
sagerb, line 92: If my suggestion to handle the raw file errors from the reset operations is implemented, then we should be able to make this a known error always, with the unknown being subjugated to a subError within the type.

sagerb (commented):
This is looking pretty good, but I would like to have my recommendations evaluated prior to merging this PR.

Please let me know if you have any questions or would like to discuss.

internal/credentials/credentialstest/credentialstest.go
dotNomad: The naming for the package fits. It is named similarly to others like `loggingtest`: https://github.com/posit-dev/publisher/blob/7830e62c2bfe119d8d47c3dcc9dce19bacef5371/internal/logging/loggingtest/mock_logger.go#L1

The file naming looks to be something like `mock_something.go` though. I can update that to be more similar to other bits of the codebase.

extensions/vscode/src/state.ts
dotNomad: I went back and forth on this comment for a while wondering if it was nitpicky, and a bit silly (since we know it is going to be `[]`) but decided to implement it. My reasoning was, if we add any features in the future to allow setting other credentials that the `reset` doesn't cover (think reading the environment variables or something) then this will be the more thorough code path.

internal/credentials/file.go
dotNomad: What do you think about handling this as a follow-up?

internal/accounts/provider_credentials.go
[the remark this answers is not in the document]
dotNomad: I left this the same since I couldn't find a great spot for the shared code. Next time we re-visit it we can change it if necessary.

internal/credentials/file.go
sagerb: We could certainly do it in a separate PR, but I think we should implement it prior to releasing it.

sagerb (approved):
Jordan and I agreed on handling low-level file errors when resetting credentials within a separate issue (https://github.com/posit-dev/publisher/issues/2513).

I'm good to have this PR merged now, with the changes that Jordan has done within https://github.com/posit-dev/publisher/pull/2512

PR #2505, opened 2024-12-20
Update home webview dependencies

dotNomad:
This updates the dependencies for the home webview. This takes care of #2490


sagerb (approved):
LGTM, approving since passing CI.

Issue #2383, opened 2024-10-10
Document collaboration flow


dotNomad:
- Add a doc
- We may have to add / refine features during this push
- Add to the VS Code README feature list

Context: https://github.com/posit-dev/publisher/pull/2352#discussion_r1793539296

dotNomad:
It isn't entirely clear where this collaboration document will live. Perhaps in the Posit Connect how-to guides or the user guide?

I am getting something written up so it is clear what will be in the document, and from there we can find a good spot for it.

PR #2507, opened 2024-12-21
Add collaboration workflow document

dotNomad:
This PR adds a document describing the collaboration flow we envision for Publisher as well as a section in the extension read me as a feature.

It isn't clear if this document will actually live in the repo, or in the Posit Connect docs. This is put up here for review. 

Tackles #2383


dotNomad:
Here is a screenshot I have saved to include in the doc.

publisher-missing-credential

docs/collaboration.md
```diff
@@ -0,0 +1,66 @@
...
+workflow guide shows how to collaborate with others on a piece of content.
+
+## Initial setup
+
+To ensure that others can collaborate without issue the deployment and
+configuration files should be available.
+
+That means including them in the repository for the project or including them
+in the `files` attribute in the configuration so they are available in the
+content bundle.
```
sagerb, line 16: With my earlier comment in mind, maybe start with a publisher creating a project that is not being collaborated on and then take the next steps to make sure it can be collaborated on, then switch to the collaborator's tasks?

docs/collaboration.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Start from a repository
+
+Clone the repo
+
+### Start from a content bundle
+
+Download the Source Version content bundle. If .posit files are included.
+
+They are included by default
```
sagerb, line 39: not sure about the period at the end of line 37 vs this next line of being included by default? 

Were you trying to address the fact that if publisher is used to deploy the project, the `.posit` files are included by default? 

That brings up another scenario we should address, when publisher was NOT used to deploy a project, what are the steps required to get collaboration to occur.

docs/collaboration.md
```diff
@@ -0,0 +1,66 @@
...
+
+### First time deploying
+
+If the deployment is pointed to a server that you have not deployed to before,
+you may need to setup your credentials.
+
+Posit Publisher makes this easy by prompting you to _Create a new Credential_
+when the selected deployment requires a credential you don't have.
+
+TODO: SCREENSHOT
```
sagerb, line 55: ok, now I see this is a work in progress, as you've got the screenshot saved...

Sorry for the initial comments if it is not yet ready.

docs/collaboration.md
```diff
@@ -0,0 +1,66 @@
...
+## While collaborating
+
+While collaborating on a piece of content, stay updated with the latest changes
+others have made.
+
+To do so, pull down the latest changes from the repository or download the
+content bundle from the latest Source Version on Posit Conect.
+
+Doing so ensures that you have the most up-to-date content, but also
+configuration settings and deployment details.
```
sagerb, line 66: We should also consider some of the impacts that the different content types have on collaboration, such as the interpreter versions and if dependencies are being updated. 

This should include a special clarification for R environments, where for now, collaborators need to agree upon runtime versions and maintain the appropriate package installations. 

It's less of an issue for Python and Quarto, so I think we can skim over the details for them, while still wanting to promote best practices of using the same versions while collaborating.

extensions/vscode/README.md
```diff
@@ -69,6 +69,18 @@ See the
 [Updating Previously Deployed Content](https://github.com/posit-dev/publisher/blob/main/docs/vscode.md#updating-previously-deployed-content)
 documentation for more information.
 
+### Seemless collaboration
+
+Collaborate easily with your team by sharing your configurations and
+deployments. Others can pick up where you left off and
+
+Allow collaborators to easily pick up where you left off by using the `.posit`
```
sagerb, line 77: Should we switch this to call out the `.posit/publisher` files rather than just the `.posit` files?

sagerb (commented):
Here are some preliminary thoughts for you to consider. I only realized that the docs were still a WIP after seeing the image hadn't made it in yet.

docs/collaboration.md
```diff
@@ -0,0 +1,66 @@
...
+[Posit Publisher](https://github.com/posit-dev/publisher)—the extension to
+publish content to Posit Connect using
+[Visual Studio Code](https://code.visualstudio.com/) and
+[Positron](https://positron.posit.co/)—enables seemless collaboration. This
+workflow guide shows how to collaborate with others on a piece of content.
+
+## Initial setup
+
+To ensure that others can collaborate without issue the deployment and
+configuration files should be available.
```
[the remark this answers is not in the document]
dotNomad: Definitely. With the way I work on documents like this I brain dump, trim, and repeat. I definitely think the explanation should be higher up.

docs/collaboration.md
dotNomad: That is interesting. I don't want to repeat too much that is in the ["Publishing from VS Code or Positron" document](https://docs.posit.co/connect/user/publishing-positron-vscode/) but I do like the idea of starting from the point of view of setting up a project for collaboration.

docs/collaboration.md
dotNomad: This is just unstructured brain dumping on my part. The intention here was to note that if the content bundle doesn't include the `.posit` files I will need to write something.

> when publisher was NOT used to deploy a project, what are the steps required to get collaboration to occur.

This is exactly what I mean to include in the final document. It will probably be a two part suggestion of:
- setup the `.posit` files by doing the initial setup above
- re-publish

It will take some refining to avoid repeating too much information. I'm not sure if the "As a collaborator" and "Initial Setup" sections are the most helpful yet.

docs/collaboration.md
dotNomad: Definitely. I think a note about how package files, and versions in the configuration would be good to include.

extensions/vscode/README.md
dotNomad: Yes that is a great point. I'll get that updated.

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+[Visual Studio Code](https://code.visualstudio.com/) or
+[Positron](https://positron.posit.co/)—enables seemless collaboration. This
+workflow guide explains how to collaborate with others on a deployed piece of
+content.
+
+## Deployment and Configuration files
+
+Posit Publisher creates and utilizes two files for deploying content. Both are
+inside of the `.posit/publish` directory in the same directory as
+your entrypoint file.
```
dotNomad, line 14: We could potentially include an example of the structure

For example the directory structure looks like this:

```
your-project/
├── /.posit
|   ├── /publish
│   |   ├── /deployments
│   │   |   └── deployment-58A3.toml
│   └── content-title-061E.toml
|   ...
└── entrypoint.py
```

But I couldn't find a good spot to place it, or decide if it was helpful or overwhelming.

dotNomad:
> Here is a screenshot I have saved to include in the doc.

I ended up removing the section about first-time deployment without a credential, but if we want this included I can update the screenshot.

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
+# Collaborative Publishing with Posit Publisher
+
+[Posit Publisher](https://github.com/posit-dev/publisher)—the extension to
+publish content to Posit Connect using
+[Visual Studio Code](https://code.visualstudio.com/) or
+[Positron](https://positron.posit.co/)—enables seemless collaboration. This
```
jonkeane, line 6:
```suggestion
[Positron](https://positron.posit.co/) or [Visual Studio Code](https://code.visualstudio.com/)
—enables seamless collaboration. This
```

One moving Positron up front, one typo

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+The Deployment file contains details on where the content was deployed and its
+state. Deployment files are automatically generated by Posit Publisher,
+updated during deploys, and not meant to be edited.
+
+An example of a Deployment file is
+`/.posit/publish/deployments/deployment-58A3.toml`.
+
+Utilizing both of these files, alongside a credential and collaborator access
+on Posit Connect, you and your collaborators can update and configure a single
+piece of content.
```
jonkeane, line 36: Would it make sense to be even more explicit here and say: you should generally not need to open or think about this file, though you *should* track it and make updates to it when git (or whichever RCS you're using) tells you do?

docs/collaboration.md
jonkeane: Agreed, I think this is too under the covers — especially because there's not much someone can do with this information. It's not like we offer them alternative layouts and want to show what our recommended one is.

jonkeane, line 14: Should we mention here about adding these to version control?

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Source control setup
+
+Deployment and Configuration files found in `./posit/publish` are intended
+to be committed into source repositories.
+
+Once tracked in source control, anyone with access can easily collaborate on the
+same piece of content.
+
+After a deploy be sure to commit any changes to the Deployment and Configuration
```
jonkeane, line 55:
```suggestion
After a deploy be sure to commit any changes to the Configuration and Deployment
```

This makes them alphabetical order which is good, but also we want to deemphasize the Deployment file whenever possible yeah?

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+files. Keeping them in sync with the content on Posit Connect will ensure that
+collaborators have the most up-to-date details to continue where others left
+off.
```
jonkeane, line 58: Nothing you need to do here, but reading this it struck me like we've all come around to: we really should be looking to the source of truth on the Connect server itself rather than keeping this information in a file on a client computer's filesystem. Then it entirely obviates the need for this sentence (among many other more impactful things).

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Content bundle setup
+
+If you and your collaborators aren't using source control, a content bundle
+containing the source code and data for a piece of content can be downloaded
+from Posit Connect. Content bundles are created when content is published and
+they work great with Posit Publisher.
```
jonkeane, line 65: > a content bundle containing the source code and data for a piece of content can be downloaded from Posit Connect

I'm not sure we need to be absolutely pedantically correct in this document, but this is true if and only if someone uploaded the source code alongside the bundle. There are a number of circumstances where people don't (e.g.  because the computation to render takes too long, the render requires access to systems that the connect server doesn't have).

> Content bundles are created when content is published and they work great with Posit Publisher.

I don't fully understand what's being conveyed here. That the bundle is created as a (hidden) step when publishing and so even if one has never gone and _made a bundle_ on their own, they still have them? Something else? 

As it reads now, it sounds to me almost like "once the content gets to the Connect server it makes a bundle of the source code of whatever produced it" which isn't quite accurate. But also maybe this is slightly duplicative of the sentence above so could be just deleted?

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+When deploying, include the Deployment and Configuration files so others can
+use them to collaborate after they download the content bundle.
```
jonkeane, line 68: Maybe this should be higher up? This is the big "if you do this things will generally just work" bit. Like if someone reading this section could read and remember a single sentence it's this one yeah?

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+See the
+[Posit Connect User Guide - Source Versions](https://docs.posit.co/connect/user/source-versions/)
+documentation for more information on the content bundle source versions and
+how to download them.
+
+> [!NOTE]
+> The content bundle will only contain files sent to Posit Connect. Be sure any
+> files that you would like collaborators to have access to are in the
+> Configuration's included `files`.
+> For more details see the [Project Files documentation for Posit Publisher](./vscode.md#project-files).
```
jonkeane, line 82: I know we don't (yet) have a huge number of call outs in the publisher doc, but across the Connect docs we have a lot and so are trying to eliminate them. Could we make this not-a-note?

Also: I'm not totally sure this is necessary here in the docs? It's absolutely true that if some file is important for collaboration they need to include it, but also if they are missing it from the deploy it's almost certainly going to fail to deploy at all, right? And (like linked there) we talk about project files when deploying elsewhere. This isn't a unique circumstance in the collaboration flow. So maybe we 🚮 and don't need to think about is it a note or not?

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Tips
+
+### Stay updated while collaborating
+
+While collaborating on a piece of content, stay updated with the latest changes
+others have made.
+
+To do so, either pull down the latest changes from the repository or download
```
jonkeane, line 91:
```suggestion
To do so, either pull down the latest changes from your source control system or download
```

Most (all?) other places we call this source control and I think it would be good to keep it here for consistency (and it avoids ambiguity around the word `repository`)

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+### First time deploying
+
+If a deployment is pointed to a server that you have not deployed to before,
+you may need to setup your credentials.
+
+Posit Publisher makes this easy by prompting you to _Create a new Credential_
+when the selected deployment requires a credential you don't have.
+
+TODO: SCREEN SHOT
```
jonkeane, line 105: This is covered in our general publishing docs, yeah? If yeah I think we can 🚮

jonkeane (commented):
Sorry for all the notifications — I wanted to make sure these got to you even if I didn't finish the full review. I did get through the whole thing now though. Thank you for all the work you put in here, this is a big chunk of text and I like the overall flow.

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+Once tracked in source control, anyone with access can easily collaborate on the
+same piece of content.
+
+After a deploy be sure to commit any changes to the Deployment and Configuration
+files. Keeping them in sync with the content on Posit Connect will ensure that
+collaborators have the most up-to-date details to continue where others left
+off.
+
+## Content bundle setup
```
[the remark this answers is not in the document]
m--: Minor, but if we go this route, we might be able to drop the "Collaborating" (since that's what this document is about) and update the title of the above section to With source control. So:

## With source control
...

## Without source control
...

docs/collaboration.md
m--: if the content author did not include the source code, is it possible to collaborate in this way?

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+If you and your collaborators aren't using source control, a content bundle
+containing the source code and data for a piece of content can be downloaded
+from Posit Connect. Content bundles are created when content is published and
+they work great with Posit Publisher.
+
+When deploying, include the Deployment and Configuration files so others can
+use them to collaborate after they download the content bundle.
+
+Posit Publisher includes both files automatically in configurations it generates
+to make this effortless.
```
m--, line 71: I'd make this part of the previous paragraph, since it's part of the same thought. So something like:

```suggestion
When deploying, include the Configuration and Deployment files so others can use them to collaborate after they download the content bundle. Posit Publisher includes both files automatically in configurations it generates.
```

extensions/vscode/README.md
```diff
@@ -69,6 +69,14 @@ See the
 [Updating Previously Deployed Content](https://github.com/posit-dev/publisher/blob/main/docs/vscode.md#updating-previously-deployed-content)
 documentation for more information.
 
+### Seemless collaboration
```
m--, line 72:
```suggestion
### Seamless collaboration
```

extensions/vscode/README.md
```diff
@@ -69,6 +69,14 @@ See the
...
+
+Collaborate with your team using shared configurations and deployments. Easily
+pick up where others left off with deployed content.
+
+See the [Collaboration Workflow](https://github.com/posit-dev/publisher/blob/main/docs/collaboration.md)
+documentation for more information.
```
m--, line 78: Minor, I might suggest using the title of the document as the link text, since it's a good explanation of what they'll read when they get there (and the page does not describe an individual workflow per se, but has options depending on how various systems are set up):

```suggestion
See [Collaborative Publishing with Posit Publisher](https://github.com/posit-dev/publisher/blob/main/docs/collaboration.md) for more information.
```

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
+# Collaborative Publishing with Posit Publisher
+
+[Posit Publisher](https://github.com/posit-dev/publisher)—the extension to
+publish content to Posit Connect using
+[Visual Studio Code](https://code.visualstudio.com/) or
+[Positron](https://positron.posit.co/)—enables seemless collaboration. This
+workflow guide explains how to collaborate with others on a deployed piece of
+content.
```
m--, line 8: Feel free to ignore if this is standard terminology, but "piece of content" is always so awkward to me 😅 we had similar terminology on Cloud and I looked for alternative phrases if they were not too stilted. Would this work?

```suggestion
workflow guide explains how to collaborate with others on deployed content.
```

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+`/.posit/publish/deployments/deployment-58A3.toml`.
+
+Utilizing both of these files, alongside a credential and collaborator access
+on Posit Connect, you and your collaborators can update and configure a single
+piece of content.
+
+## Collaborator access
+
+The original creator of a piece of content on Posit Connect needs to give others
+collaborator access to update the content.
```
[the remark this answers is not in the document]
m--: and we could do "needs to" -> "must", if we even want fewer words. I might also suggest "original creator of a piece of content" -> "content author" — it runs the risk of implying this can be granted generally, but I think the link to the Connect docs should clear that up.

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+
+When deploying, include the Deployment and Configuration files so others can
+use them to collaborate after they download the content bundle.
+
+Posit Publisher includes both files automatically in configurations it generates
+to make this effortless.
+
+See the
+[Posit Connect User Guide - Source Versions](https://docs.posit.co/connect/user/source-versions/)
+documentation for more information on the content bundle source versions and
```
m--, line 75:
```suggestion
See
[Posit Connect User Guide - Source Versions](https://docs.posit.co/connect/user/source-versions/)
for more information on the content bundle source versions and
```

I think this reads fine without "the ... documentation" words, but if you prefer them here feel free to keep them.

docs/collaboration.md
dotNomad: Fixed the seamless typo in https://github.com/posit-dev/publisher/pull/2507/commits/11ab74221986acc238fff506e565f17c31c83abf

The ordering is matching our Connect User Guide document: https://docs.posit.co/connect/user/publishing-positron-vscode/

docs/collaboration.md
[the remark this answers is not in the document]
dotNomad: I simplified this instruction a bit user the word reduction that @m-- suggested and included the "on Posit Connect" that @jonkeane suggested.

https://github.com/posit-dev/publisher/pull/2507/commits/b7dcdf41c8d9184e4a44c2e0f92f9f3840d43f1c

docs/collaboration.md
dotNomad: I switched up the order in the places where they are mentioned in tandem: https://github.com/posit-dev/publisher/pull/2507/commits/d1a331899032dbd0ba593c99a01cc0581f0af045

docs/collaboration.md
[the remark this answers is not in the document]
dotNomad: I changed up the headers in https://github.com/posit-dev/publisher/pull/2507/commits/762f4e8804c58fdb4749d7ac37794d7119b3705f to avoid "Content bundle" being the header that people will glance at to find what they are looking for. "Without source control" is much more glance-able.

docs/collaboration.md
dotNomad: > I'm not sure we need to be absolutely pedantically correct in this document, but this is true if and only if someone uploaded the source code alongside the bundle. There are a number of circumstances where people don't

I pulled this language straight from [the document I link to below](https://docs.posit.co/connect/user/source-versions/) to keep things uniform in explanation if that link is followed to understand more. We could be more Publisher specific instead and say that the content bundle includes all of the included files sent to Connect, but that isn't absolutely pedantically correct either.

> As it reads now, it sounds to me almost like "once the content gets to the Connect server it makes a bundle of the source code of whatever produced it" which isn't quite accurate. But also maybe this is slightly duplicative of the sentence above so could be just deleted?

I didn't think that first sentence enforced that every piece of published content had a content bundle to be potentially utilized with Publisher. It felt vague enough that if you didn't know what a content bundle was it wasn't clear that one was created on deploy. It is a bit duplicative so we could remove it if we think the first sentence is sufficient.

> if the content author did not include the source code, is it possible to collaborate in this way?

No it is not possible. That is why the `[!NOTE]` below was written to call out that this method only includes files actually sent to Connect.

docs/collaboration.md
dotNomad: Replaced in https://github.com/posit-dev/publisher/pull/2507/commits/de88b380ecec4692ae69ab7a9e71b080eb6927b3. 👍 great suggestion.

docs/collaboration.md
dotNomad: We don't mention this functionality in the [VS Code doc in the repo](https://github.com/posit-dev/publisher/blob/main/docs/vscode.md#credentials) nor in [the Connect User Guide doc](https://docs.posit.co/connect/user/publishing-positron-vscode/). I can move it into that VS Code document under Credentials though.

extensions/vscode/README.md
dotNomad: Thanks for catching this - fixed in: https://github.com/posit-dev/publisher/pull/2507/commits/e1ce73d49e28ac0fc3418e9dcccbdca622da0665

docs/collaboration.md
jonkeane: > No it is not possible. That is why the [!NOTE] below was written to call out that this method only includes files actually sent to Connect.


OOOOOOH that's what it was supposed to be noting. I didn't make that connection _at all_ when I read this through the first time, so I think we do need to do some work to tie those things together

docs/collaboration.md
dotNomad: I re-wrote the intro for the Deployment file section a bit to be more explicit about how to treat the file. I didn't outline that it should be tracked since the details about source control or utilizing the content bundle are  detailed below.

> The Deployment file details where the content was deployed and its state.
Deployment files are automatically generated by Posit Publisher and updated
during deploys. They are not meant to be edited. Posit Publisher will take care
of this file for you so you do not need to think about it aside from the
collaboration tips below.

docs/collaboration.md
dotNomad: I mentioned it at the beginning of the source control section, but we could definitely move it up. I tried to keep that separate to avoid bouncing any user that wasn't using source control.

docs/collaboration.md
dotNomad: Great point. I put the tip to commit the files right at the start of the source control section - I should do the same here.

docs/collaboration.md
dotNomad: Moved this over to the VS Code doc since it was more generally applicable.

docs/collaboration.md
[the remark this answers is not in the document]
dotNomad: Removing. If we move or copy this over to Connect docs I can always re-add it.

docs/collaboration.md
```diff
@@ -0,0 +1,105 @@
...
+## Deployment and Configuration files
+
+Posit Publisher creates and utilizes two files for deploying content. Both are
+inside of the `.posit/publish` directory in the same directory as
+your entrypoint file.
+
+#### The Configuration file
+
+The Configuration file describes how the content will be deployed to Posit
+Connect.
```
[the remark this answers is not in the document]
dotNomad: Worth a call out. I'll include that in the section 👍

docs/collaboration.md
dotNomad: I simplified these sections down to bullet points, and integrated the naming information into them. Thanks for the tip 👍

docs/collaboration.md
dotNomad: I removed all instances of "piece of content" from the doc. That wording is weird I agree.

docs/collaboration.md
dotNomad: I've written the "Without source control" section based on all of the feedback removing both the note, and trying to get across the information in a more condensed and simpler way.

Let me know if you think it communicates the intended information better.

docs/collaboration.md
dotNomad: I removed the note in https://github.com/posit-dev/publisher/pull/2507#discussion_r1927933840 trying to collapse it into the information above.

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
+# Collaborative Publishing with Posit Publisher
+
+This workflow guide explains how to collaborate with others on deployed content.
+
+## Configuration and Deployment files
+
+Posit Publisher creates and utilizes two `.toml` files for deploying content.
+Both are inside of the `.posit/publish` directory alongside your entrypoint
+file.
```
m--, line 9: This made me think the entrypoint file is in `.posit/publish` (it's not, right?), maybe a comma + an extra word would make it a bit clearer:

```suggestion
Both are in the `.posit/publish` directory, found alongside your entrypoint file.
```

(Also "inside of" -> "in", mostly due to inside/alongside in the same sentence.)

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
+# Collaborative Publishing with Posit Publisher
+
+This workflow guide explains how to collaborate with others on deployed content.
+
+## Configuration and Deployment files
+
+Posit Publisher creates and utilizes two `.toml` files for deploying content.
```
m--, line 7: I think "use" is slightly more accurate than "utilize" here, but I also think we can probably drop it altogether as the use is implied here by the creation:

```suggestion
Posit Publisher creates two `.toml` files for deploying content.
```

If we want a second (or different) verb, maybe maintains or something along those lines? But I think this is fine without.

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+Configuration and Deployment files found in `./posit/publish` are intended
+to be committed into source repositories.
+
+Once tracked in source control, anyone with access can easily collaborate on the
+same content.
+
+After a deploy be sure to commit any changes to the Configuration and Deployment
+files. Keeping them in sync with the content on Posit Connect will ensure that
+collaborators have the most up-to-date details to continue where others left
+off.
```
m--, line 39:
```suggestion
After a deploy, be sure to commit any changes to the Configuration and Deployment
files. Keeping them in sync with the content on Posit Connect ensures that
collaborators have the most up-to-date details to continue where others left
off.
```

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Without source control
+
+Include the Deployment and Configuration files in your deployment and others can
+use them to collaborate by downloading the files from Posit Connect. Posit
+Publisher includes both automatically in configurations it generates.
```
m--, line 45: The word "configurations" in the last line here makes me wonder if we're talking about the Configuration file. Maybe we can drop it? Does this still make sense? 

```suggestion
Include the Deployment and Configuration files in your deployment and others can
use them to collaborate by downloading the files from Posit Connect. Posit
Publisher includes both files by default.
```

(Switching "automatically" to "by default" since it's possible to change what happens, right?)

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+
+Whenever content is deployed to Posit Connect a content bundle is created
```
m--, line 47:
```suggestion
Whenever content is deployed to Posit Connect, a content bundle is created
```

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+### Stay updated while collaborating
+
+While collaborating on content, stay updated with the latest changes others have
+made.
+
+To do so, either pull down the latest changes from your source control system or
+download the content bundle from the latest Source Version on Posit Conect.
+
+Doing so ensures that you have the most up-to-date content, but also
+configuration settings and deployment details.
```
m--, line 68: The title of the section is "stay updated while collaborating", so I think you can drop the first paragraph, as it's essentially the same thing, and tweak the second to fold in the third (and fix a typo):

```suggestion
### Stay updated while collaborating

Depending on whether or not you're using source control, make sure to either pull down the latest changes from your source control system or download the content bundle from the latest Source Version on Posit Connect before deploying. This ensures that you have the most up-to-date content, configuration settings, and deployment details.
```
Added "before deploying" here so give the user some sense of _when_ they ought to do this. If there's a different recommended cadence here, we should specify whatever that is — maybe "before deploying or whenever your collaborators have made changes" but the sentence is already pretty long 🫣

docs/vscode.md
```diff
@@ -205,6 +205,16 @@ To remove a credential, right-click on it and select `Delete`.
 For instructions on how to create a Connect API key, see the
 [Connect documentation](https://docs.posit.co/connect/user/api-keys/index.html#api-keys-creating).
 
+#### First time deploying
+
+If a deployment is pointed to a server that you have not deployed to before,
+you may need to setup a credential.
```
m--, line 211:
```suggestion
you may need to set up a credential.
```

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+  settings.
+- The Deployment file details where the content was deployed and its state.
+  Deployment files are automatically generated by Posit Publisher in the
+  `deployment` directory and updated during deploys. They are not meant to be
+  edited. Posit Publisher will take care of this file for you so you do not need
+  to think about it aside from the collaboration tips below.
+
+Utilizing both of these files, alongside a credential and collaborator access
+on Posit Connect, you and your collaborators can update and configure the same
+content. To grant others collaborator access see
```
jonkeane, line 25:
```suggestion
Utilizing both of these files, alongside a credential and collaborator access
on Posit Connect, you and your collaborators can update and configure the same
content. To grant others collaborator access see
```

This is very minor, please feel free to ignore, but I found `alongside a credential and collaborator access` a bit odd here. It's of course true you would need a credential and collaborator access, but I'm not sure calling that out here is all that helpful? Is there something you're trying to call out here besides something like basic requirements? These both might be low-level enough that we don't need to list them (like we don't list things like a network connection to the connect server, a Python or R binary, etc.)

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+[Posit Connect User Guide- Access list settings](https://docs.posit.co/connect/user/content-settings/#set-collaborators).
```
jonkeane, line 26: Is the space after the dash intended?

docs/collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+## Without source control
+
+Include the Deployment and Configuration files in your deployment and others can
+use them to collaborate by downloading the files from Posit Connect. Posit
+Publisher includes both automatically in configurations it generates.
+
+Whenever content is deployed to Posit Connect a content bundle is created
+containing the files. Ensure files you would like collaborators to have access
+to are in the [Configuration's Project `files`](./vscode.md#project-files).
+Content bundles can be downloaded using the **Source Versions** tool.
```
jonkeane, line 50:
```suggestion
```

I would remove this (1) it's (probably) going to be renamed soon and (2) this belongs in the Connect docs more than anything. I doubt many folks would do this, but you could also get the bundles via the SDKs or the API directly

docs/collaboration.md
dotNomad: Great suggestion.

docs/collaboration.md
dotNomad: I trimmed the wording here just a touch, but otherwise kept it the same as your suggestion. 👍 

> from ~your~ source control ~system~

docs/collaboration.md
dotNomad: >  Is there something you're trying to call out here besides something like basic requirements? These both might be low-level enough that we don't need to list them (like we don't list things like a network connection to the connect server, a Python or R binary, etc

That is fair. I wanted to remind users that collaborator access should be granted when giving someone the files to collaborate (through Git or another method). I know frequently when collaborating on a deployment I have forgotten to add collaborator access until the other person let me know the deploy failed, so figured it was worth the callout here.

docs/collaboration.md
dotNomad: Good call. I do have the link to the connect docs right below as well to inform how to download content bundles so this was a bit duplicative.

docs/collaboration.md
dotNomad: Oops, there is a missing space there to match the other links. Great catch.

docs/collaboration.md
dotNomad: After looking at this for a minute I removed the "alongside a credential and collaborator access" bit, and just add a bit before the documentation link

> Utilizing both of these files you and your collaborators can update and
configure the same content. To grant others collaborator access to content see
[Posit Connect User Guide - Access list settings](https://docs.posit.co/connect/user/content-settings/#set-collaborators).

So there is still a link to follow if you hadn't collaborated before, or need the reminder, but it isn't as repetitive or calling out basic things.

PR #2512, opened 2025-01-06
PR feedback for credentials reset

dotNomad:
This contains some small, proposed changes to the #2504 PR.


sagerb (approved):
These look good, but I know you proposed a few other changes. Were you going to action those?

sagerb:
It looks like we were using `mock` as a suffix (found from https://github.com/posit-dev/publisher/blob/f909de635e70ef5bee23248bffe12b75a378c8b6/internal/inspect/python_mock.go)

extensions/vscode/src/state.ts
```diff
@@ -302,9 +302,11 @@ export class PublisherState implements Disposable {
     try {
       const api = await useApi();
       const response = await api.credentials.reset();
-      this.credentials = [];
       const warnMsg = errCredentialsCorruptedMessage(response.data.backupFile);
       window.showWarningMessage(warnMsg);
+
+      const listResponse = await api.credentials.list();
+      this.credentials = listResponse.data;
```
sagerb, line 309: Great, thanks!

dotNomad:
> It looks like we were using `mock` as a suffix (found from https://github.com/posit-dev/publisher/blob/f909de635e70ef5bee23248bffe12b75a378c8b6/internal/inspect/python_mock.go)

It looks like we have a mix. We can clean this up a bit in a separate bit of work.
CleanShot 2025-01-06 at 16 38 55

PR #2516, opened 2025-01-08
Simplify debugging the agent and the extension at the same time (full-stack)

sagerb:
… of the UX and the agent

## Intent

This PR updates the launch configuration for the vscode extension, to have two targets:
- `Debug with internal API agent`
- `Debug with external API agent on 9001`

The first one maintains the functionality we had before, but the second one supports the ability to launch the VSCode extension so that it expects there is an external (independently managed) API agent at port 9001.

`services.ts` has been updated to implement this functionality, but by default, will continue its prior behavior of managing the agent process itself. The functionality is protected by also checking that it has been launched in development mode so that we will not have this functionality leak out to customers.

I have updated the `extensions/vscode/CONTRIBUTING.md` file to reflect these changes.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## User Impact

Functionality is not exposed to customers by limiting extension logic to require execution in development mode.

## Automated Tests

Not applicable

## Directions for Reviewers

1. Confirm prior debugging capability is still maintained by using the launch configuration named `Debug with internal API agent`
2. Confirm new debugging capability works by following the updated steps within `extensions/vscode/CONTRIBUTING.md` for `Debug with external API agent on 9001`.
3. Build the extension and install the `vsix` file into a normal running VSCode instance. Confirm that the extension launches and talks with the agent, by viewing activity within the Publisher Output window.

## Checklist

### DIFF
--- a/extensions/vscode/.vscode/launch.json
+++ b/extensions/vscode/.vscode/launch.json
@@ -6,7 +6,7 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Debug with external agent on 9001",
+      "name": "Debug with external API agent on 9001",
       "type": "extensionHost",
       "request": "launch",
       "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
@@ -16,7 +16,7 @@
       "cwd": "${workspaceFolder}/../../test/sample-content"
     },
     {
-      "name": "Debug with internal agent",
+      "name": "Debug with internal API agent",
       "type": "extensionHost",
       "request": "launch",
       "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
--- a/extensions/vscode/.vscode/launch.json
+++ b/extensions/vscode/.vscode/launch.json
@@ -6,23 +6,23 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Debug with external API agent on 9001",
-      "type": "extensionHost",
-      "request": "launch",
-      "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
-      "outFiles": ["${workspaceFolder}/dist/**/*.js"],
-      "preLaunchTask": "${defaultBuildTask}",
-      "env": { "POSIT_PUBLISHER_USE_EXTERNAL_AGENT": "TRUE" },
-      "cwd": "${workspaceFolder}/../../test/sample-content"
-    },
-    {
-      "name": "Debug with internal API agent",
+      "name": "Run Extension",
       "type": "extensionHost",
       "request": "launch",
       "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
       "outFiles": ["${workspaceFolder}/dist/**/*.js"],
       "preLaunchTask": "${defaultBuildTask}",
       "env": { "POSIT_PUBLISHER_USE_EXTERNAL_AGENT": "FALSE" },
+      "cwd": "${workspaceFolder}/../../test/sample-content"
+    },
+    {
+      "name": "Run Extension using external API agent on 9001",
+      "type": "extensionHost",
+      "request": "launch",
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
+      "outFiles": ["${workspaceFolder}/dist/**/*.js"],
+      "preLaunchTask": "${defaultBuildTask}",
+      "env": { "POSIT_PUBLISHER_USE_EXTERNAL_AGENT": "TRUE" },
       "cwd": "${workspaceFolder}/../../test/sample-content"
     },
     {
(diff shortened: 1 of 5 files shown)


extensions/vscode/.vscode/launch.json
```diff
@@ -6,12 +6,24 @@
...
-      "name": "Run Extension",
+      "name": "Debug with external API agent on 9001",
       "type": "extensionHost",
       "request": "launch",
       "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
       "outFiles": ["${workspaceFolder}/dist/**/*.js"],
-      "preLaunchTask": "${defaultBuildTask}"
+      "preLaunchTask": "${defaultBuildTask}",
+      "env": { "POSIT_PUBLISHER_USE_EXTERNAL_AGENT": "TRUE" },
+      "cwd": "${workspaceFolder}/../../test/sample-content"
```
dotNomad, line 16: Are the `cwd` attributes important here? We didn't have it previously with the default Debug.

extensions/vscode/.vscode/launch.json
```diff
@@ -6,12 +6,24 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Run Extension",
+      "name": "Debug with external API agent on 9001",
```
dotNomad, line 9: Super nitpicky, but could we have the "Run Extension" option listed first so it is selected first as the debug option?

extensions/vscode/src/extension.ts
```diff
@@ -43,11 +49,24 @@ export async function activate(context: ExtensionContext) {
...
+  const useExternalAgent =
+    context.extensionMode === ExtensionMode.Development &&
+    process.env.POSIT_PUBLISHER_USE_EXTERNAL_AGENT === "TRUE";
+  console.log(
+    "Starting Context in extension mode: %s, with useExternalAgent set to %s",
+    context.extensionMode,
+    useExternalAgent,
+  );
+
+  let port = await ports.acquire();
```
dotNomad, line 61: Should we skip acquiring a port if we are using an external agent?

extensions/vscode/CONTRIBUTING.md
```diff
@@ -50,17 +50,46 @@ just test
...
+
+1. First launch a debug session from a VSCode window which has opened the repo's base directory
+   (see [API Debugging](./../../CONTRIBUTING.md#debugging-in-vs-code))
+2. Then launch another debug session (using the `Debug with external API agent on 9001` configuration)
+   from a separate VSCode window which has opened the `extensions/vscode` folder.
+
+Notes when using `Debug with external API agent on 9001`:
+
+- Both the agent's and extension's `launch.json` files are setup using the same working
+  subdirectory of `test/sample-content`. If you change it in one, you will need to change it in the other.
```
dotNomad, line 74: Ah this answers my question about `cwd` it seems like it is necessary for it to match? I don't remember ever adding `cwd` when I was doing simultaneous debugging, but perhaps you caught something?

extensions/vscode/CONTRIBUTING.md
```diff
@@ -50,17 +50,46 @@ just test
 
 Utilize VSCode's built-in debugger to troubleshoot and diagnose issues.
 
-Launch the debugger by pressing `F5` or go to `Run > Start Debugging`.
+Switch to the debugger extension view within VSCode and select one of the two
+debug configurations (specified within `extensions/vscode/.vscode/launch.json`):
+
+- `Debug with internal API agent`
```
dotNomad, line 56: I'd suggest simplifying the name to "Run Extension" since it implies a bundled agent already. Nitpicky though, feel free to ignore.

extensions/vscode/CONTRIBUTING.md
dotNomad, line 1: Thanks for adding the notes for how to accomplish simultaneous debugging 🎉

dotNomad (approved):
All of my review comments are nitpicks or questions. This is fantastic. Thanks for putting the work into this to make it easier to do.

extensions/vscode/.vscode/launch.json
sagerb: Yes, they are. You have to launch the agent within the same CWD as the extension has as an open workspace, otherwise, all paths go out of sync.

extensions/vscode/.vscode/launch.json
sagerb: Sure

extensions/vscode/src/extension.ts
sagerb: I can swap the logic yes. That would maintain that port available for other consumers, so it is a good point.

extensions/vscode/CONTRIBUTING.md
sagerb: It depended on what you were doing. If you were publishing, then there is too many cases where working directories need to match.

extensions/vscode/CONTRIBUTING.md
sagerb: no problem.

dotNomad (approved):
Thanks for making those changes, and for getting this in. It will be so helpful 🎉

Issue #2179, opened 2024-08-26
When canceling a deployment I immediately get a Last Deployment Successful message


kgartland-rstudio:
The `Last Deployment Successful` message is displayed immediately after a deployment is canceled.

https://github.com/user-attachments/assets/1d646f9a-4fd1-4ccb-958d-69ca7498245d

dotNomad:
We can use the same pattern we use for Secrets where we use the Connect API to get the status of the last deployment / publish time. We need to investigate if there is a good API endpoint for us to use for that, and what the work would entail.

We should double check that a deployment failure isn't hidden by a cancellation.

dotNomad:
The issue here is that cancellation doesn't actually cancel the deployment - it merely dismisses it from the sidebar and closes the VS Code notification.

Any deployment with a cancelled local ID has its events / callbacks suppressed but several events don't contain a `localId` to compare.

So what occurs is the deployment file can be written when we hit the "Create Deployment Record" step, but the actual failure could occur later - in the case of a failure happening during the validate step. Because the in progress deployment is dismissed and we are showing the "Last Deployment" status in the sidebar it shows every status as the deployment file is written.

To reproduce this you can create a deployment that deploys with no issue, but has an error when running. For example: having a typo on the `fastapi import`:

```
from fastapi import FstAPI
```

Deploying writes a valid deployment record, until validation occurs then we get:

```
[deployment_error]
code = 'deployedContentNotRunning'
message = 'Deployed content does not seem to be running. See the logs in Connect for details.'
operation = 'publish/validateDeployment'
```

dotNomad:
Currently we write the deployment record (via [`writeDeploymentRecord`](https://github.com/posit-dev/publisher/blob/cea9ca5d201389a796ede453c045e077e1cece19/internal/publish/publish.go#L201)):
- [in the middle of our `publishWithClient` before creating/uploading the bundle](https://github.com/posit-dev/publisher/blob/cea9ca5d201389a796ede453c045e077e1cece19/internal/publish/publish.go#L304)
- [at the end of creating and uploading the bundle](https://github.com/posit-dev/publisher/blob/20a3977dfd9b219042f50331f72fea3330dea289/internal/publish/bundle.go#L108)
- [and when we emit error events](https://github.com/posit-dev/publisher/blob/cea9ca5d201389a796ede453c045e077e1cece19/internal/publish/publish.go#L134)

[...]

PR #2525, opened 2025-01-11
Follow-on #2 for dismissed deployment support PR

sagerb:
## Intent

Resolves: https://github.com/posit-dev/publisher/issues/2179

This PR contains follow-on changes to https://github.com/posit-dev/publisher/pull/2522. These changes have been separated by request to facilitate reviews in progress.

This PR addresses:
- removing the  flash of success status when canceling a deployment.
- Also fixed filtering to include some log messages from connect which had the local ID value in snake_case rather than camelCase. This suppresses a message from the Connect server which occurred when a second redeployment was progressing while the original, cancelled deployment was still processing.

image

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling

### DIFF
--- a/extensions/vscode/src/events.ts
+++ b/extensions/vscode/src/events.ts
@@ -177,15 +177,24 @@ export class EventStream extends Readable implements Disposable {
   }
 
   private processMessage(msg: EventStreamMessage) {
-    const localId = msg.data.localId;
-    if (localId && this.canceledLocalIDs.includes(localId)) {
+    // Some log messages passed on from Connect include
+    // the localId using snake_case, rather than pascalCase.
+    // To filter correctly, we need to check for both.
+
+    const localIdForm1 = msg.data.localId;
+    if (localIdForm1 && this.canceledLocalIDs.includes(localIdForm1)) {
       // suppress and ignore
       return;
     }
-    // Trace message
-    // console.debug(
-    //   `eventSource trace: ${event.type}: ${JSON.stringify(event)}`,
-    // );
+    const localIdForm2 = msg.data.local_id;
+    if (localIdForm2 && this.canceledLocalIDs.includes(localIdForm2)) {
+      // suppress and ignore
+      return;
+    }
+
+    // // Trace message
+    // console.debug(`eventSource trace: ${msg.type}: ${JSON.stringify(msg)}`);
+
     // Add the message to the messages array
     this.messages.push(msg);
     // Emit a 'message' event with the message as the payload
--- a/extensions/vscode/src/events.ts
+++ b/extensions/vscode/src/events.ts
@@ -190,7 +190,7 @@ export class EventStream extends Readable implements Disposable {
     // Trace message
     // Uncomment the following code if you want to dump every message to the
     // console as it is received.
-    console.debug(`eventSource trace: ${msg.type}: ${JSON.stringify(msg)}`);
+    // console.debug(`eventSource trace: ${msg.type}: ${JSON.stringify(msg)}`);
 
     // Add the message to the messages array
     this.messages.push(msg);
--- a/extensions/vscode/src/events.ts
+++ b/extensions/vscode/src/events.ts
@@ -181,19 +181,16 @@ export class EventStream extends Readable implements Disposable {
     // the localId using snake_case, rather than pascalCase.
     // To filter correctly, we need to check for both.
 
-    const localIdForm1 = msg.data.localId;
-    if (localIdForm1 && this.canceledLocalIDs.includes(localIdForm1)) {
-      // suppress and ignore
-      return;
-    }
-    const localIdForm2 = msg.data.local_id;
-    if (localIdForm2 && this.canceledLocalIDs.includes(localIdForm2)) {
+    const localId = msg.data.localId || msg.data.local_id;
+    if (localId && this.canceledLocalIDs.includes(localId)) {
       // suppress and ignore
       return;
     }
 
-    // // Trace message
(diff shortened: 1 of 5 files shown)


extensions/vscode/src/events.ts
```diff
@@ -177,15 +177,24 @@ export class EventStream extends Readable implements Disposable {
   }
 
   private processMessage(msg: EventStreamMessage) {
-    const localId = msg.data.localId;
-    if (localId && this.canceledLocalIDs.includes(localId)) {
+    // Some log messages passed on from Connect include
+    // the localId using snake_case, rather than pascalCase.
+    // To filter correctly, we need to check for both.
+
+    const localIdForm1 = msg.data.localId;
```
dotNomad, line 184: If it is always one or the other can we grab either one?

```suggestion
    const localIdForm1 = msg.data.localId || msg.data.local_id;
```

extensions/vscode/src/events.ts
```diff
@@ -177,15 +177,24 @@ export class EventStream extends Readable implements Disposable {
...
       return;
     }
-    // Trace message
-    // console.debug(
-    //   `eventSource trace: ${event.type}: ${JSON.stringify(event)}`,
-    // );
+
+    // // Trace message
+    // console.debug(`eventSource trace: ${msg.type}: ${JSON.stringify(msg)}`);
+
```
dotNomad, line 197: Should we just remove this since it is commented out?

extensions/vscode/src/state.ts
```diff
@@ -167,6 +167,25 @@ export class PublisherState implements Disposable {
...
+  updateContentRecord(
+    newValue: ContentRecord | PreContentRecord | PreContentRecordWithConfig,
+  ) {
+    const existingContentRecord = this.findContentRecord(
+      newValue.saveName,
+      newValue.projectDir,
+    );
+    if (existingContentRecord) {
+      const crIndex = this.contentRecords.findIndex(
+        (contentRecord) => contentRecord.id === existingContentRecord?.id,
```
dotNomad, line 179: Should we be matching on `id` here? If we are updating a pre-deployment won't this check not give us any index and push a new content record to the array?

Actually since we are using this on the failure only we are nearly guaranteed to have an `id`, however since the method is for general use in the state it feels like it should be matching on the other unique information - `saveName` and `projectDir` like the `findContentRecord` does.

dotNomad (commented):
This works great to remove the flicker of the prior status.

I am however still getting the "This build was superseded by a more recent deployment" message though if I quickly re-deploy after hitting cancel. It looks like that occurs on the newer deployment. Is that still appearing from the original deploy, or is it possible to get that on the new one? The code looks like it should be suppressed if it this occurs on the cancelled deploy.

extensions/vscode/src/events.ts
sagerb: I've left it in, because I come back and need full message tracing every month or so. Let me add a comment that indicates to un-comment this if full message tracing is needed.

extensions/vscode/src/state.ts
sagerb: Good point. I'll update it to use the same criteria as `findContentRecord` does or possibly even simply call it.

sagerb:
> I am however still getting the "This build was superseded by a more recent deployment" message though if I quickly re-deploy after hitting cancel. It looks like that occurs on the newer deployment. Is that still appearing from the original deploy, or is it possible to get that on the new one? The code looks like it should be suppressed if it this occurs on the cancelled deploy.

I'll try reproduction of that once again, but we may have to debug on your mac. I haven't seen this again since my changes.

extensions/vscode/src/events.ts
sagerb: first one which is defined, I think. We'll confirm as we debug the issue on your machine. I've changed the code to use your suggestion (it was a good one).

sagerb:
Changes have been pushed up @dotNomad.

extensions/vscode/src/events.ts
dotNomad: Oops looks like your change left the debug uncommented 😅

dotNomad:
> Able to reproduce this by deploying to dogfood instead of a local connect server. I confirmed the message is coming from the connect server and the correct deployment task.
> 
> I've created a Connect issue: [posit-dev/connect#29511](https://github.com/posit-dev/connect/issues/29511), to ask for a rewording of this message.

Thank you for tracking that down, and creating the issue. That is super helpful.

dotNomad (approved):
Lets get that debug for the events commented up again, and removal the optional chaining this is looking great.

extensions/vscode/src/events.ts
sagerb: Great catch. Fixed!

extensions/vscode/src/state.ts
```diff
@@ -167,6 +167,27 @@ export class PublisherState implements Disposable {
...
+  ) {
+    const existingContentRecord = this.findContentRecord(
+      newValue.saveName,
+      newValue.projectDir,
+    );
+    if (existingContentRecord) {
+      const crIndex = this.contentRecords.findIndex(
+        (contentRecord) =>
+          contentRecord.deploymentPath ===
+          existingContentRecord?.deploymentPath,
```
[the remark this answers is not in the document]
sagerb: I'll make the change. Doing it independently from your commit suggestion, just to keep from needing to do a merge with my local changes.

PR #2536, opened 2025-01-14
Disable vitest thresholds `autoUpdate`

dotNomad:
This PR sets the `coverage.thresholds.autoUpdate` feature of `vitest` to `false` to avoid off by `0.01` coverage reporting errors.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling

### DIFF
--- a/.github/workflows/home-view-unit-test.yaml
+++ b/.github/workflows/home-view-unit-test.yaml
@@ -15,5 +15,3 @@ jobs:
           cache-dependency-path: "**/package-lock.json"
       - run: npm install
       - run: npm test
-      - name: Check Vitest thresholds
-        run: git diff --name-only --exit-code vite.config.ts
--- a/extensions/vscode/webviews/homeView/vite.config.ts
+++ b/extensions/vscode/webviews/homeView/vite.config.ts
@@ -40,11 +40,12 @@ export default defineConfig({
     coverage: {
       enabled: true,
       thresholds: {
-        functions: 30.13,
-        lines: 17.46,
-        branches: 44.82,
-        statements: 17.46,
-        autoUpdate: true,
+        functions: 30,
+        lines: 17,
+        branches: 44,
+        statements: 17,
+        // avoid auto-updating thresholds to avoid off by 0.01 differences
+        autoUpdate: false,
       },
     },
   },


sagerb (approved):
I'm good with disabling this and investigating the CI failure where CI's coverage % does not equal local calculations where the auto-update is occurring.

https://github.com/posit-dev/publisher/actions/runs/12778026743/job/35620705238?pr=2524

extensions/vscode/webviews/homeView/vite.config.ts
```diff
@@ -39,13 +39,6 @@ export default defineConfig({
     environment: "jsdom",
     coverage: {
       enabled: true,
-      thresholds: {
-        functions: 30.13,
-        lines: 17.46,
-        branches: 44.82,
-        statements: 17.46,
-        autoUpdate: true,
-      },
```
jonkeane, line 48:
```
      thresholds: {
        functions: 30,
        lines: 17,
        branches: 44,
        statements: 17,
        autoUpdate: false,
      },
```

.github/workflows/home-view-unit-test.yaml
```diff
@@ -15,5 +15,3 @@ jobs:
           cache-dependency-path: "**/package-lock.json"
       - run: npm install
       - run: npm test
-      - name: Check Vitest thresholds
-        run: git diff --name-only --exit-code vite.config.ts
```
jonkeane, line 19: FWIW I don't think this step is the issue, it fails on the one above if I'm reading correctly

sagerb: We encountered a failure earlier at this step, as the test step outputs a file w/ autoupdate, and then this step makes sure it is the same (and not showing as different).

dotNomad: > FWIW I don't think this step is the issue, it fails on the one above if I'm reading correctly

Correct, removing the thresholds from the config removes the check against the thresholds preventing the failure. We no longer need this if we don't need to use the `coverage.thresholds.autoUpdate` feature to ensure we are keeping the config up-to-date.

extensions/vscode/webviews/homeView/vite.config.ts
sagerb: That earlier failure I referenced above occurred when I attempted to do this myself, with autoupdate enabled:

https://github.com/posit-dev/publisher/actions/runs/12777773860/job/35619266385?pr=2524

dotNomad: `autoUpdate` has been set to false, thanks for the suggestion.

.github/workflows/home-view-unit-test.yaml
dotNomad: This has been removed since we are no longer using `autoUpdate` and we don't need to keep `vite.config.ts` up-to-date with the coverage threshold results.

sagerb:
Change looks great!

PR #2538, opened 2025-01-16
Remove localid from deployment record

sagerb:
## Intent

This PR removes the need to include the local_id within the deployment record and reduces the number of parameters needed to invoke the deployment writeFile method.

It also refactors "Aborted_At" to become "Dismissed_At." This affects the schema, variables, and logs.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

To enable removing the local_id out of the deployment file, I created an in-memory registry, updated the deployment's writeFile method to consult it, and refactored the existing calls to eliminate the need to have an override boolean. 

## User Impact

Deployment records that were used on this branch for prior iterations will not be valid. This will not affect users outside of Posit.

## Automated Tests

Automated tests were updated and added.

## Checklist


extensions/vscode/src/api/types/contentRecords.ts
```diff
@@ -45,7 +45,7 @@ type ContentRecordRecord = {
   saveName: string;
   createdAt: string;
   localId: string;
```
dotNomad, line 47: The `localId` should also be removed. I'm not seeing it in the response any longer with the removal.

extensions/vscode/src/test/unit-test-utils/factories.ts
```diff
@@ -44,7 +44,7 @@ export const preContentRecordFactory = Factory.define<PreContentRecord>(
     saveName: `Report ${sequence}`,
     createdAt: new Date().toISOString(),
     localId: "",
```
dotNomad, line 46: Similar note about the removal of `localId`

extensions/vscode/src/test/unit-test-utils/factories.ts
```diff
@@ -70,7 +70,7 @@ export const contentRecordFactory = Factory.define<ContentRecord>(
...
```
dotNomad, line 72: Similar note about the removal of `localId`

internal/schema/schemas/posit-publishing-record-schema-v3.json
```diff
@@ -70,14 +70,10 @@
...
-    "local_id": {
-      "type": "string",
-      "description": "ID of active/latest deployment sequence."
-    },
-    "aborted_at": {
+    "dismissed_at": {
       "type": "string",
       "format": "date-time",
-      "description": "Date and time that the current deployment flow was aborted. Is empty if flow was not aborted.",
+      "description": "Date and time that the current deployment flow was dismissed. Is empty if flow was not dismissed.",
```
dotNomad, line 76: I should have suggested this on the original wording. This makes a bit more sense to me, but feel free to ignore.

```suggestion
      "description": "Date and time that the deployment process was dismissed. Is empty if the deployment process was not dismissed.",
```

If changed here the draft should also have this change.

internal/services/api/post_deployment.go
```diff
@@ -124,6 +126,12 @@ func PostDeploymentHandlerFunc(
 		}
 
 		go func() {
+			// take ownership of this deployment when we start, release when we are done.
+			// release will first check to make sure we still have ownership before releasing
+			// someone else's ownership
+			deployment.ActiveDeploymentRegistry.Set(newState.SaveName, string(newState.LocalID))
```
dotNomad, line 132: I wasn't getting _any_ file writes on my machine when testing this locally. I did some debugging and `newState.SaveName` was an empty string: `''`

I am getting the `path.String()` correctly in the `WriteFile` method, but is this the correct thing we should be setting here?

dotNomad (commented):
Big fan of the `dismissedAt` rename to keep everything uniform.

extensions/vscode/src/api/types/contentRecords.ts
sagerb: Good point, thanks for finding these!

extensions/vscode/src/test/unit-test-utils/factories.ts
sagerb: thanks again.

extensions/vscode/src/test/unit-test-utils/factories.ts
sagerb: thanks again.

internal/schema/schemas/posit-publishing-record-schema-v3.json
sagerb: Changed both to be your recommendation. Thanks.

internal/services/api/post_deployment.go
sagerb: Fixed now in latest commits.

sagerb:
>Big fan of the dismissedAt rename to keep everything uniform.

Glad you liked it!

sagerb:
@dotNomad The latest commit (https://github.com/posit-dev/publisher/pull/2538/commits/f51a6027077e1d0d74b36282a603d02d4bb09914) addresses both the issue you were seeing with the status being lazily updated with success after dismissal, as well as one in which I saw a double update of successful -> dismissal.

The cause for your issue was simple enough, the bundle update was specifying an override to update the file even though it shouldn't have. This was first resolved by changing the override parameter, but then subsequently was eliminated by my slight rework of allowing an override on that method.

To solve the issue I was seeing, in which we were seeing two file refreshes come through consisting of our first update to a reset deployment file at the beginning of the deployment process (which showed as success in the web UX) and then a second update as the dismissal was written to the deployment record (which showed as dismissed). This switch was visible because we were also forcing a preemptive update to the deployment record in memory (within the webview store), which was then a delta to showing the reset object, and then another delta to show the dismissed object. This was visible because we had also inserted an event to indicate the deployment was completed, so our "deployment in progress" UX element was hidden, showing the status out of the deployment file. I solved this by eliminating the initial update/reset of the deployment file at the start of the deployment. (I had added this when I had gone the route of adding the localID into the deployment file, but no longer needed it when we removed it.) This eliminated the two updates (which were asynchronous in the background), eliminating the underlying stimulus for the update. 

Please take a look at my changes and please verify the functionality. Reach out if you see anything weird. Thanks!

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
```diff
@@ -410,7 +410,7 @@ const lastStatusDescription = computed(() => {
       ? "Not Yet Updated"
       : "Not Yet Deployed";
   }
-  if (isAbortedContentRecord.value) {
+  if (isDismissedContentRecord.value) {
```
dotNomad, line 413: Should we move this check above the `isPreContentRecord` check in this method? To avoid the situation where you cancel a deploy prior to getting an ID and the status showing "Not Yet Deployed" even though it has been dismissed.

internal/deployment/deployment_owner_registry.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (o *DeploymentOwnerRegistry) Clear(deploymentPath string, localId string) {
+	o.Lock()
+	defer o.Unlock()
+
+	owner := o.m[deploymentPath]
+	if owner == localId {
+		delete(o.m, deploymentPath)
+	}
```
dotNomad, line 39: Nitpick: do we need to check if `o.m[deploymentPath]` exists?

> // The delete built-in function deletes the element with the specified key
// (m[key]) from the map. If m is nil or there is no such element, delete
// is a no-op.

dotNomad (approved):
I pulled this down and gave it some extensive testing. I only hit 2 things worthy of a comment. One was a nitpick on code simplicity, and the other is a relatively minor UI change request that we can discuss.

Thank you for taking the time to really iron this feature out.

extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
sagerb: Yes, in fact, it should go above the `deploymentError` check. This pointed out that we should be resetting the error when we set the `DismissedAt` attribute since we are not resetting the entire deployment object at the beginning of the deployment.

internal/deployment/deployment_owner_registry.go
sagerb: No, it was designed so that there must be a match of what localID the caller thinks is registered at the deploymentPath key, otherwise, it won't be deleted. This is to protect the staleness of information that parallel threads may have.

dotNomad: > it was designed so that there must be a match of what localID the caller thinks is registered at the deploymentPath key, otherwise, it won't be deleted

Ah that makes sense now. Thanks for the explanation.

dotNomad (approved):
Fantastic changes 💯

PR #2545, opened 2025-01-22
Include connectivity to server check in invalid server or credentials error

### LINKED ISSUE #2450: Error when the Connect server is down is not descriptive
In this case I got `502 Bad Gateway` when going to the server directly, but the error recommends checking the URL and API key.

Image

dotNomad:
This PR changes the `errInvalidServerOrCredentials` to be more board to include the case where the Connect server is down or cannot be connected to. Now rather than only suggesting to check the provided server URL and API key it now also suggested checking "connectivity to the server".

## Intent

Resolves #2450

### DIFF
--- a/internal/clients/connect/client_connect.go
+++ b/internal/clients/connect/client_connect.go
@@ -86,7 +86,7 @@ func (u *UserDTO) CanPublish() bool {
 	return u.UserRole == AuthRoleAdmin || u.UserRole == AuthRolePublisher
 }
 
-var errInvalidServerOrCredentials = errors.New("could not validate credentials; check server URL and API key or token")
+var errInvalidServerOrCredentials = errors.New("could not validate credentials; check connectivity to the server, the URL, and API key")
 var errInvalidApiKey = errors.New("could not log in with the provided credentials")
 var errInvalidServer = errors.New("could not access the server; check the server URL and try again")
 


internal/clients/connect/client_connect.go
```diff
@@ -86,7 +86,7 @@ func (u *UserDTO) CanPublish() bool {
 	return u.UserRole == AuthRoleAdmin || u.UserRole == AuthRolePublisher
 }
 
-var errInvalidServerOrCredentials = errors.New("could not validate credentials; check server URL and API key or token")
+var errInvalidServerOrCredentials = errors.New("could not validate credentials; check connectivity to the server, the URL, and API key")
 var errInvalidApiKey = errors.New("could not log in with the provided credentials")
 var errInvalidServer = errors.New("could not access the server; check the server URL and try again")
```
sagerb, line 91: With a Server URL connectivity issue when creating credentials, I encountered this error. Did we not want to also update this message to check connectivity?

2025-01-22 at 8 31 AM

marcosnav (approved):

internal/clients/connect/client_connect.go
sagerb: Should we also update the second and third errors to show the server URL with which we're trying to connect?

2025-01-22 at 8 33 AM

sagerb (changes requested):
A few quick improvements requested.

internal/clients/connect/client_connect.go
dotNomad: > Did we not want to also update this message to check connectivity?

For this error it does include "could not access the server" which to me covers the case where the server is down. Were you thinking something different with the error?

dotNomad: > we also update the second and third errors to show the server URL with which we're trying to connect

Since those two are used during or right after inputing the URL I don't think it is necessary to repeat it like that. I could be more convinced on the second, but then I worry about error length.

sagerb: > For this error it does include "could not access the server" which to me covers the case where the server is down. Were you thinking something different with the error?

Fair enough.

>Since those two are used during or right after inputting the URL, I don't think it is necessary to repeat it like that. I could be more convinced on the second, but then I worry about error length.

You are correct in that connectivity to the URL has been verified when we get to this step. So when we fail here, the API key may no longer be valid or our validation of the server URL has issues. I think it would be useful to have the server URL present if I were trying to track it down (I think instant visual validation might help), and I'm more interested in providing the information to the user rather than the brevity of the error message. But I'm not strongly tied to this, I'll leave it to you to weigh the benefits and costs.

sagerb (approved):
I'm good with whatever you decide regarding my concerns. Thanks for considering them.

marcosnav:
Verified! Marking issue as Done
Screen Shot 2025-01-27 at 15 14 33

Issue #2542, opened 2025-01-21
Providing a `type` that isn't mapped in `connectContentTypeMap` sends empty `appMode`


dotNomad:
As we learned in https://github.com/posit-dev/publisher/issues/2476#issuecomment-2605451591 when providing a `type` in the Configuration file that isn't in `connectContentTypeMap` the manifest sent to Connect [provides `""`](https://github.com/posit-dev/publisher/blob/0b43655aba2281ba17aee425ef39b59a98abb330/internal/clients/connect/apptypes.go#L281) (as it is the `UnknownMode`).

In an effort to better support Publisher working when Connect updates without code changes the `appMode` should be sent using the provided `type` string in the Configuration file rather than falling back to `""` if it is unrecognized.

Reference:

https://github.com/posit-dev/publisher/blob/0b43655aba2281ba17aee425ef39b59a98abb330/internal/clients/connect/apptypes.go#L258

PR #2546, opened 2025-01-22
Add troubleshooting link to documentation when `publish/failure` occurs

### LINKED ISSUE #2476: Handle deploying to a Connect server without Gradio support

dotNomad:
This PR adds:
- a new Troubleshooting document linked from our `docs/index.md`
  - The start of this document includes information for how to address the "Unrecognized app mode" error during deployment.
- a change to the `window.showErrorMessage` call on `publish/failure` to include a new "See **Troubleshooting docs** for help."
  - The **Troubleshooting docs** text will link to the new troubleshooting document.

<details>
  <summary>Preview</summary>
  
CleanShot 2025-01-21 at 18 42 50

</details> 

## Intent

Resolves #2476

## Type of Change

- - [x] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [x] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I only included the troubleshooting link at our `window.showErrorMessage` call from `publish/failure` as opposed to always showing the Troubleshooting link because the only content in our troubleshooting doc has to do with publish failure.

As we add more to the document we can easily replace `window.showErrorMessage` usage with usage of the new `showErrorMessageWithTroubleshoot` helper I created here.

I can change this based on feedback on this PR.

## User Impact

Users will now see a handy **Troubleshoot docs** link when deploying to a Connect server without Gradio support - and really for any `publish/failure` error message they receive.

## Directions for Reviewers

The easiest way I found to test this is to change our two schemas to support a type that is not supported like `"nonsense"`.

Then use that in a configuration file and attempt to deploy - leading to the error notification above. Note that the app mode isn't listed. That is due to https://github.com/posit-dev/publisher/issues/2542

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -2,7 +2,7 @@
 
 This document contains some common issues and solutions for Posit Publisher.
 
-### The Configuration has a schema error
+## The Configuration has a schema error
 
 If deploying results in the following error:
 
@@ -19,7 +19,7 @@ Carefully check your Configuration file for any errors, and refer to the
 [Configuration File reference documentation](https://github.com/posit-dev/publisher/blob/main/docs/configuration.md)
 for more details on the fields and values that are expected.
 
-### Unrecognized app mode
+## Unrecognized app mode
 
 If when deploying you see the following error:
 
@@ -27,7 +27,7 @@ If when deploying you see the following error:
 
 there are few things to check.
 
-#### Using `type = 'unknown'`
+### Using `type = 'unknown'`
 
 If your Configuration file contains `type = 'unknown'`, either from the
 generated Configuration being unable to identify the type of your content, or
@@ -40,7 +40,7 @@ Configuration.
 To fix this you will need to find and set the `type` in your Configuration file.
 Supported types can be found in the [Configuration File Reference documentation](https://github.com/posit-dev/publisher/blob/main/docs/configuration.md#type).
 
-#### Using a `type` that is not supported by your server
+### Using a `type` that is not supported by your server
 
 Another possibility is Posit Connect, on the server you are deploying to, is
 older than the version that introduced support for the `type` set in your
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -2,20 +2,53 @@
 
 This document contains some common issues and solutions for Posit Publisher.
 
-### Unrecongized app mode
+### The Configuration has a schema error
+
+If deploying results in the following error:
+
+> Failed to deploy. The Configuration has a schema error
+
+This occurs when there is a problem with the Deployment's Configuration file.
+This can occur due to a missing field, an incorrect value, or even a typo.
+
+Using an extension like [Even Better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml)
+can help catch problems in the Configuration by providing syntax highlighting,
+completion, and hover text using the schema provided by Posit Publisher.
+
+Carefully check your Configuration file for any errors, and refer to the
+[Configuration File reference documentation](https://github.com/posit-dev/publisher/blob/main/docs/configuration.md)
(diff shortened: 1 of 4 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Unrecongized app mode
+
+If when deploying you see the following error:
+
+> Cannot process manifest: Unrecognized app mode
+
+It is likely due to Posit Connect, on the server you are deploying to, being
+older than the version that introduced support for the `type` set in your
+configuration file.
```
jonkeane, line 13: This section is good, but we should have a section above it with other common type mismatches. This is definitely one possible one — and folks who use our inspection is only one of two possible situations they could get themselves into, but there is also the possibility that someone edited the config file to put some other type that's unknown.

I would structure it like this:

> There's an issue with the type ...
> 
> ## Unknown / unset `type`
> 
> In this scenario you will need to find the type and set it...
> 
> ## Using a `type` that isn't supported by your server
> 
> Connect might be out of date ...
> 
> ## A typo in `type`
> 
> If you hand edited the configuration file, you might have had a typo in the type field. Check that.

docs/troubleshooting.md
```diff
@@ -0,0 +1,28 @@
+# Troubleshooting Posit Publisher
+
+This document contains some common issues and solutions for Posit Publisher.
+
+### Unrecongized app mode
```
marcosnav, line 5:
```suggestion
### Unrecognized app mode
```

extensions/vscode/src/utils/window.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2025 by Posit Software, PBC.
+
+import { window } from "vscode";
+
+export function showErrorMessageWithTroubleshoot(
```
marcosnav, line 5: This is a nice approach! wrapping the vscode usage here helps on mocking this module where `showErrorMessageWithTroubleshoot` is used for tests.

We could definitely do the same for other messaging methods we have on the code base.

sagerb (changes requested):
Jon's recommendations make sense to me.

docs/troubleshooting.md
dotNomad: Oops, thanks for catching this.

docs/troubleshooting.md
dotNomad: > This section is good, but we should have a section above it with other common type mismatches.

Thank you, and fantastic idea. I added two more sections in https://github.com/posit-dev/publisher/pull/2546/commits/0a100d652d73ffa12f6b7612da9cb34099eecddd

I kept the organization of

```
### Errors

> Exact error text

#### Solutions
```

let me know if you think that is working well.

The two sections I added capture:
- The "Failed to deploy. The Configuration has a schema error" which occurs when there is a typo, missing section (like no `[python]` section when the Configuration is set to `type = "python-fastapi"` for example), or incorrect values
- A new solution section under "Unrecognized app mode" when using `type = 'unknown'`

docs/troubleshooting.md
```diff
@@ -0,0 +1,61 @@
+# Troubleshooting Posit Publisher
+
+This document contains some common issues and solutions for Posit Publisher.
+
+### The Configuration has a schema error
```
jonkeane, line 5: Minor nit, but do we want to go from level 1 to level 3 headings like this? Or should this (and the ones below) be level 2?

jonkeane (approved):
This looks great, thanks for this!

docs/troubleshooting.md
dotNomad: Good callout fixed in [e9be268](https://github.com/posit-dev/publisher/pull/2546/commits/e9be2685cf2d3e1e20824a63170db274b2fe65d9)

sagerb (approved):
Changes look great to me.

PR #2549, opened 2025-01-22
Run npm upgrade in homeView package

dotNomad:
Simple PR that runs `npm upgrade` in the homeView package. This supersedes #2547 to capture the upgrade of `vite`


sagerb (approved):
LGTM

PR #2552, opened 2025-01-23
Allow empty string for `dismissed_at` in schema

dotNomad:
I noticed that with the changes in #2522 that deployment files were showing schema errors through Even Better TOML. I didn't get errors when using Publisher, but the JSON schema was saying that `""` wasn't a `date-time`.

This allows the `dismissed_at` attribute be either an empty string or a `date-time`.

CleanShot 2025-01-22 at 16 32 57

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling


sagerb (approved):
Thanks for picking this up!

dotNomad:
Schema has been sent up to S3 🚀

Issue #1503, opened 2024-05-02
Don't allow `requirements.txt` and `renv.lock` files to be excluded from projects


kgartland-rstudio:
You can exclude the `requirements.txt` and `renv.lock` files in the Project Files section, subsequent deployments fail when trying to read such files in Connect.

---

- [ ] ~Add a pre-flight check to ensure the package file is included~ Superseded by a separate issue #2566
- [ ] Add a warning in the sidebar when the package file is not included in the selected configuration
- [ ] Disallow the Files API from removing the package files
- [x] Disable the checkboxes in the Project Files view (done in #2555)

dotNomad:
There is a pre-flight check already for the inclusion of the `Python.PackageFile`: https://github.com/posit-dev/publisher/blob/a7429cb6124bc786dfabd1d6faa8a3cac7c9b8a7/internal/clients/connect/capabilities.go#L69

The inclusion of the `R.PackageFile` fails earlier when attempting to get the R packages for the manifest.

dotNomad:
Using the same method for checking for package file inclusion for R is not recommended as it has an issue with wildcard inclusions #2563

PR #2555, opened 2025-01-25
Disable `ProjectFile` checkbox when it is a package file

dotNomad:
This PR disables the checkbox for R and Python package files in the Project Files view, as well as includes a new tooltip when this is the case. It works very similar to #2487 

## Intent

Part of #1503

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## User Impact

Users can no longer un-check package files' checkboxes in the Project Files view - they can still check them to include the file if they are not included.

## Automated Tests

New automated tests are included for the new tooltip.

## Directions for Reviewers

Ensure that the R and Python package files, no matter the details in the configuration, can be unchecked.

Checking the checkbox for the file should still be available if the configuration is manually edited and the package file needs to be added.

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/ProjectFile.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/ProjectFile.vue
@@ -4,7 +4,7 @@
     :checked="isIncluded"
     :disabled="isDisabled"
     :list-style="listStyle"
-    :disable-opacity="isEntrypoint"
+    :disable-opacity="isEntrypoint || isPackageFile"
     :indent-level="file.indent + 1"
     :expandable="file.isDir"
     :tooltip="tooltip"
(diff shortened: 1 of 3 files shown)


sagerb (approved):
LGTM. I verified the functionality. 

There was some strangeness when I opened some older deployments with configurations where the package file was not included. I had one circumstance in which the tree had the file not checked but disabled, however the file list included the `renv.lock` file. I think this may have been a refresh issue, and once I reloaded the window, everything worked fine. I did not have this issue on any other deployments or configurations that I checked.

Outside of that one occurrance, I was able to check the package file when it was not included in the files list, and it was added. I was then unable to remove it as designed by the disabled click.

Looks good to merge!

PR #2561, opened 2025-01-28
Only show the view content button with a valid link

### LINKED ISSUE #2373: If a deployment fails before we can create anything on Connect, we still display an active `View Content` button
If I try to deploy content to Connect but it fails before anything is created, under the failure message there's still a View Content button but it leads nowhere.

Here I was able to add the dogfood server using `http://` rather than `https://` and deploy but that deployment didn't result in content being created.  However, we still show the View Content button which doesn't lead anywhere.

Image

dotNomad:
This PR changes the behavior around the "View Content" button to only show it if we have a URL to go to. Previously we checked if the deployment wasn't a pre-deployment, but a failed deployment will not have a valid ID or links.

## Intent

Resolves #2373

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## User Impact

If an error occurs during the deployment that causes the deployment to no longer be a pre-deployment, but still doesn't have an ID the "View Content" button no longer is shown.

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
+++ b/extensions/vscode/webviews/homeView/src/components/EvenEasierDeploy.vue
@@ -221,10 +221,7 @@
             />
           </div>
         </template>
-        <div
-          v-if="!isPreContentRecord(home.selectedContentRecord)"
-          class="last-deployment-details"
-        >
+        <div v-if="showContentButton" class="last-deployment-details">
           <vscode-button
             appearance="secondary"
             @click="viewContent"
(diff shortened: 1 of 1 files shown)


marcosnav (approved):
🎉

marcosnav:
## Verification screenshots

**Previously**
Screen Shot 2025-01-28 at 16 11 20

**Now - no button**
Screen Shot 2025-01-28 at 16 07 50

PR #2568, opened 2025-01-29
Support optional python and r section attributes

### LINKED ISSUE #2468: Update manifest.json R version population
We should update how we decide which R version we specify for running on Connect. This work would update the the `R.version` field in the config to be optional, and document what it means to set it. This should also update how we build the `manifest.json` accordingly:

1. From the `R.version` in the config, if set (this field is now optional)
2. From the current active version of R in the IDE, if the `R.version` is not set in config

The answer to this should populate the `platform` field in the `manifest.json`, when we build it.

Note that this doesn't lean on the R version defined in `renv.lock`, which, ultimately, we may want to do. If we tackle that change, it will be in a follow-on issue.

-----

After analyzing feedback from the original PR that was created to address this issue (https://github.com/posit-dev/publisher/pull/2568), we've decided that the alternative design which was informally discussed would be a better direction to go in.

With the implementation of default values, for R and Python, derived from the active interpreters at the time of deployment, the challenge has been how to keep the concept of a default wanted vs. a locked-in (supplied) value within the extension UX, which depends on values for these fields.

The attributes / responsibility of this alternative design would involve maintaining the internal representation of the configuration objects as they are represented on disk, with the addition of default value fields. The deployment process (in the agent) would be responsible for resolving the defaults, at the time needed when generating the maniftest. The extension UX would be responsible for converting the configuration objects queried from the agent w/ the active "expected" default values. 

Current thoughts on this path: the configuration object can can be extended to include default values inserted during transit. This would maintain being able to identify the difference, but would make the users of this field responsible for looking possibly at both values. (However, this can be done within the existing store, with an explicit computed value.)

### LINKED ISSUE #2470: Update manifest.json Python version (platform) population

sagerb:
## Intent

The changes within this PR support the initialization of config files to have default attributes associated with R and Python dependencies, allowing the sections to be empty within the config file. When values are present, they are used by the back end, but when they are not present, the current values from the active interpreters are used, as the deployment manifest is built from the config file. 

Resolves #2468 
Resolves #2470

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

To implement this functionality, it is important to have the config file written out without the attributes, but when it is passed around internally (and retrieved by the UX), that it have the current defaults inserted into the config object. Since the UX only patches the config file, it is important that the handling code really does "patch" rather than replace the object to disk (which would write out the defaults, which then become user-provided values).

To support the insertion of the defaults internally, some of the APIs needed to be updated to pass in the active interpreters. The insertion has been isolated into one specific spot; the point at which the code builds the config object from the file.

Since so many locations in the codebase are now using the interpreters, I've also modified the method in which they are initialized. Rather than pass around the executable string, the particular interpreter is built from that executable string, and then passed around to be used where needed. This is important to be done at the API level, so that when a user changes their selection or settings for the active interpreters are picked up at the next API call.

## User Impact

New feature: The scenario goes like this:
1. A publisher has created a deployment that uses their currently active R and/or Python interpreters as the versions requested in the manifest.
2. The next day, they upgrade or change their active interpreters and then deploy again. These re-deployments once again request the active R and/or Python interpreters within the manifest.
3. If at any point the user sets the attribute values within the R or Python section, they will be used rather than floating to the active interpreter values.

## Automated Tests

Unit tests updated.

## Directions for Reviewers

Best to review this PR via commits. The list of commits have been updated to describe the changes.

- [Update Schema to make attributes inside of R and Python be optional.](https://github.com/posit-dev/publisher/commit/20554974c0b69a48a76c66b5b310b4ae47ede1fb)
- [Change to optionally populate R and Python sections during initial configuration initialization if told to. This will return new configs with empty sections (signaling to use defaults from interpreters), but still allow testing.](https://github.com/posit-dev/publisher/commit/154df3a42ce1d6e0e2595ec992f0c3565c32978f) 
- [extract active R and Python paths from API requests](https://github.com/posit-dev/publisher/commit/527c3e2c810da992445a119633e6849e1cf04aa7)
- [Omit R and Python version from config when empty](https://github.com/posit-dev/publisher/commit/ce821637c451e7af348777cc2d8ebf36acf5a10d)
- [Pass interpreters in versus creating interpreters within inspect](https://github.com/posit-dev/publisher/commit/e4a3f16b2edc8143658f8f8680db574c814b69da)
- [add logging to know what is being passed up in manifest](https://github.com/posit-dev/publisher/commit/430fe3f944c35b6b73b8bad15f06e3e07f528e74)
- [update initialization of publishing state to allow the loading of the config to fill in the missing pieces using the active interpreters (in config.FromFile)](https://github.com/posit-dev/publisher/commit/07f2844592a77ed03ffd02d215048c66b5e91d09) 
- [Update fromFile to fill in the empty fields into the config object, using the active interpreters](https://github.com/posit-dev/publisher/commit/fbe4a18cf3d963bc3e70452951be59354d0a9217) 
- [update unit tests for agent](https://github.com/posit-dev/publisher/commit/22a91732f4600ce9bc10f0d8d34f016ac69d1e32)
- [update unused CLI command handlers just to compile...](https://github.com/posit-dev/publisher/commit/629eae0b860d3623bcf741bd7d9138efd2c1c990)
- [Utilize a TS class to make sure we pass in the proper interpreter (R vs. Python), since they are only strings by nature and order in the call functions can get very messed up.](https://github.com/posit-dev/publisher/commit/dcaf22d3e4f41762de82b15cadc09d63cf0d4adb) 
- [Update the UX API signatures to include R & Python](https://github.com/posit-dev/publisher/commit/998d0c463108c0f1284b5b381529f3a47de4e7fa)

## Checklist


dotNomad (commented):
I'll note that we should definitely change the Configuration file reference document as part of this PR since this will make the Python settings and R settings sections out of date.

sagerb:
> I'll note that we should definitely change the Configuration file reference document as part of this PR since this will make the Python settings and R settings sections out of date.

I pushed up an update!

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -20,25 +22,41 @@ export class Configurations {
   // 200 - success
   // 404 - not found
   // 500 - internal server error
-  get(configName: string, dir: string) {
+  get(
+    configName: string,
+    dir: string,
+    r: RExecutable,
+    python: PythonExecutable,
+  ) {
```
marcosnav, line 30: Could you ellaborate more on why it is required to send an R and Python executables to get a configuration? Shouldn't the configuration response be what the existing file contains?

dotNomad: > but when it is passed around internally (and retrieved by the UX), that it have the current defaults inserted into the config object.

As a part of keeping this PR as narrow as possible we didn't change the behavior in the API and frontend.

Currently in `main` the defaults are inserted into the response even if the file doesn't have those attributes set (like the `packageFile = 'requirements.txt` settings). Rather than changing both the backend and frontend here we kept the API behavior the same.

marcosnav: I think I understand the reason of having the resulting config in the UI to have some defaults, but it feels confusing the way it is being handled.

This method is one of many that now accepts R and Python executables, but feels out of place, for example the `getPythonPackages` calling `configurations/{name}/packages/python` also accepts and `R` executable, and I find it confusing for future folks to have a method responsible to get python packages to accept an R executable.

The `go` method `FromFile` now accepts overriding defaults as `activeRInterpreter` and `activePythonInterpreter`, and we are now having a method that at first look suggests to load the configuration as it is on file, but consumers of it could find surprises from it.

My suggestion would be, to take a step back a bit and think how these signatures bring clarity or confusion to future folks working on this, specifically from the point that future folks might not have all the context on how this works or the history behind all this.

dotNomad: Great point. The effort here was to avoid making changes on the frontend, but creating confusion by needing to pass in the Python and R executables everywhere is confusing.

sagerb: >The go method FromFile now accepts overriding defaults as activeRInterpreter and activePythonInterpreter

To clarify, it's not overriding defaults but actually implementing them. The defaults come from the active interpreters, passed in at the time of the API request. Our current architecture requires all the levels of calls to pass these down into the deepest functions that need it. Unfortunately, that is very deep in this case.

I understand about the confusion. Let me think about this, and I'll signal for another review. Thanks.

sagerb:
I've thought about the approach used in this PR and I've concluded that we introduced confusion because we were attempting to not modify the extension UX.

I'm going to update the related issue with our learnings, so that we can implement something else here.

I'll be closing this PR without merging.

PR #2576, opened 2025-01-30
Add ability to update the target guid for a deployment at any time (from the ... menu)

### LINKED ISSUE #2400: Redeploying to an existing content target with wrong content type
When I redeploy to update a previous content item and there's a mismatching content type, I get the proper error but don't have a path to easily correct it as the `update previous deployment` option is overridden by the new error.

Scenario:
Grab a URL to a piece of content from the server that you want to target
Accidentally chose the wrong URL
Got the proper failure 🎉 
Now I get the right URL and want to deploy to that content, but I can't now since the `update previous deployment` link is gone.

Screenshot 2024-10-16 at 12 46 54 PM

The work-around is to create another new deployment but that's not very clear.

sagerb:
…rom ... menu)

## Intent

Add the ability to associate an existing deployment with a different GUID. This simply offers the same capability that we had previously offered ahead of the first deployment within the UX. This option is now always available via the ... menu when a deployment has been selected.

Resolves #2400 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I started by simply exposing existing functionality into the context menu and adding a success notification.

I then realized that the menu item doesn't make any sense if there is no credential match, so I've added and updated a context within the extension that can be used to show or hide the menu item.

New screenshots:

When there are credentials to the server and the active selection has been deployed:
image

When there are no credentials to the server:
image

When the active selection has not yet been deployed:
image

It will show when other errors w/ the config are present:
image

I also added an informational message when it has completed calling the patch API.
2025-01-30 at 10 32 AM

## User Impact

New feature!


extensions/vscode/src/actions/showAssociateGUID.ts
```diff
@@ -44,12 +44,17 @@ export async function showAssociateGUID(state: PublisherState) {
...
+    const result = await api.contentRecords.patch(
       targetContentRecord.deploymentName,
       targetContentRecord.projectDir,
       {
         guid,
       },
     );
+    if (result.status === 200) {
+      window.showInformationMessage(
+        `Deployment is now associated with Content GUID ${guid} as requested. An error will be displayed if this deployment cannot update this deployment on the Connect Server.`,
```
dotNomad, line 56: I don't understand the second half of this message? What are you trying to communicate here?

extensions/vscode/src/actions/showAssociateGUID.ts
```diff
@@ -44,12 +44,17 @@ export async function showAssociateGUID(state: PublisherState) {
...
     const api = await useApi();
-    await api.contentRecords.patch(
+    const result = await api.contentRecords.patch(
       targetContentRecord.deploymentName,
       targetContentRecord.projectDir,
       {
         guid,
       },
     );
+    if (result.status === 200) {
```
dotNomad, line 54: Do we need to check the status, or can we surround this in a try catch?

extensions/vscode/package.json
```diff
@@ -135,6 +135,11 @@
         "title": "Select Active Configuration For Deployment",
         "category": "Posit Publisher"
       },
+      {
+        "command": "posit.publisher.homeView.showAssociateDeployment",
```
dotNomad, line 139: Nit to use verb language for the command, feel free to ignore.

```suggestion
        "command": "posit.publisher.homeView.associateDeployment",
```

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -179,6 +179,18 @@ export const useHomeStore = defineStore("home", () => {
...
+  watch([serverCredential], () => updateSelectionCredentialStatus());
+
+  const updateSelectionCredentialStatus = () => {
+    const hostConduit = useHostConduitService();
+    hostConduit.sendMsg({
+      kind: WebviewToHostMessageType.UPDATE_SELECTION_CREDENTIAL_STATE,
+      content: {
+        state: serverCredential !== undefined ? "true" : "false",
+      },
+    });
```
dotNomad, line 191: This is always sending back `true`. Looks like it might be a missing `.value` on `serverCredential`.

dotNomad: I don't understand the purpose of the message sending however. This is checking if the current deployment has a credential, but if we are associating with a different piece of content we could be pointing to an entirely different server. What is this supposed to be catching?

extensions/vscode/package.json
```diff
@@ -325,6 +330,10 @@
           "command": "posit.publisher.homeView.showSelectConfigForDeployment",
           "when": "webviewId == 'posit.publisher.homeView' && webviewSection == 'even-easier-deploy-more-menu-matching-configs'"
         },
+        {
+          "command": "posit.publisher.homeView.showAssociateDeployment",
+          "when": "webviewId == 'posit.publisher.homeView' && webviewSection == 'even-easier-deploy-more-menu-matching-configs' && posit.publish.selection.haveCredentialMatch == 'true'"
```
dotNomad, line 335: This shows the context menu option while a pre-deployment shows the association link.

We could try to hide it using message passing or we could generalize the label to remove the "with a **Different** Deployment". 🤔

extensions/vscode/src/actions/showAssociateGUID.ts
sagerb: Trying to describe that we aren't error checking content type match during this operation. It will be checked at deployment.

extensions/vscode/webviews/homeView/src/stores/home.ts
sagerb: The purpose is to enable or disable the ability to associate, within the ... menu. I've seen it work, so I'm surprised there is an error there, but I'll check it out. What it implements is that when we do not have a credential for the server, so we couldn't contact the server to patch the deployment file, so it would fail always. So the option isn't offered.

I'll check the logic, like I mentioned, I'm not sure with this how I was seeing the vscode context set correctly per the state, as shown in the screen shots.

extensions/vscode/package.json
sagerb: I'm not following you. Can you provide a screenshot? I'm looking at
image

extensions/vscode/src/actions/showAssociateGUID.ts
sagerb: Yes we should surround it. Thanks.

extensions/vscode/package.json
sagerb: sounds like a good suggestion. Thanks.

extensions/vscode/package.json
dotNomad: Yeah that screenshot shows what I'm talking about. We have both "update that previous deployment" available as a link, and the "Associate with a Different Deployment on Connect". The wording is just strange since it isn't associated with any deployment yet. That may be too in the weeds with how this is working though.

extensions/vscode/src/actions/showAssociateGUID.ts
sagerb: I've simplified the message.

extensions/vscode/src/actions/showAssociateGUID.ts
sagerb: There is a 204 that can also be returned for the request, which would not be thrown by the Axios status code handler, but that would only occur for a malformed request. I think we're good to switch this to a try/catch, so I've done so.

extensions/vscode/package.json
sagerb: Updated.

extensions/vscode/package.json
sagerb: I've added another context state within the extension, which indicates if the actively selected content record is a PreContentRecord or not. This allows the menu item to now be shown only when it has been deployed, which should address the confusion you identified here.

jonkeane:
I just `s/quid/guid/` on the title cause I thought that was a typo, but if it's not sorry!

extensions/vscode/package.json
dotNomad: Thanks for doing that, it is a small thing, but it should make it easier to understand for users.

marcosnav (approved):

PR #2579, opened 2025-01-31
Option to disable keychain credentials

### LINKED ISSUE #2519: Add configuration option to prioritize use of credential file over keyring
For debugging and testing, we need a way of testing the credential file usage on platforms that have working keyring support. The easiest/cleanest way to implement this is to add an extension configuration option. This would seem to be a good example of something that users might also want to use, so I'm recommending we expose it publically through VSCode's configuration facility.

sagerb:
## Intent

This PR introduces a configuration option to disable the use of the KeyRing for credential storage and instead use the file system. This is extremely useful for development as well as a good escape hatch for customers who may not have a working KeyRing environment (or preference there-towards).

Resolves #2519 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I first added a new configuration option for the publisher extension, which is enabled by default:
2025-01-31 at 11 23 AM

I then added a corresponding command line option for the agent. This was a more straightforward change than passing the option for every credential-related call. The impact of this decision is that the user must restart the extension after changing the configuration setting. Since this won't be very common, I believe this was a reasonable optimization to keep the complexity low.
```
$(just executable-path) ui -h
Usage: publisher ui [<path>] [flags]

Arguments:
  [<path>]    Sets the current working directory for the agent.

Flags:
  -h, --help                  Show context-sensitive help.
  -v, --verbose               Enable verbose logging. Use -vv or --verbose=2 for debug logging.

      --listen=HOST[:PORT]    Network address to listen on.
      --use-keychain          Use Keychain services to store/manage credentials.
```

After those two changes were in place, I added a helper function within the extension to access the setting, then updated the service spawning code to reflect the current configuration value.  The agent then stores that value and each time the agent's credential code initializes, it either bypasses or attempts to use the KeyChain dependency.

The agent's debug launch profile has been updated to support launching the agent with the command line set. I set the initial value to `false` as I found myself wanting to debug using the file-based credentials more often than the KeyChain alternative. If this is not desired by the developer, they can easily change it within their own config.

## User Impact

New feature, but default maintains the behavior they had previously.

## Automated Tests

Unit tests added to validate the desired behavior.

## Directions for Reviewers

Launching a debug version of the extension from your VSCode instance for the extension, will work for verification. The configuration value is respected and maintained after being set, when you then reload the window with `Developer: Reload Window`. I'd suggest verification of the functionality by:
1. Confirm the configuration setting for the extension is set to enabled (checked). If have to change it, then reload the VSCode window
2. Activate the publisher extension and view the credentials. You should be able to confirm that your existing credentials is listed.
3. Change the configuration setting to disabled and reload the window. 
4. Activate the publisher extension and view the credentials. You should either have none or whatever you had stored within the `~.connect-credentials` file.
5. Add a new credential and make sure it shows up in your credential list
6. Reload the window again and make sure the credentials have remained.
7. Change the configuration setting back to enabled and reload the window
8. You should be able to validate that the original KeyChain based credentials are now shown.

## Checklist


dotNomad (commented):
It looks like this was branched off of the CLI removal. Can we change the base, or rebase this to isolate just the keychain credential changes? That will probably get CI passing again.

sagerb:
> It looks like this was branched off of the CLI removal. Can we change the base, or rebase this to isolate just the keychain credential changes? That will probably get CI passing again.

Oh yes, sorry about that. I've rebased the diff and updated from that updated branch as well. Thanks!

dotNomad (approved):
This is very clean. I just had a suggestion to trim up the description of the setting, and make it more similar to other VS Code setting descriptions.

sagerb:
To get CI to pass, we have to merge #2584 first.

Issue #2510, opened 2025-01-03
How can I pass extra settings to manifest.json?


nealrichardson:
The [latest Connect release](https://docs.posit.co/connect/news/#posit-connect-2024.12.0-new) allows you to opt-in to using `uv` to install packages on Connect by adding an entry to the manifest.json. Is there a way I can specify that in the publisher TOML such that it will pass through to the manifest (which I assume is being written on the fly?)

Related question: what happens if I generate a manifest with `rsconnect-python` and then use Publisher to publish?

dotNomad:
> Is there a way I can specify that in the publisher TOML such that it will pass through to the manifest (which I assume is being written on the fly?)

You are correct that the manifest gets written on the fly when deploying. There isn't a way right now to write something in the config that gets passed plainly to the generated manifest without changes to the code.

> Related question: what happens if I generate a manifest with rsconnect-python and then use Publisher to publish?

Publisher always uses the manifest it writes from the selected deployment / configuration combo, and does not use already created manifests.

For the new `allow_uv` attribute in the Python Package Manager section, in the current state of Publisher, we would have to add `allow_uv` as a configuration option and understand how to use it in our manifest generation https://github.com/posit-dev/publisher/blob/7b8c1095768055479df82e43969dd134e296de08/internal/bundles/manifest.go#L155

PR #2587, opened 2025-02-12
Optional Python and R attributes within configuration file

### LINKED ISSUE #2468: Update manifest.json R version population

### LINKED ISSUE #2470: Update manifest.json Python version (platform) population

sagerb:
…n sections and filling in of defaults when deploying. Also added new types and new unit tests.

## Intent

The changes within this PR support the initialization of config files to have default attributes associated with R and Python dependencies, allowing the sections to be empty within the config file. When values are present, they are used by the back end, but when they are not present, the current values from the active interpreters are used, as the deployment manifest is built from the config file. 

Resolves #2468 
Resolves #2470

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

This is the second attempt to implement this functionality, with the previous one being closed without merging (https://github.com/posit-dev/publisher/pull/2568). I have simplified the approach within this PR and believe it is significantly addresses the concerns voiced during the previous PR's review.

The schema has been updated to allow all of the attributes present within the R and Python sections of the configuration to be optional. The presence of the section (`[r]` or `[python]`) has been maintained as the indicator that the deployment has a dependency on the particular interpreter.

Queries of a Configuration maintains its values as persisted. Empty attributes indicate that defaults will be applied.

The queries from the extension will return the configuration objects with empty attributes where defaults are expected. Because the extension relies upon having the attributes resolved (either from values present within the file or from the current default values determined by the active interpreter), the extension now calls a new API to get the default values from the active interpreters (GET /api/interpreters) and resolves the defaults within each configuration object prior to updating its state (store).

When a deployment is initiated, the agent resolves the defaults in the configuration file before generating the manifest file from it.

To remove any potential errors when passing in an R interpreter path vs. a Python interpreter path, I created a typescript type for each. This way I was able to get compilation errors rather than catching the coding issues during unit tests.

## User Impact

New feature: The scenario goes like this:
1. A publisher has created a deployment that uses their currently active R and/or Python interpreters as the versions requested in the manifest.
2. The next day, they upgrade or change their active interpreters and then deploy again. These re-deployments once again request the active R and/or Python interpreters within the manifest.
3. If at any point the user sets the attribute values within the R or Python section, they will be used rather than floating to the active interpreter values.

## Automated Tests

Unit tests were added and modified for the new functionality.

## Directions for Reviewers

The important changes in order are:
1. Schema Changes making the attributes optional
    - internal/schema/schemas/posit-publishing-schema-v3.json
    - internal/schema/schemas/draft/posit-publishing-schema-v3.json
2. Addition of Interpreters
    - internal/interpreters/python_mock.go
    - internal/interpreters/python_test.go
    - internal/interpreters/python.go
    - internal/interpreters/r_mock.go
    - internal/interpreters/r_test.go
    - internal/interpreters/r.go
3. Addition of Get Interpreters API
    - internal/services/api/get_interpreters_test.go
    - internal/services/api/get_interpreters.go
4. Use of the new API and proper passing of the active interpreters into a few calls
    - extensions/vscode/src/api/resources/Configurations.ts
    - extensions/vscode/src/api/resources/ContentRecords.ts
    - extensions/vscode/src/api/resources/Interpreters.ts
    - extensions/vscode/src/api/resources/Packages.ts
    - extensions/vscode/src/api/types/configurations.ts
    - extensions/vscode/src/api/types/interpreters.ts
5. Filling in default values when configurations are queried and stored into the extension's state
    - extensions/vscode/src/state.ts
5. Filling in default values as part of the deployment process
    - internal/state/state.go
    - internal/state/state_test.go
The rest of the file updates are swaps to the new TS Types or signature changes.

## Checklist


internal/config/types.go
```diff
@@ -123,17 +125,63 @@ func (c *Config) HasSecret(secret string) bool {
...
 
 type Python struct {
-	Version        string `toml:"version" json:"version"`
+	Version        string `toml:"version,omitempty" json:"version"`
 	PackageFile    string `toml:"package_file,omitempty" json:"packageFile"`
 	PackageManager string `toml:"package_manager,omitempty" json:"packageManager"`
 }
 
+func (p *Python) FillDefaults(
+	pythonInterpreter *interpreters.PythonInterpreter,
```
marcosnav, line 134: You shouldn't need to use pointers on interfaces,
removing `*` from here (`pythonInterpreter interpreters.PythonInterpreter`) will make things easier below:
```
pythonInterpreter.IsPythonExecutableValid()
```

internal/config/types.go
```diff
@@ -123,17 +125,63 @@ func (c *Config) HasSecret(secret string) bool {
...
+) {
+	if p != nil && pythonInterpreter != nil && (*pythonInterpreter).IsPythonExecutableValid() {
```
marcosnav, line 136: Since you are already within the struct method, I don't think you need to check `p != nil`

internal/config/types.go
```diff
@@ -123,17 +125,63 @@ func (c *Config) HasSecret(secret string) bool {
...
+		python := *pythonInterpreter
+		if p.Version == "" {
+			pythonVersion, pythonVersionError := python.GetPythonVersion()
```
marcosnav, line 139: with the change suggested above, this will get easier: ~python := *pythonInterpreter~
```
if p.Version == "" {
	pythonVersion, pythonVersionError := pythonInterpreter.GetPythonVersion()
```

All these comments apply for R below too.

marcosnav (commented):
Still reviewing, but I already have some comments that you might want to take a look...

internal/config/types_defaults_test.go
```diff
@@ -0,0 +1,269 @@
...
+)
+
+type ConfigFillDefaultsSuite struct {
+	suite.Suite
+	cwd                      util.AbsolutePath
+	log                      logging.Logger
+	rInterpreter             *interpreters.RInterpreter
+	rMissingInterpreter      *interpreters.RInterpreter
+	pythonInterpreter        *interpreters.PythonInterpreter
+	pythonMissingInterpreter *interpreters.PythonInterpreter
```
marcosnav, line 24: Ditto here about interfaces and pointers, generally this isn't needed and complicates things

marcosnav (commented):
@sagerb This PR looks much better than the last one, let me know when it is ready for another look.

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -80,19 +81,18 @@ export class Configurations {
...
-      {
-        python,
-        r,
-      },
+      {},
       {
         params: {
           dir,
+          python: python !== undefined ? python.pythonPath : undefined,
+          r: r !== undefined ? r.rPath : "",
```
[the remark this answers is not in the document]
tdstein: Does this also work? `r: r?.rPath`

internal/schema/schemas/draft/posit-publishing-schema-v3.json
```diff
@@ -115,17 +114,15 @@
         "package_manager": {
           "type": "string",
           "default": "pip",
-          "enum": ["pip", "conda", "pipenv", "poetry", "none"],
           "description": "Package manager that will install the dependencies. If package-manager is none, dependencies are assumed to be pre-installed on the server. The default is 'pip'.",
-          "examples": ["pip"]
+          "examples": ["pip", "conda", "pipenv", "poetry", "none"]
```
tdstein, line 118: Does `uv pip` work?

marcosnav (changes requested):
Getting back to this one and for clarity. The use of pointers on interfaces should be removed.

extensions/vscode/src/api/resources/Configurations.ts
```diff
@@ -80,19 +81,18 @@ export class Configurations {
   // 500 - internal server error
   inspect(
     dir: string,
-    python?: string,
-    r?: string,
+    python: PythonExecutable | undefined,
+    r: RExecutable | undefined,
```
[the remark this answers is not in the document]
sagerb: great catch!

extensions/vscode/src/api/resources/Configurations.ts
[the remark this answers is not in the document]
sagerb: I ended up using a combination of the suggestions:
```
 python: python?.pythonPath,
 r: r?.rPath,
```

extensions/vscode/src/types/shared.ts
```diff
@@ -29,3 +29,21 @@ export type DeploymentObjects = {
...
+}
+
+export type PythonExecutable = {
+  pythonPath: string;
+};
+export function NewPythonExecutable(pythonPath: string): PythonExecutable {
+  return {
+    pythonPath,
+  };
+}
```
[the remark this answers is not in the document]
sagerb: great idea. Thanks!

internal/config/types.go
sagerb: The code and tests have been reworked. Thanks!

internal/config/types.go
sagerb: Did not know that!

internal/config/types.go
sagerb: Updated for python as well as R.

internal/config/types_defaults_test.go
sagerb: The code and tests have been reworked. Thanks.

internal/schema/schemas/draft/posit-publishing-schema-v3.json
sagerb: No, not at this time. That is an outstanding issue (https://github.com/posit-dev/publisher/issues/2510) which uses the requirements for adding UV support as the example. For now, it won't work.

sagerb:
@marcosnav I've pushed changes that should address all of your concerns. Thanks so much for the feedback. Can you re-review?

marcosnav (approved):

PR #2595, opened 2025-02-28
Add `*.lic` to `.gitignore`

jonkeane:
Adding `*.lic` to `.gitignore` so that accidentally committing license files requires additional steps.


dotNomad (approved):
Thanks for the quick PR! 🚀

PR #2600, opened 2025-03-12
Feature default connect server

jeffvroom:
New setting name: positPublisher.defaultConnectServer

- Also support /etc/rstudio/rsession.conf's default server setting to help workbench admins manage this in one place (using XDG_CONFIG_DIRS that defaults to /etc)

## Intent

Addresses workbench feature request: https://github.com/rstudio/rstudio-pro/issues/7466

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [x] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

First check the new extension setting. If that's not set, fallback to RStudio's config value.

## User Impact

No changes unless users configure this value.

## Automated Tests

Happy to provide a test if that would help. 

## Checklist


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   configuration with `type = 'unknown'` suggesting the framework be set. (#2515)
 - Icons have been added to the deployment select dropdown to indicate the
   deployment's server and entrypoint. (#2483)
+- Added new config setting positPublisher.defaultConnectServer to pre-populate the Server URL. Also checks `/etc/rstudio/rsession.conf` for `default-rsconnect-server`. (rstudio-pro#7466)
```
dotNomad, line 31: Thanks for writing up a changelog item for this change. `1.10.0` has already been released so it would be best to add a new section above perhaps like this:

```
## [Unreleased]

### Added

- Added new config setting `positPublisher.defaultConnectServer` to pre-populate
  the Server URL. Also checks `/etc/rstudio/rsession.conf` for
  `default-rsconnect-server`. (rstudio-pro#7466)
```

That way when the next release of Publisher is cut we can easily use this line.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -75,10 +76,17 @@ export async function newCredential(
...
   async function inputServerUrl(input: MultiStepInput, state: MultiStepState) {
     const thisStepNumber = assignStep(state, "inputServerUrl");
-    const currentURL =
+    let currentURL =
       typeof state.data.url === "string" && state.data.url.length
         ? state.data.url
         : "";
+
+    if (currentURL === "") {
+       const configURL: string | null = getXDGConfigProperty("rstudio/rsession.conf", "default-rsconnect-server");
```
dotNomad, line 85: It looks like this section only got the `XDGConfigProperty` treatment did you mean to include the `currentURL = extensionSettings.defaultConnectServer();` checking here as well? It does seem like a good spot for it.

Even better could provide a single function that either returns the `defaultConnectServer` or the XDG Config Property depending for both of these to use.

extensions/vscode/src/utils/config.ts
```diff
@@ -0,0 +1,47 @@
+import * as fs from 'fs';
+import * as path from 'path';
```
dotNomad, line 2: Nitpicky but just to match other parts of the codebase and use the default export:

```suggestion
import fs from "fs";
import path from "path";
```

extensions/vscode/src/utils/config.ts
```diff
@@ -0,0 +1,47 @@
...
+
+
+export function getXDGConfigProperty(configFilePath: string, propertyName: string): string | null {
+  console.log("In getXDGConfigProperty with: " + configFilePath + " and: " + propertyName);
+  let filePath = getConfigFile(configFilePath);
+  if (!filePath) {
+    console.log("No file path with path found");
+    return null;
+  }
+  const fileContent = fs.readFileSync(filePath, 'utf-8');
```
dotNomad, line 28: It may be optimal to use `fs.readFile` and `fs.exists` above and make both of these functions asynchronous then `await` their calls in the two `multiStepInputs` files.

That would prevent other VS Code operations from halting while files are being read or checked.

extensions/vscode/src/utils/config.ts
```diff
@@ -0,0 +1,47 @@
...
+    const trimmedLine = line.trim();
+
+    if (trimmedLine.startsWith('#')) {
+        continue;
+    }
+
+    const [name, value] = trimmedLine.split('=', 2);
+
+    if (name && name.trim() === propertyName) {
+      return value ? value.trim() : '';
```
dotNomad, line 42: It looks like the rest of this function returns `null` if there isn't a value, but this returns an empty string. Should it return `null` instead?

extensions/vscode/src/utils/config.ts
```diff
@@ -0,0 +1,47 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+function getConfigFile(configFilePath: string): string | null {
+  const configDirs = process.env.XDG_CONFIG_DIRS ? 
+         process.env.XDG_CONFIG_DIRS.split(':') : 
+         ['/etc'];
```
dotNomad, line 7: Will this work on Windows machines or cause a file failure? I'm not sure if we need to use `path.sep` from the `path` module or something else.

dotNomad (commented):
Thanks for contributing to Publisher this PR is looking great.

I had a few questions and suggestions in comments.

The VS Code extension config setting worked great for me on macOS for the new deployment flow. Is there a good way on macOS for me to test the `default-rsconnect-server` without setting the `XDG_CONFIG_DIRS` environment variable directly?

I noticed that this failed our linting check in CI. To fix that you can run `npm run format` from the root dir after installing dependencies with `npm install`. If you need any help getting that working let me know!

extensions/vscode/src/multiStepInputs/newCredential.ts
jeffvroom: Great catch and suggestion, thanks!

extensions/vscode/src/utils/config.ts
jeffvroom: Yeah, I thought about doing that but it makes it awkward in the code for config values to be async. Since it's now linux only, and will just be an extra 'stat' call to a local directory for most users maybe it's ok? Or I could load the value when the plugin starts. Happy to do that if you think it's worth it.

extensions/vscode/src/utils/config.ts
jeffvroom: Nice catch!

extensions/vscode/src/utils/config.ts
jeffvroom: I don't think we use this file on windows or mac so I made it linux only with an explicit platform check. Thanks also for catching that!

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added new config setting positPublisher.defaultConnectServer to pre-populate the Server URL. Also checks `/etc/rstudio/rsession.conf` for `default-rsconnect-server`. (rstudio-pro#7466)
```
nealrichardson, line 12:
```suggestion
- Added new config setting `positPublisher.defaultConnectServer` to pre-populate the Server URL. Also checks `/etc/rstudio/rsession.conf` for `default-rsconnect-server`. (rstudio-pro#7466)
```

Also, maybe for the second sentence in that note, is it worth saying that that is for configuration on Posit Workbench? Maybe even just say it automatically picks it up from configuration on Posit Workbench (rather than naming the option and file), and link to the Workbench docs about setting it?

extensions/vscode/package.json
```diff
@@ -46,6 +46,11 @@
             "markdownDescription": "Controls whether Posit Publisher will use the system keychain to store credentials. When disabled, or on a system without a keychain, `~/.connect-credentials` is used. After changing this setting, you must restart VS Code for it to go into effect.",
             "type": "boolean",
             "default": true
+          },
+          "positPublisher.defaultConnectServer": {
+            "markdownDescription": "Provides the default URL for the connect server when defining new deployments.",
```
nealrichardson, line 51:
```suggestion
            "markdownDescription": "Provides the default URL for the Posit Connect server when defining new deployments.",
```

extensions/vscode/src/extension.ts
```diff
@@ -169,4 +170,19 @@ export const extensionSettings = {
     );
     return value !== undefined ? value : true;
   },
+  defaultConnectServer(): string {
+    const configuration = workspace.getConfiguration("positPublisher");
+    let value: string | undefined = configuration.get<string>(
+      "defaultConnectServer",
+    );
+
+    if (value === undefined || value === "") {
```
nealrichardson, line 179: Maybe add a comment explaining that this is to pull it from where Workbench puts it?

nealrichardson (commented):
Thank you for the PR! Here are a few cosmetic suggestions. 

I don't know the state of the test suite in publisher, but it would be nice if this had unit test coverage.

extensions/vscode/src/extension.ts
jeffvroom: Yes, good call on that comment.

CHANGELOG.md
jeffvroom: Thanks, nice catch on that.

jeffvroom:
> Thank you for the PR! Here are a few cosmetic suggestions.
> 
> I don't know the state of the test suite in publisher, but it would be nice if this had unit test coverage.

Thanks for all the great feedback. I learn so much contributing to a new project, and appreciate your patience. 

To me it feels like this might be one of those "too much effort, too little reward" to automate testing here at the unit level, unless I missed some testing infrastructure.  Once it's pulled into workbench, we'll test it there, and if it doesn't work for someone, they will let us know :)  Should we coordinate on release schedule?  Ours nexst release is scheduled for mid April freeze. I think we're working on letting users update extensions after the fact, and this is a minor thing obviously.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added new config setting positPublisher.defaultConnectServer to pre-populate the Server URL. Also checks `/etc/rstudio/rsession.conf` for `default-rsconnect-server`. (rstudio-pro#7466)
+- Added a new config setting `positPublisher.defaultConnectServer` to pre-populate the Server URL. Also checks the [default RStudio/Posit Workbench setting](https://docs.posit.co/ide/server-pro/rstudio_pro_sessions/rstudio_connect_server.html) (rstudio-pro#7466)
```
nealrichardson, line 12: I doubt there are any versions of RStudio Workbench where this is relevant, are there?

```suggestion
- Added a new config setting `positPublisher.defaultConnectServer` to pre-populate the Server URL. Also checks the [default Posit Workbench setting](https://docs.posit.co/ide/server-pro/rstudio_pro_sessions/rstudio_connect_server.html) (rstudio-pro#7466)
```

extensions/vscode/src/extension.ts
```diff
@@ -176,6 +176,7 @@ export const extensionSettings = {
       "defaultConnectServer",
     );
 
+    // For RStudio and Posit Workbench users, look here as a final step
```
nealrichardson, line 179: Feel free to disregard if I'm missing something, but I don't think this is relevant for RStudio, right?

```suggestion
    // For Posit Workbench users, look here as a final step
```

jeffvroom: The file `rsession.conf`'s `default-rsconnect-server` setting is also in open-source, so it will potentially help RStudio users. I realized that when I started documenting it.

sagerb (changes requested):
Thanks so much for submitting a PR with this improvement. Your code looks good and I really appreciate your willingness to jump right into the codebase with your change.

I have provided a minor requested change that will hopefully improve the user's experience without providing an error message, as well as a real minor NIT on a description string.

Thanks again and I'll look forward to reviewing the changes.

CHANGELOG.md
jeffvroom: Both RStudio and Workbench are relevant.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -75,10 +75,15 @@ export async function newCredential(
...
-    const currentURL =
+    let currentURL =
       typeof state.data.url === "string" && state.data.url.length
         ? state.data.url
         : "";
+
+    if (currentURL === "") {
+      currentURL = extensionSettings.defaultConnectServer();
+    }
+
```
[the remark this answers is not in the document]
jeffvroom: Great idea. I added the logic though it was a little awkward to share more code.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -531,10 +531,14 @@ export async function newDeployment(
...
     if (newCredentialByAnyMeans()) {
-      const currentURL = newDeploymentData.newCredentials.url
+      let currentURL = newDeploymentData.newCredentials.url
         ? newDeploymentData.newCredentials.url
         : "";
 
+      if (currentURL === "") {
+        currentURL = extensionSettings.defaultConnectServer();
+      }
+
```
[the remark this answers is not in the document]
jeffvroom: Yeah, it's a little duplicated logic and maybe worth a refactor on the next big change. I don't want to add to our testing burden by getting heavy handed.

extensions/vscode/src/utils/config.ts
sagerb: I agree with @dotNomad's opinion, so if you don't mind, yes please.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -69,6 +69,16 @@ export async function newCredential(
...
+  function hasExistingCredential(serverURL: string): boolean {
+    return (
+      credentials.find((credential) => {
+        const existing = normalizeURL(credential.url).toLowerCase();
+        const newURL = normalizeURL(serverURL).toLowerCase();
+        return newURL.includes(existing);
+      }) !== null
+    );
+  }
+
```
sagerb, line 81: Would you mind moving this logic into `state.ts`, and call it `findCredentialByURL`? That way, it will sit beside the existing `findCredential` (which operates on name), and be available to use from your `newCredential` code as well. 

I'd suggest also following that same pattern that `findCredential` uses, to simply return null when it can't find it, and then you can evaluate the result against it to answer the question if it already exists.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -526,6 +526,16 @@ export async function newDeployment(
...
+  function hasExistingCredential(serverURL: string): boolean {
+    return (
+      credentials.find((credential) => {
+        const existing = normalizeURL(credential.url).toLowerCase();
+        const newURL = normalizeURL(serverURL).toLowerCase();
+        return newURL.includes(existing);
+      }) !== null
+    );
+  }
+
```
sagerb, line 538: This would be changed with my request to move this functionality into a common location (with `state.ts`).

sagerb (commented):
Just a request to move some common functionality into the location that we've come to expect it and you should be all set. The code is looking great. Thanks for your help with this one.

extensions/vscode/src/multiStepInputs/newCredential.ts
jeffvroom: I moved that function to state.ts but imported just the function since it doesn't seem like newCredential and newDeployment are designed to use PublisherState directly. Let me know if there's a better way to do that.

sagerb: >it doesn't seem like newCredential and newDeployment are designed to use PublisherState directly

@jeffvroom , you are absolutely right. I'm sorry for the incorrect direction that I provided you. 

Tomorrow, I'm going to push up some changes to this branch that should satisfy what I was looking for without causing you any additional effort. When I do get a commit pushed up, I'll post another message and if you like those changes, we can get this PR merged. Thanks again for all of your effort on this.

jeffvroom: Cool, that will make this a lot cleaner. Thanks for your help.

sagerb:
@jeffvroom Just wanted to check if you've completed everything you wanted on this PR. I've resolved a merge conflict with `main` and pushed that to the branch.

jeffvroom:
Yep, looks good to me. I just tested the latest with workbench and it works great. I think it's ready to merge. We could probably cover this feature in the workbench automation. Let me know what else I can do to help.

sagerb:
> probably cover this feature in the workbench automation. Let me know what else I can do to help.

If you can do that, I'm good with approving and merging this! Thanks so much for your help with the publisher!

sagerb (approved):
We've worked through the requests and it looks good to me.

PR #2604, opened 2025-03-13
Upgrade `test/e2e` dependencies

dotNomad:
Upgrades the dependencies in the `test/e2e` directory which also upgrades `axios` to the latest version.

## Intent

Supersedes #2603


sagerb (approved):
I built successfully and ran a quick manual test. This change looks good. Thanks for doing it!

PR #2605, opened 2025-03-13
Upgrade golang.org/x/net dependency

dotNomad:
Upgrades the `golang.org/x/net` dependency to the latest version.

Adjusts a few more packages as well as a result of `go mod vendor`

## Intent

Supersedes #2602


sagerb (approved):
I built and ran the extension. This change looks good.

PR #2606, opened 2025-03-21
Add multiple Quarto, R and Python versions into e2e docker images

### LINKED ISSUE #2589: E2E tests: Add R and Quarto installations to docker image hosting the code-server
To fully test the functionality within the Publisher extension within the e2e environment, there should be multiple versions of Python and R present, as well as a base version of Quarto. These should be compatible enough to match the versions installed within the Connect docker image.

The E2E test infrastructure is currently using the image `codercom/code-server`, which does not have the required R and Quarto installations built in. This is blocking our ability to write e2e tests that deploy R and Quarto content. (It does have python3 installed on it.) It would appear that one solution to this is to find another docker image which supports installing additional packages, such as https://hub.docker.com/r/linuxserver/code-server

branch in progress: sagerb-replace-codeserver-docker-volume

sagerb:
## Intent

Resolves #2589

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

Created a new docker image to base the connect and vscode/Cypress images off of. Within this image, I install the following interpreter versions:
- Python:
    - v3.10.16
    - v3.11.11
    - v3.12.9
- R:
    - v4.1.3
    - v4.2.3
    - v4.3.3
    - v4.4.2
- Quarto:
    - v1.4.556
    - v1.5.52
    - v1.6.42
    - v1.7.6

Created a new docker image to host VSCode based on the image above.
Updated the connect server image to derive from the base image above.

Tests were updated to work within the new environment.

Justfile recipes were updated a bit.
- added `build-images` recipe
- added `build-publisher` recipe
- reduced the scope of the `dev` recipe a bit (while maintaining the old recipe in `old-dev`)
- added `e2e` recipe to run tests

## User Impact

none - tooling only

## Automated Tests

e2e tests are automated, but do not yet run in CI. They must be run manually.

## Directions for Reviewers

To run the e2e tests:
```bash
cd `test/e2e`
just build-images
just dev
```

manually run each of the test specs.

## Checklist


test/e2e/Dockerfile.connect
```diff
@@ -1,32 +1,81 @@
...
+# Setup users and groups using cloned script from Connect's docker image: 
+# docker/ubuntu24/Dockerfile
+
+# An incoming Jenkins uid/gid alters how the account is created. Allows
+# matching uid/gid on the launching Jenkins agent. Configured late to keep
+# more layers reusable.
+ARG JENKINS_GID=999
+ARG JENKINS_UID=999
+
+# Create the jenkins user with the same id:gid as the jenkins-node
```
marcosnav, line 18: Is this setup of jenkins, users and groups really needed for Publisher tests?

marcosnav (commented):
First pass to this, still taking a look

test/e2e/Dockerfile.vscode
```diff
@@ -1,7 +1,38 @@
...
-FROM codercom/code-server:39 AS code-server
+FROM e2ebase AS code-server
+
+RUN export TZ=America/New_York
+
+# 
+# Install code-server
+# 
+
+RUN curl -fsSL https://code-server.dev/install.sh | sh
```
marcosnav, line 13: Isn't there a way to keep using `FROM codercom/code-server:39 AS code-server`? Could you ellaborate a bit more on why it is needed to install and setup code-server by hand?

marcosnav:
Tried it out, the majority of tests are passing, except for "New Credentials" which is consistently failing. Seems the changes on `selectors.js` are related to the failure.

test/e2e/Dockerfile.connect
sagerb: I matched what was being done within the Connect dev image, so that the code would run the same. There are certain scenarios in which we would fail to deploy because of content ownership and collaboration. So between these two factors, as well as making it easier to maintain (IMO - when different than connect's dev image, update it), I think there is an advantage to not reducing these small items away.

test/e2e/Dockerfile.vscode
sagerb: codercom/code-server image uses Debian GNU/Linux 12. I encountered issues trying to install the additional software needed for the enhanced environment, so I followed the instructions available from the folks who provide that image to install the code-server myself upon the ubuntu22 image. This allowed me to simply replicate the supported software and packages between the client environment and the server environment as well, which I saw as a simplification for the testing environment.

sagerb:
> Tried it out, the majority of tests are passing, except for "New Credentials" which is consistently failing. Seems the changes on `selectors.js` are related to the failure.

I'll take a look again. All the tests were consistently passing when I submitted the PR. Thanks.

test/e2e/Dockerfile.connect
marcosnav: Got it, I see that it could be tempting to have the same setup that we have in Connect dev image, still I'll push a bit back on this because:
1. We would need to remember to update in both places, easy to get out of sync
2. Putting unnecessary effort on syncing something that potentially is not needed

I think there is a fine line between having e2e tests for `Publisher + Connect` vs e2e tests publishing to Connect. There are tests that should be exclusive to Connect. 

What you have in `Dockerfile.base` is great and totally needed to test requirements between Publisher and Connect, aside from that, I don't think having things like groups and RunAs specifics is needed, a few users should be enough to test publishing scenarios including permissions and collaboration.

sagerb:
>Tried it out, the majority of tests are passing, except for "New Credentials" which is consistently failing. Seems the changes on selectors.js are related to the failure.
@marcosnav I ran this again, for the first time on my new laptop. All the tests, including the new credentials, worked successfully for me.

I seem to remember that there was an old credential which stayed around in the test environment from the previous code, which I had to manually remove. This was a one-time operation, which should only be applicable to those who have run the e2e tests previously. Can you try `just start code-server`, and see if publisher has any existing credentials configured? If so, can you remove it, exit out, and then try the tests again?

@dotNomad Could you possibly give this PR a run, to see if the tests work for you? Just follow the steps listed above for `reviewers`, it shouldn't take too long.

dotNomad:
> @dotNomad Could you possibly give this PR a run, to see if the tests work for you? Just follow the steps listed above for `reviewers`, it shouldn't take too long.

Tried this out. 

```bash
cd `test/e2e`
just build-images
just dev
```

Brought up the Cypress UI, and I ran each of the specs. 

`credentials.cy.js`, `deployments.cy.js`, `embedded-deployments.cy.js`, and `error-err-config.cy.js` failed with: `command not found: rsconnect`. 

I assume I have to install `rsconnect` prior to running the e2e tests. Is there a recommended way to do that other than having a virtualenv loaded up prior to running `just dev`?

sagerb:
@dotNomad Not sure, that was part of the original work that Marcos implemented. I just went ahead and installed it globally.

sagerb:
@dotNomad I've pushed a requirements.txt file and verified that I could run the e2e tests using a virtual environment created from the `test/e2e` subdirectory. So I think this should be a path which will keep you from having to install `rsconnect-python` globally, although, I have the feeling you'll eventually want to.

Just be sure to run the `just` command from that subdirectory, after you have created a virtual environment in that subdirectory, activated it and have run `pip install -r requirements.txt`.

dotNomad:
> So I think this should be a path which will keep you from having to install `rsconnect-python` globally, although, I have the feeling you'll eventually want to.

I cannot install it globally with my Python installed via `brew` unfortunately, but I did setup a virtual env as you suggested. I re-tested the specs:
- I got a failure on "Create Deployments fastAPI in subdirectory of workspace"
  - It timed out looking for `.quick-input-widget` which is strange given the screenshot it is showing me. The `quick-input-widget` `div` is there
- I re-ran `credentials.cy.js` several times and they passed each time for me
- Everything else passed without issue

If I can assist more let me know :)

dotNomad:
> It timed out looking for .quick-input-widget which is strange given the screenshot it is showing me. The quick-input-widget div is there

I increased the timeout to `10000` on the `.quick-input-widget` get. I re-tested 3 times and it passed each time

dotNomad (approved):
This worked great for me locally with the timeout change, and the changes here look good 😎 

I will note I'm not a Dockerfile expert so if you do want a more thorough look I'd recommend getting a second approval. I did look through them and it looked like there were already discussions in the things I would have asked about 👍

PR #2612, opened 2025-03-28
Only enable publishing log view after user has initiated their first …

### LINKED ISSUE #2596: Don't show the log panel until its helpful.
See https://github.com/posit-dev/positron/issues/6596 for the Positron issue that this came from. 

We shouldn't show the logs panel (even just in the tab list) until someone at least activates the extension, but there's an argument that if it's all going to be empty anyway (e.g. if someone hasn't published yet in this session) no good reason to have listed and taking up space there.

What hooks do we have in VSCode and Positron to hide this until it is actually useful and helpful?

sagerb:
…deployment operation

## Intent

This PR suppresses the Posit Publisher Log panel view until the user has clicked the deployment button from within the publishing extension. Until they do so, it remains hidden as it has no value until a deployment is initiated.

Resolves #2596 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I added an additional context within the view's `when` definition, initialized it to "false" at extension startup, and set it to "true"  when the user initiates the deployment.

On load, first view:

image

After clicking the deploy button:

image

## User Impact

There shouldn't be any impact since we do not keep logs from previous runs (since the agent is killed and restarted each time the VSCode window is restarted).

## Automated Tests

This is not something we could test in unit tests as it is implemented within VSCode's infrastructure. It would be possible to test within e2e testing, but that infrastructure is still under development.

## Directions for Reviewers

Should be able to reproduce the user scenario easily enough. Open VSCode, activate the publisher extension, verify the output panel does not list the Posit Publisher Log. Deploy something and verify that the log window is available once you clicked the deploy your project button.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+### Fixed
+
+### Changed
+
+- The Publisher Log Panel now stays hidden until the user initiates a deployment operation
+  from within the extension. It remains visible until the VSCode window is restarted. (##2596)
+
 ## [1.10.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- The Publisher Log Panel now stays hidden until the user initiates a deployment operation
-  from within the extension. It remains visible until the VSCode window is restarted. (##2596)
+- The Publisher Log Panel has now been renamed to "PUBLISHER" and now will stay hidden until the
+  user initiates a deployment operation from within the extension. It remains visible until the
+  VSCode window is restarted. (#2596)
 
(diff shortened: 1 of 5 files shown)


jonkeane:
Thanks for this! 

This is separate from the specifics of this PR (but might be a good campsite rule fold in...). This might have come up before, but seeing the screenshots (especially in the context of conserving space), would it be possible (and advisable) to have this panel be called "Posit Publisher" instead of "Posit Publisher Log"? It's "only" four characters, but every bit helps.

sagerb:
> This is separate from the specifics of this PR (but might be a good campsite rule fold in...). This might have come up before, but seeing the screenshots (especially in the context of conserving space), would it be possible (and advisable) to have this panel be called "Posit Publisher" instead of "Posit Publisher Log"? It's "only" four characters, but every bit h

Certainly! I can make that change within this PR.

tdstein:
> Thanks for this!
> 
> This is separate from the specifics of this PR (but might be a good campsite rule fold in...). This might have come up before, but seeing the screenshots (especially in the context of conserving space), would it be possible (and advisable) to have this panel be called "Posit Publisher" instead of "Posit Publisher Log"? It's "only" four characters, but every bit helps.

Or even just "Publisher" if that makes sense within the Positron context.

sagerb:
Made the change as requested for the panel title.

Screenshot 2025-04-08 at 1 11 48 PM

PR #2617, opened 2025-04-08
Support detecting python interpreter requirements from standard files

amol-:
## Intent

Newer connect versions support a `environment.python.requires` option that tells the server which version of Python can be used to run the content. This constraint is meant to be detected from `.python-version`,`pyproject.toml` and `setup.cfg`. The support was already done in `rsconnect-python` and this PR adds the same capability to the publisher.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The `defaultPythonInspector` will use a newly added `PyProjectPythonRequires` entity to determine if the python project has any constraint over the Python version that should be used to run the project.
If such constraint is detected the `defaultInitialize` will add it to the newly generated project configuration which will then be used to generate the Manifest including that option.

## User Impact

If the user had provided a interpreter version constraint in any of the 3 supported files, the constraint will now be propagated to Connect.

## Automated Tests

- [ ] Added a suite of tests for PyProjectPythonRequires
- [ ] Added an test for a newly generated manifest out of a project with requirements

## Directions for Reviewers

There is an open question about the fact that currently the information detected from the inspector is mostly ignored and replaced with hardcoded answers from the interpreters API. So that might heavily influence this PR structure as it evolves.

## Checklist

### DIFF
--- a/internal/bundles/manifest.go
+++ b/internal/bundles/manifest.go
@@ -53,9 +53,14 @@ type Metadata struct {
 	HasParameters   bool            `json:"has_parameters,omitempty"`   // True if this is content allows parameter customization.
 }
 
+type EnvironmentPython struct {
+	PythonRequires string `json:"requires"` // The Python version to use for the content environment
+}
+
 type Environment struct {
-	Image    string `json:"image"`    // The image to use during content build/execution
-	Prebuilt bool   `json:"prebuilt"` // Determines whether Connect should skip the build phase for this content.
+	Image    string             `json:"image"`            // The image to use during content build/execution
+	Prebuilt bool               `json:"prebuilt"`         // Determines whether Connect should skip the build phase for this content.
+	Python   *EnvironmentPython `json:"python,omitempty"` // If non-null, specifies the Python environment
 }
 
 type Python struct {
@@ -174,6 +179,15 @@ func NewManifestFromConfig(cfg *config.Config) *Manifest {
 				Name:        cfg.Python.PackageManager,
 				PackageFile: cfg.Python.PackageFile,
 			},
+		}
+		// If the configuration specifies a specific python version constraint
+		// (e.g. ">=3.8"), declare the environment requires that version.
+		if cfg.Python.PythonRequires != "" {
+			m.Environment = &Environment{
+				Python: &EnvironmentPython{
+					PythonRequires: cfg.Python.PythonRequires,
+				},
+			}
 		}
 	}
 	if cfg.Jupyter != nil {
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -128,6 +128,7 @@ type Python struct {
 	Version        string `toml:"version,omitempty" json:"version"`
 	PackageFile    string `toml:"package_file,omitempty" json:"packageFile"`
 	PackageManager string `toml:"package_manager,omitempty" json:"packageManager"`
+	PythonRequires string `toml:"python_requires,omitempty" json:"pythonRequires"`
 }
 
 func (p *Python) FillDefaults(
--- a/internal/schema/schemas/draft/posit-publishing-schema-v3.json
+++ b/internal/schema/schemas/draft/posit-publishing-schema-v3.json
@@ -116,6 +116,12 @@
           "default": "pip",
           "description": "Package manager that will install the dependencies. If package-manager is none, dependencies are assumed to be pre-installed on the server. The default is 'pip'.",
           "examples": ["pip", "conda", "pipenv", "poetry", "none"]
+        },
+        "python_requires": {
+          "type": "string",
+          "default": "",
+          "description": "Python interpreter version required to run the content. If not specified it will be detected from the one in use.",
+          "examples": [">3.8", "<3.9", "=3.5"]
         }
       }
     },
--- a/internal/schema/schemas/posit-publishing-schema-v3.json
+++ b/internal/schema/schemas/posit-publishing-schema-v3.json
@@ -105,6 +105,12 @@
           "default": "pip",
           "description": "Package manager that will install the dependencies. If package-manager is none, dependencies will not be installed.",
           "examples": ["pip", "none"]
+        },
+        "python_requires": {
+          "type": "string",
+          "default": "",
+          "description": "Python interpreter version required to run the content. If not specified it will be detected from the one in use.",
+          "examples": [">3.8", "<3.9", "=3.5"]
         }
       }
     },
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,10 @@ require (
 	golang.org/x/net v0.37.0
 )
 
-require golang.org/x/crypto v0.36.0 // indirect
+require (
+	golang.org/x/crypto v0.36.0 // indirect
+	gopkg.in/ini.v1 v1.67.0 // indirect
+)
 
 require (
 	github.com/alessio/shellescape v1.4.2 // indirect
--- a/go.sum
+++ b/go.sum
@@ -103,6 +103,8 @@ gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127/go.mod h1:Co6ibVJAznAaIkqp8
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c h1:Hei/4ADfdWqJk1ZMxUNpqntNwaWcugrBjAiHlqqRiVk=
 gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c/go.mod h1:JHkPIbrfpd72SG/EVd6muEfDQjcINNoR0C8j2r3qZ4Q=
 gopkg.in/errgo.v2 v2.1.0/go.mod h1:hNsd1EY+bozCKY1Ytp96fpM3vjJbqLJn88ws8XvfDNI=
+gopkg.in/ini.v1 v1.67.0 h1:Dgnx+6+nfE+IfzjUEISNeydPJh9AXNNsWbGP9KzCsOA=
+gopkg.in/ini.v1 v1.67.0/go.mod h1:pNLf8WUiyNEtQjuu5G5vTm06TEv9tsIgeAvK8hOrP4k=
 gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
 gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
--- a/internal/initialize/initialize.go
+++ b/internal/initialize/initialize.go
@@ -125,6 +125,7 @@ func (i *defaultInitialize) inspectProject(base util.AbsolutePath, pythonExecuta
 			return nil, err
 		}
 		cfg.Files = append(cfg.Files, fmt.Sprint("/", pyConfig.PackageFile))
+		cfg.Python.PythonRequires = pyConfig.PythonRequires
 	}
 
 	rInspector, err := i.rInspectorFactory(base, rExecutable, log, i.rInterpreterFactory, nil)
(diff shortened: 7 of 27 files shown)


internal/bundles/manifest.go
```diff
@@ -53,9 +53,14 @@ type Metadata struct {
...
+type EnvironmentPython struct {
+	PythonRequires string `json:"requires"` // The Python version to use for the content environment
+}
+
 type Environment struct {
-	Image    string `json:"image"`    // The image to use during content build/execution
-	Prebuilt bool   `json:"prebuilt"` // Determines whether Connect should skip the build phase for this content.
+	Image    string             `json:"image"`            // The image to use during content build/execution
+	Prebuilt bool               `json:"prebuilt"`         // Determines whether Connect should skip the build phase for this content.
+	Python   *EnvironmentPython `json:"python,omitempty"` // If non-null, specifies the Python environment
```
sagerb, line 63: I didn't catch this before when Brian was going over the feature. Why didn't y'all place this under the python section in the manifest?

internal/config/types.go
```diff
@@ -128,6 +128,7 @@ type Python struct {
 	Version        string `toml:"version,omitempty" json:"version"`
 	PackageFile    string `toml:"package_file,omitempty" json:"packageFile"`
 	PackageManager string `toml:"package_manager,omitempty" json:"packageManager"`
+	PythonRequires string `toml:"python_requires,omitempty" json:"pythonRequires"`
 }
 
 func (p *Python) FillDefaults(
```
sagerb, line 134: Should we merge the discovered versions together in a priority that makes sense, as we fill out the default? For example, use the version discovered through `python_requires.go` rather than the interpreter runtime version?

Or, should we not even supply a default python version when we do have a python requires value? I ask because I'm seeing an issue where the manifest is being built as follows:

```
	"python": {
		"version": "3.9.3",
		"package_manager": {
			"name": "pip",
			"package_file": "requirements.txt"
		}
	},
	"environment": {
		"image": "",
		"prebuilt": false,
		"python": {
			"requires": "<4.0"
		}
	},
```
and the Connect server is responding with an error code indicating that `3.9.3` is not available on the server. 

Shouldn't this deploy correctly, since the python requires is anything less than 4? (Server has 3.11.3 on it.)

internal/interpreters/python_requires.go
```diff
@@ -0,0 +1,90 @@
...
+	ProjectPath util.AbsolutePath
+}
+
+func NewPyProjectPythonRequires(projectPath util.AbsolutePath) *PyProjectPythonRequires {
+	return &PyProjectPythonRequires{
+		ProjectPath: projectPath,
+	}
+}
+
+func (p *PyProjectPythonRequires) GetPythonVersion() (string, error) {
```
sagerb, line 27: Should we rename this to indicate that it is a python version range?

internal/interpreters/python_requires.go
```diff
@@ -0,0 +1,90 @@
...
+	if version, err := p.readPyProjectToml(); err == nil && version != "" {
+		return version, nil
+	}
+	if version, err := p.readSetupCfg(); err == nil && version != "" {
+		return version, nil
+	}
+	return "", errors.New("no python version requirement found")
+}
+
+func (p *PyProjectPythonRequires) readPythonVersionFile() (string, error) {
```
sagerb, line 40: They look pretty simple, but would you mind adding some comments describing the format expected in each of these files (maybe an example)?

internal/schema/schemas/draft/posit-publishing-schema-v3.json
```diff
@@ -116,6 +116,12 @@
           "default": "pip",
           "description": "Package manager that will install the dependencies. If package-manager is none, dependencies are assumed to be pre-installed on the server. The default is 'pip'.",
           "examples": ["pip", "conda", "pipenv", "poetry", "none"]
+        },
+        "python_requires": {
```
sagerb, line 120: Since this attribute is within the [python] section, can we simply call it `requires` rather than `python_requires`?

internal/schema/schemas/draft/posit-publishing-schema-v3.json
```diff
@@ -116,6 +116,12 @@
...
+          "type": "string",
+          "default": "",
+          "description": "Python interpreter version required to run the content. If not specified it will be detected from the one in use.",
```
sagerb, line 123: We probably need to describe how it relates to `python.version` and that this one includes the ability to specify a range. We should also specify what files are searched in.

I suggest we also update the description of the original version field to describe its relationship with python_requires.

internal/schema/schemas/posit-publishing-schema-v3.json
```diff
@@ -105,6 +105,12 @@
           "default": "pip",
           "description": "Package manager that will install the dependencies. If package-manager is none, dependencies will not be installed.",
           "examples": ["pip", "none"]
+        },
+        "python_requires": {
+          "type": "string",
+          "default": "",
+          "description": "Python interpreter version required to run the content. If not specified it will be detected from the one in use.",
+          "examples": [">3.8", "<3.9", "=3.5"]
```
sagerb, line 113: same comments as above.

sagerb (changes requested):
The direction implemented within this PR is exactly what I was trying to describe. Thank you so much for diving in.

Most of my comments are minor, but I do think we need to determine what is going on when the publisher sets both the python.version and the environments.python.requires within the manifest. I don't think we're currently seeing the expected behavior.

I look forward to reviewing the updates! Thanks!

internal/bundles/manifest.go
```diff
@@ -214,7 +228,12 @@ func (manifest *Manifest) AddFile(path string, fileMD5 []byte) {
 }
 
 func (manifest *Manifest) ToJSON() ([]byte, error) {
-	return json.MarshalIndent(manifest, "", "\t")
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetEscapeHTML(false)
+	enc.SetIndent("", "\t")
+	err := enc.Encode(manifest)
+	return buf.Bytes(), err
```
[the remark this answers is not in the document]
amol-: The requires attribute raised the issue, but it was probably a good idea anyway.
The odd behavior was that requirements like `>=3.8` were being escaped to `\u003e=3.8` for now valid reason. It just made harder to debug the manifest, and as we don't inject the manifest into HTML it was unecessary

internal/config/types.go
amol-: The idea is that we set both `python.version` and `environment.python.requires`.
Older versions of connect will care about `python.version` which is set like it always did and the behavior for the use is unchanged.
Newer versions of connect will ignore `python.version` and only care about `environment.python.requires`.

The `python` section of the manifest now simply reports the state of the inspected project environment, but no longer dictates the shape of the remote execution environment, which is instead `environment.python`.

internal/schema/schemas/draft/posit-publishing-schema-v3.json
amol-: The name `python_requires` is to be consistent with the option itself has, but maybe using `requires_python` which is how instead the option is named in the `.toml` could be an improvement?

I'd like to keep the option bound to the names it has in the actual python config files

amol-:
> 
> Could you also add some of the version files into the sample projects located under `test/sample-content`?
> 

Yep, I still have to work on the Tests 🙂
Just wanted to make sure we agreed on the direction to go before building too much on top of it.

PS: I'd like to propose an experience improvement for the user.
Instead of generating an empty `python` section in the toml

```
[python]
```

I propose we do show the detected values commented out when we preview the `.toml`

```
[python]
# Python support was detected
# This are the settings that we identified when inspecting your projects,
# leave them commented if you want to allow automatic detection of those settings at deploy time
# or uncomment them if you want to force them independently from the selected environment
#
# python_requires = ">=3.8"
# package_manager = "pip"
# requirements = "requirements.txt"
```

This will retain the behavior that you want because the options are commented out,
but will also be informative for the user about what we detected and thus are probably going to use at deploy time too,
while allowing the user to easily influence the project if for example they had requirements in another file.

internal/bundles/manifest.go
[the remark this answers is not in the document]
sagerb: sounds good! Thanks for the explanation.

internal/config/types.go
sagerb: OK, so in that case, shouldn't the manifest I mentioned above deployed?

sagerb:
> I propose we do show the detected values commented out when we preview the `.toml`

Totally agree that this would be an improvement, and it has come up in the past. There are a lot of benefits for the user if we can accomplish this. Unfortunately, the TOML parser we use does not support comments other than what we insert at the beginning. Unless we walk away from the go TOML library we are using or create a replacement, we're blocked from making this great improvement. Mike had last looked at it and he said we'd have to create a new AST parser which supported the commented fields. Currently, if a user comments something out, we loose that comment when we update the file. We've also considered just updating the file directly, but that is also something we haven't been able to schedule as it is not trivial with the parsing still required.

sagerb:
>Unless we walk away from the go TOML library we are using or create a replacement, we're blocked from making this great improvement.

If you have any ideas on how we could accomplish this, please speak up! I've thought about using packages from other languages, but the cons for introducing it didn't make it feel like a good approach. 

The original idea was to generate a configuration file with most of the options documented but commented out so that users could be educated and easily uncomment to make changes. I still love the idea; I just have to figure out how to get it done.

internal/config/types.go
amol-: If you were running the latest Connect checkout it should have deployed yes.
maybe @Lytol notices anything that I'm missing

amol-:
If our purpose is only to use the comments to "preview" the user the configuration we initially detected, I think it would be a big deal if the comments are stripped away by any subsequent rewrite. At that point we already informed the user.
I agree it would be great to have those comments around always, but I think that even seeing them just the first time the .toml is generated is already a quality of life improvement.

In that case I guess we could have an additional step to the toml generation that prefixes those options with a `#`.
It's a bit of an hack, but it can be implemented isolated enough that it would add complexity to the code design
(just have a function like `commentPythonOptions` applied after the toml string is generated and before writing it)

internal/config/types.go
sagerb: Yes, I believe I was running against the latest Connect on `main`.

internal/schema/schemas/draft/posit-publishing-schema-v3.json
sagerb: Thanks for making this change.

sagerb:
The discussion of adding comments and our parser stripping away is not new, and I'd suggest, while it has come up here again, that we do not enlarge the scope of this PR by addressing it here. (existing issue: https://github.com/posit-dev/publisher/issues/2023)

Let's get this PR over the finish line, so that it can make the upcoming release. 

It looks to me that only this comment is unanswered? https://github.com/posit-dev/publisher/pull/2617#discussion_r2035848759

Is there anything else that you have pending for this PR?

amol-:
> It looks to me that only this comment is unanswered? [#2617 (comment)](https://github.com/posit-dev/publisher/pull/2617#discussion_r2035848759)
> 

Sorry, forgot to reply to that one.
I tried to explain the design decision and why I think it makes sense to keep that part separated from the interpreter itself.

> Is there anything else that you have pending for this PR?
>

Nope, it should be done.
I think we only have to wait for @Lytol to confirm on https://github.com/posit-dev/publisher/pull/2617#discussion_r2044751349 to double check that the manifest I'm generating is actually the expected one.

internal/config/types.go
jonkeane: Hmmm. Do you have an example of this being an issue @sagerb ? 

I just tried with the following on dogfood, and [it worked for me](https://rsc.radixu.com/connect/#/apps/0767dce4-2210-414a-816b-636bbc294018/logs?logKey=Tj5OmvEaP2xC1u1d) using the [stock-dashboard-python jumpstart](https://github.com/posit-dev/connect/tree/main/examples/stock-dashboard-python):

```
"python": {
    "version": "3.11.8",
    "package_manager": {
      "name": "pip",
      "package_file": "requirements.txt",
      "allow_uv": true
    }
  },
  "environment": {
		"image": "",
		"prebuilt": false,
		"python": {
			"requires": "<4.0"
		}
	},
```

The `python.version:3.11.8` which isn't available on the server, but it followed `environment.python.requires: <4.0` and is running with Python version 3.12.4. Which is what PEP440 says (and was what I expected given the backend implementation. 

That seems correct to me. But maybe I'm missing something? What were you seeing? 

[full manifest.json](https://github.com/user-attachments/files/19763644/manifest.json)

sagerb: @jonkeane Yes, you are seeing what I expected to see. I will try it again! Thanks!

PR #2618, opened 2025-04-08
Handle workspace trust states

### LINKED ISSUE #2611: Publisher Icon is often missing when workspace folder starts out untrusted.
From Trevor Nederlof:

I have noticed when using VS Code on PWB sometimes the publisher icon fails to show up when I open a folder with a Python app (Streamlit and Shiny for Python). The weird part is sometimes it works fine with the exact same workflow. I recorded a video showing the following:
- Launch a VS Code session (Publisher icon shows up)
- Open a Streamlit app (Publisher icon does not show up)
- Open a Shiny for Python app (Publisher icon does not show up)
- Launch a new VS Code session (Publisher icon shows up)
- Open a Streamlit App (Publisher icon does show up!)

Video showing behavior: https://www.loom.com/share/b3153497fb484e52b188c290b05f6c1a?sid=2f905741-a964-49d2-98c5-3234c7b97c54

Follow up questions:

Q: did you open the same directory in the second and last bullets there? 
This is only a suspicion, but watching the video, the publisher extension looks like each time the publisher extension doesn't load it's just after you had to mark as trusted the folder pop up happened.
But then when it did load that popup (presumably already been clicked and) didn't show up.

A: Yes it was the same directory each time. Hmmm the trusted box only pops up seemingly randomly on Workbench, maybe there is a pattern though, I have yet to find a way to permanently get rid of it

Q: Just to help debug: The Publisher icon on the left should always load when the extension is installed even if no folder is open. The content of the workspace shouldn’t matter at all.
There is an Output view (you can access it with the command palette command Output: Show Output Channels and select Posit Publisher). If you encounter unexpected behavior getting the output is super helpful for us.
In this case if the extension isn’t even activating we may not have that output channel which is also helpful to know.
I’m wondering if we need to handle the potential un-trusted state better as well, that seemed to be a pattern, but its unclear if that is it for sure.
Also thanks for letting us know, and for supplying the video. This is super helpful!

A: Additional output doesn't seem to have provided.

Reported by others as well.

sagerb:
## Intent

This PR provides the publisher extension with the functionality to support running within Restricted Mode. When in restricted mode, all publishing functionality is hidden, with only a button made available to manage the trust of the workspace (which is common with other extensions).

If and when the workspace is trusted by the user, the extension will commence its initialization. If the workspace is untrusted once again, the publisher returns to its restricted view.

Resolves #2611 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

At activation, the code will check if it is being activated in trusted mode or not. If it is, it proceeds with initialization. If restricted mode is active instead, it registers an event handler for when trust is granted to proceed with initialization.

The views and commands configured within `package.json` have been updated to use the context that VSCode API provides (`isWorkspaceTrusted`)

When in restricted mode, clicking on the publisher icon will show this view:

Screenshot 2025-04-08 at 4 04 31 PM

## User Impact

User should see no difference in functionality for trusted folders. In restricted mode, they will now see an experience much more common to what other extensions do.

## Directions for Reviewers

Launch VSCode with the extension and open a folder which you have not trusted. If you need to force one to be untrusted, you can use the pallete command of "Workspaces: Manage Workspace Trust" and untrust the open workspace.

Then confirm that you only see the view described above. Confirm that the button will bring up the trust management page, and click trust. Confirm that the publisher extension initializes and is functional. You can then untrust the workspace and confirm that the extension switches back to the restricted mode.

Confirm that you see no difference from before if you open a trusted workspace/folder.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- The Publisher extension will activate, but not initialize when the VSCode operates in Restricted Mode.
+  Initialization will commence when/if the workspace is trusted (which can be done via a button within the
+  publisher's view).
+
 ## [1.10.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- The Publisher extension will activate, but not initialize when the VSCode operates in Restricted Mode.
-  Initialization will commence when/if the workspace is trusted (which can be done via a button within the
-  publisher's view).
+- Improved the extension's behavior in VS Code workspace restricted mode.
+  The extension will activate, but prompt to manage workspace trust so it can initialize.
 
 ## [1.10.0]
 
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -19,8 +19,8 @@
   },
   "capabilities": {
     "untrustedWorkspaces": {
-      "supported": false,
-      "description": "This extension does not support untrusted workspaces because it is not safe to deploy untrusted code."
+      "supported": "limited",
+      "description": "This extension is not functional in Restricted Mode, but will recognize the change from restricted to trusted."
     }
   },
   "categories": [
@@ -232,40 +232,40 @@
         {
           "command": "posit.publisher.deployWithEntrypoint",
           "group": "navigation",
-          "when": "posit.publish.activeFileEntrypoint == true"
+          "when": "posit.publish.activeFileEntrypoint == true && isWorkspaceTrusted"
         }
       ],
       "view/title": [
         {
           "command": "posit.publisher.files.refresh",
-          "when": "view == posit.publisher.files",
+          "when": "view == posit.publisher.files && isWorkspaceTrusted",
           "group": "navigation@2"
         },
         {
           "command": "posit.publisher.pythonPackages.edit",
-          "when": "view == posit.publisher.pythonPackages",
+          "when": "view == posit.publisher.pythonPackages && isWorkspaceTrusted",
           "group": "navigation@1"
         },
         {
           "command": "posit.publisher.pythonPackages.refresh",
-          "when": "view == posit.publisher.pythonPackages",
+          "when": "view == posit.publisher.pythonPackages && isWorkspaceTrusted",
           "group": "navigation@2"
         },
         {
           "command": "posit.publisher.pythonPackages.scan",
-          "when": "view == posit.publisher.pythonPackages",
+          "when": "view == posit.publisher.pythonPackages && isWorkspaceTrusted",
           "group": "navigation@3"
         },
         {
           "command": "posit.publisher.homeView.refresh",
-          "when": "view == posit.publisher.homeView",
+          "when": "view == posit.publisher.homeView && isWorkspaceTrusted",
           "group": "navigation@1"
         }
       ],
       "commandPalette": [
         {
           "command": "posit.publisher.init-project",
-          "when": "workbenchState == folder  && posit.publish.state == 'initialized'"
+          "when": "workbenchState == folder && posit.publish.state == 'initialized' && isWorkspaceTrusted"
         },
         {
           "command": "posit.publisher.homeView.deleteCredential",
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- The Publisher extension will activate, but not initialize when the VSCode operates in Restricted Mode.
+  Initialization will commence when/if the workspace is trusted (which can be done via a button within the
+  publisher's view).
```
dotNomad, line 14: Just some word smithing suggestions:

> Improved the extension's behavior in VS Code workspace restricted mode. The extension will activate, but prompt to manage workspace trust so it can initialize.

extensions/vscode/package.json
```diff
@@ -343,49 +343,49 @@
         },
         {
           "command": "posit.publisher.homeView.associateDeployment",
-          "when": "webviewId == 'posit.publisher.homeView' && webviewSection == 'even-easier-deploy-more-menu-matching-configs' && posit.publish.selection.hasCredentialMatch == 'true' && posit.publish.selection.isPreContentRecord == 'false'"
+          "when": "webviewId == 'posit.publisher.homeView' && webviewSection == 'even-easier-deploy-more-menu-matching-configs' && posit.publish.selection.hasCredentialMatch == 'true' && posit.publish.selection.isPreContentRecord == 'false' "
```
dotNomad, line 346: Accidental space here? There are a few of these in the file - wonder where those came from.

extensions/vscode/package.json
```diff
@@ -441,25 +449,30 @@
           "name": "Log",
           "contextualTitle": "Log",
           "icon": "$(output)",
-          "when": "workbenchState == folder && posit.publish.state == 'initialized'"
+          "when": "!isWorkspaceTrusted"
```
dotNomad, line 452: This looks like a mistake. It looks like this will need to change due to the work you just merged to `main` though.

extensions/vscode/src/extension.ts
```diff
@@ -151,6 +148,32 @@ export async function activate(context: ExtensionContext) {
...
+  // We are not trusted yet... so if and when we are, then start the initialization
+  context.subscriptions.push(
+    workspace.onDidGrantWorkspaceTrust(() => {
+      console.log("Trust mode granted by user.");
+      console.log(
+        "Proceeding forward with initializion for extension within a trusted workspace",
+      );
+      initializeExtension(context);
+    }),
+  );
```
dotNomad, line 174: Do we need to dispose of this listener when the workspace is granted trust?

extensions/vscode/src/extension.ts
```diff
@@ -151,6 +148,32 @@ export async function activate(context: ExtensionContext) {
...
+    initializeExtension(context);
+    return;
+  }
+  console.log(
+    "activated within a restricted workspace. Initialization is paused until trust is granted.",
+  );
+
+  // We are not trusted yet... so if and when we are, then start the initialization
+  context.subscriptions.push(
+    workspace.onDidGrantWorkspaceTrust(() => {
```
dotNomad, line 167: I'm not seeing any subscription listener for untrusting a workspace, but when I test that in the extension it looks like it works as expected. I'm guessing this is because VS Code re-activates the extension?

dotNomad (commented):
Tested locally and this is working great.

I had a few questions for clarification. I can take another look when the merge conflicts are fixed - it looks like some adjustments will need to be made to the `package.json` to get this to play nice with #2612

extensions/vscode/src/extension.ts
```diff
@@ -151,6 +148,32 @@ export async function activate(context: ExtensionContext) {
...
+      console.log("Trust mode granted by user.");
+      console.log(
+        "Proceeding forward with initializion for extension within a trusted workspace",
+      );
+      initializeExtension(context);
```
jonkeane, line 172: Does this mean that users won't even need to reload the extension once trust is granted to get the full functionality?

extensions/vscode/src/extension.ts
sagerb: That's what I was seeing.

extensions/vscode/src/extension.ts
sagerb: No, it appears to only be happening once. As you saw below, VSCode reloads the extensions when trust is taken away.

extensions/vscode/src/extension.ts
sagerb: Correct! I was bouncing back and forth and our view was updating correctly. To try this, just bring up the Workspace Trust Settings (via the button) and then go between trust and untrust.

extensions/vscode/package.json
sagerb: Great catch! Thanks!

dotNomad (approved):
This looks great. Thank you for taking the time to dig into my questions.

PR #2626, opened 2025-04-22
Release 1.12.0 prep

sagerb:
## Intent

Preparing for Publisher Release 1.12.0

## Approach

Followed instructions within `/CONTRIBUTING.md` section `Before Releasing`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,22 +5,60 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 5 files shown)


marcosnav:
Looks good!

Just to note that there is upcoming work, just like `"requires_python"`, but for R. @amol- is working on it and it'd be great if both pieces can land in the same release since it is the same feature and feels a bit incomplete without R. 

Do you think this release can wait a bit until the R portion lands?

jonkeane:
Our release process is quick enough (so long as we don't bundle too much and it becomes onerous to write out the change long), that we should ship what we have now and then ship what lands next, next. We can cut a release next week or even later this week if we need to

sagerb:
@marcosnav Given Jon's response above, can I get an approval on the PR?

marcosnav (approved):
🚀

PR #2632, opened 2025-04-28
snowflake support (MVP)

christierney:
## Intent

MVP allowing Publisher to deploy to Connect running in Snowflake SPCS.

We cannot use Connect API Keys in this scenario. Instead, the user must have Snowflake connections configured on their host, which the Publisher will then use to authenticate requests to Connect.

The MVP only supports key pair authentication, meaning you must have a connections.toml with a `private_key_file`, `account`, and `user` defined.

Public documentation is not included in this PR as SPCS support is not being publicized yet.

Fixes posit-dev/connect#30701

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Modify "new credential" flow:
- "input server URL" sends a test request to the backend; the response now includes "server type"
- if "server type" is "snowflake", the next step is "input snowflake connection"; otherwise existing "input API key" is used
- "input snowflake connection" asks the backend to return available snowflake connections (must be already configured on host)
  - connections are tested against the server URL
  - if only 1 option, pick that option
  - if > 1, present quickpick
  - if 0, error
- server nickname step is unchanged

Modify "new deployment" flow: the optional "add credential" subflow is modified similar to above.

Add "snowflake connection" field to account/credential records. Either API Key or Snowflake Connection must be populated. This means there is now a new Credential version.

Add snowflake server type detection. This also means I needed to modify the deployment schema to accept "snowflake" as a server type.

Add Snowflake authenticator that generates and applies auth headers to HTTP client requests.

Add `/api/snowflake-connections` endpoint to report snowflake connections that work against a given server url.

Add code to load snowflake connection configs.

## User Impact

Most users will not be impacted. The publishing flow will continue to work as before.

Users deploying to Connect in SPCS will now be able to use Publisher. The tool will autodetect when they are deploying to SPCS and do the right thing. Errors will include guidance and links to Snowflake documentation.

## Automated Tests

Updated/added unit tests for new and modified code.

## Directions for Reviewers

Both methods for adding a credential (as a standalone flow, or as part of adding a deployment" have been tested. Chaita and I have tested deployments to dogfood via API key (unchanged) and deployments to https://nagoiaib-duloftf-posit-software-pbc-dev.snowflakecomputing.app/ via key pair auth.

Manual testing notes here: https://docs.google.com/document/d/1Yy6nAveS3c2Qunaw9hMgh9715IrI0oAp5ugssSCnXTg/edit?tab=t.0

## Checklist


extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -33,13 +36,46 @@ export async function newCredential(
...
+      );
+      window.showErrorMessage(
+        `Unable to query Snowflake connections. ${summary}`,
+      );
+      throw error;
+    }
+
+    if (!connectionList.length) {
+      const msg = `No working Snowflake connections found for ${serverUrl}. Please configure a connection in your environment before creating a credential.`;
+      window.showErrorMessage(msg);
```
christierney, line 66: HELP: this error messages shows up as expected, but there is also a separate "Internal Error" message. Not sure how to correctly handle this error.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -261,6 +305,57 @@ export async function newCredential(
...
+        buttons: [],
+        shouldResume: () => Promise.resolve(false),
+        ignoreFocusOut: true,
+      });
+
+      if (!pick || !isQuickPickItemWithSnowflakeConnection(pick)) {
+        return;
+      }
+
+      state.data.apiKey = pick.label; // FIXME: new prop for connection name
```
christierney, line 347: HELP: state has url, name, and apiKey, which are sent to the backed to create a new credential. Right now I'm reusing apiKey to store the Snowflake connection name, as a temporary hack. What's the right approach here?

We will always need to save _either_ an API Key (normal Connect) OR a Snowflake connection name (SPCS), never both at once.

Options:
- add a new field, snowflakeConnectionName. Either this or apiKey must be non-empty. One will always be empty. Potentially more confusing validation code, but very clear.
- keep using one field, but give it a more generic name like "key" or "authData" or something. Simpler validation logic (just keep what we have now, i.e. check that the field is populated.)
- keep the hack. Use the field named apiKey, but it's not always an apiKey. Super simple. Kinda gross 🤢

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -174,11 +214,15 @@ export async function newCredential(
...
+
+    if (accountType === "snowflake") {
+      return (input: MultiStepInput) => inputSnowflakeConnection(input, state);
+    }
     return (input: MultiStepInput) => inputAPIKey(input, state);
   }
 
   // ***************************************************************
-  // Step #2:
+  // Step #2, option A:
```
sagerb, line 225: I've given up on step numbers. Feel free to update this and 2b to indicate what kind of input this step is prompting for.

sagerb: The existing code uses a pattern similar to a filter. It simply decides if it needs to do anything and, if not, passes the request on to the next step handler. 

This will dictate the need to check in each of the two steps if they need to be visible or not (with the snowflake connection input checking if there is an API key and skipping itself, and the API key prompt doing the same thing in reverse). 

I think you'll be able to heavily lean on evaluating value of `accountType` that you've established with the server URL testing.  Don't forget the user can go both forward and backward through this flow, and change earlier values which must be accounted for when the step's visibility is evaluated.

extensions/vscode/src/multiStepInputs/newCredential.ts
sagerb: To record what we discussed on the zoom call... My recommendation is to add an additional field for snowflake connection name into the credential type, which should allow for clear understanding of the use. Don't forget, moving forward in this manner may require updates to the serialization of the credential into the key store and the credential file.

This also ties into my earlier comments of moving the snowflake connection stuff into credentials as well.

sagerb (commented):
@christierney this is a great start, and I want to say that you did an excellent job of just diving in to figure this out. Well done!

Most of my feedback tries to pull back some of the complexity that you probably don't need and moving a few things around. But in general, the approach is solid and I especially like how there is no separate prompt required for the user to select that they want a snowflake connection vs. a connect server API key.

I'm also going to ask @dotNomad to provide some input here.

Please feel free to reach out if you have any questions!

extensions/vscode/src/multiStepInputs/newCredential.ts
dotNomad: Adding a new field does mean we have to know a bit more about the credentials in the webview / extension code but that shouldn't be a problem as we can determine which credential type it is based on that new field.

extensions/vscode/src/api/types/credentials.ts
```diff
@@ -20,5 +20,6 @@ export type CredentialUser = {
 export type TestResult = {
   user: CredentialUser | null;
   url: string | null;
+  accountType: string | null;
```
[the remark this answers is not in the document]
dotNomad: Definitely agree with this. We have the three auth types in the backend and we are doing checks to see if `accountType === "snowflake"` in this PR.  An enum would be very helpful. There are some examples in `extensions/vscode/src/api/types/contentRecords.ts`

extensions/vscode/src/api/client.ts
```diff
@@ -10,6 +10,7 @@ import { Interpreters } from "./resources/Interpreters";
 import { Packages } from "./resources/Packages";
 import { Secrets } from "./resources/Secrets";
 import { EntryPoints } from "./resources/Entrypoints";
+import { SnowflakeConnections } from "./resources/SnowflakeConnections";
```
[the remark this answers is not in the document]
dotNomad: I'd keep it a separate resource since the API endpoint is separate. If the snowflake connections were returned in the Credentials API then I'd suggest moving them into the same file.

If the snowflake connections end up being the same as the credentials object and used the same logic, endpoint, etc then moving them into the same resource makes the most sense - with good typing to distinguish between the results of getting all credentials.

notes.md
dotNomad, line 1: It looks like this may have been accidentally committed?

dotNomad (commented):
I really like the flow that you've laid out in your approach. It matches a lot of the same patterns we already have for selecting if there is only 1 option.

extensions/vscode/src/api/client.ts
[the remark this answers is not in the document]
christierney: Since snowflake connections are managed separately (via the standard Snowflake config files) I think it makes sense to keep separate. The only link to Credentials is that we need to store a connection name in the creds, but we specifically don't want to take over managing all the Snowflake connection data.

internal/accounts/account.go
```diff
@@ -3,20 +3,31 @@ package accounts
...
-	ApiKey      string          `json:"-"`            // For Connect servers
+	ServerType          ServerType    `json:"type"`         // Which type of API this server provides
+	Source              AccountSource `json:"source"`       // Source of the saved server configuration
+	Name                string        `json:"name"`         // Nickname
+	URL                 string        `json:"url"`          // Server URL, e.g. https://connect.example.com/rsc
+	Insecure            bool          `json:"insecure"`     // Skip https server verification
+	Certificate         string        `json:"-"`            // Root CA certificate, if server cert is signed by a private CA
+	AccountName         string        `json:"account_name"` // Username, if known
+	ApiKey              string        `json:"-"`            // For Connect servers
+	SnowflakeConnection string        `json:"-"`
```
zackverham, line 14: I like having the doc comments for each of these fields - could be nice to include one here too

internal/credentials/file_test.go
```diff
@@ -137,28 +137,39 @@ func (s *FileCredentialsServiceSuite) TestLoadFile() {
...
+				Version:             1,
+				URL:                 "https://c4.connect-server:3939/connect",
+				ApiKey:              "abcdeC2aqbh7dg8TO43XPu7r56YDh004",
+				SnowflakeConnection: "",
+			},
+			"jonsnow": {
+				GUID:                "4b080144-bbea-4bf6-9819-4b0ec3546f42",
+				Version:             1,
+				URL:                 "https://example.snowflakecomputing.app/connect",
+				ApiKey:              "",
```
zackverham, line 171: for my own understanding - what happens if we have a credentials file that has both an api key and a snowflake connection set? Is that an error condition, and if so, how does it get surfaced?

internal/api_client/auth/snowflake.go
```diff
@@ -0,0 +1,152 @@
+package auth
+
+// Copyright (C) 2025 by Posit Software, PBC.
```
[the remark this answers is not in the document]
christierney: Frankly I am not sure how to test their weird off-spec OAuth implementation other than manually... I should probably take another crack at writing _something_ though. (Maybe this would be a good reason to move the JWT stuff, into an interface that I can stub.)

internal/credentials/file_test.go
christierney: I think all that would happen is we would use the API key and ignore the snowflake connection, due to the order of the cases in `auth.go`. It would indicate a developer error. The existing validation of credentials seems a little scattered to me but it might be good to guard against this somewhere, yeah.

internal/api_client/auth/snowflake.go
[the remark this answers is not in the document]
christierney: (there's also no other tests in this package, which is not a good excuse though!)

internal/accounts/account.go
mconflitti-pbc: Likely a bigger change than necessary but just a thought. Remove both fields and create a new one called Credential that is a string and then have another field for the CredentialType. It could simplify some things later on perhaps. Should always be strings since this is coming from input from the UI (i think).

internal/api_client/auth/snowflake.go
```diff
@@ -0,0 +1,152 @@
...
+		scope = fmt.Sprintf("session:role:%s %s", role, ingressURL)
+	}
+
+	param := url.Values{
+		"grant_type": []string{"urn:ietf:params:oauth:grant-type:jwt-bearer"},
+		"scope":      []string{scope},
+		"assertion":  []string{signedToken},
+	}
+
+	snowflakeTokenEndpoint := fmt.Sprintf("https://%s.snowflakecomputing.com/oauth/token", account)
```
mconflitti-pbc, line 134: i dont think this will ever change. again could be affected by privatelink work cc: @Rikagx

internal/credentials/file_test.go
mconflitti-pbc: re: my comment before about potentially refactoring to a single credential field.

mconflitti-pbc (approved):
Overall looks great! Couple comments especially related to concurrent work for snowflake privatelink that could possible affect workbench. Otherwise, looks like it is in working order based on the various tests we all did. Great work on all of this!

extensions/vscode/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `DESCRIPTION` file and `renv.lock` file. The detected version
   will fill the `requires_r` field in configuration (#2636)
 
+- Introduced experimental support for deploying to SPCS. (#2632)
```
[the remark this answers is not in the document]
christierney: I wasn't sure if I should mention it at all before we went public with the feature! But yeah I'll update.

internal/credentials/file_test.go
christierney: we did discuss that possibility (https://github.com/posit-dev/publisher/pull/2632#discussion_r2066896115, sorry I had marked it resolved) and decided separate fields made the purpose more clear, though I do think there are pros/cons both ways!

internal/api_client/auth/snowflake.go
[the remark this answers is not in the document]
christierney: added tests.

internal/api_client/auth/snowflake.go
```diff
@@ -0,0 +1,152 @@
+package auth
```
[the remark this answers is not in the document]
christierney: I ended up moving these in the course of writing tests. I kept the general shape the same to make it easy to switch to https://github.com/rstudio/snowflake-lib/ in the future.

extensions/vscode/CHANGELOG.md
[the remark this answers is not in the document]
christierney: Updated.

internal/accounts/server_type.go
```diff
@@ -25,17 +30,26 @@ func (t ServerType) Description() string {
...
+		return "", err
+	}
+	host := u.Hostname()
+	if strings.HasSuffix(host, ".posit.cloud") {
+		return ServerTypeCloud, nil
+	} else if strings.HasSuffix(host, ".rstudio.cloud") {
+		return ServerTypeCloud, nil
+	} else if strings.HasSuffix(host, ".shinyapps.io") {
+		return ServerTypeShinyappsIO, nil
+	} else if strings.HasSuffix(host, ".snowflakecomputing.app") {
```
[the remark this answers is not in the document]
christierney: Added detection for privatelink.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -33,13 +36,46 @@ export async function newCredential(
   const api = await useApi();
   let credentials: Credential[] = [];
 
+  let serverType: ServerType;
+  let connections: SnowflakeConnection[] = [];
+  let connectionQuickPicks: QuickPickItemWithIndex[];
+
+  const getSnowflakeConnections = async (serverUrl: string) => {
```
marcosnav, line 43: This method is exactly the same as one added in `newDeployment.ts`, it would be good to have it pulled out to be reused.

Sorry that there are some areas that look like dumping grounds and here I'm asking you to change to a modular approach. But it would help to start doing it.

WDYT about having a new `snowflake.ts` file in the same dir with this `getSnowflakeConnections` method in it, to be imported and reused, also that would help you to have unit tests for it without having to deal with all that's happening in this file.

Also sorry to ask you to create tests in a mostly un-tested code base but promoting it would also help a lot 🙇 

For examples on mocking you can take a look at `deployHandlers.test.ts` and `state.test.ts`

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -174,96 +214,159 @@ export async function newCredential(
...
+  // Step: Enter the API Key
   // ***************************************************************
   async function inputAPIKey(input: MultiStepInput, state: MultiStepState) {
-    const thisStepNumber = assignStep(state, "inputAPIKey");
-    const currentAPIKey =
-      typeof state.data.apiKey === "string" && state.data.apiKey.length
-        ? state.data.apiKey
-        : "";
-    let validatedURL = "";
+    if (serverType !== ServerType.SNOWFLAKE) {
```
marcosnav, line 225: I think it might be more appropriate to do this check and use the proper method above in line 218.

```js
if (serverType !== ServerType.SNOWFLAKE) {
    return (input: MultiStepInput) => inputSnowflakeConnection(input, state);
}
return (input: MultiStepInput) => inputAPIKey(input, state);
```

This change would also simplify checks below.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -174,96 +214,159 @@ export async function newCredential(
...
+      });
+
+      // only one of api key and snowflake connection should be configured
+      state.data.apiKey = apiKey;
+      state.data.snowflakeConnection = "";
+      state.data.url = validatedURL;
+      state.lastStep = thisStepNumber;
+      return (input: MultiStepInput) => inputSnowflakeConnection(input, state);
+    }
+    return inputSnowflakeConnection(input, state);
```
marcosnav, line 306: This is a bit odd, calling `inputSnowflakeConnection` regardless of the serverType check. With the suggestion above this method can stay as it is.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -174,96 +214,159 @@ export async function newCredential(
...
+  }
+
+  // ***************************************************************
+  // Step: Enter the Snowflake connection name
+  // ***************************************************************
+  async function inputSnowflakeConnection(
+    input: MultiStepInput,
+    state: MultiStepState,
+  ) {
+    if (serverType === ServerType.SNOWFLAKE) {
```
marcosnav, line 316: With the suggestion above this check is done in advance and not necessary here.

marcosnav (commented):
Sorry, I'll be annoying with this one 🙇 

I haven't finished reviewing but already have some comments to work on.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -630,7 +672,7 @@ export async function newDeployment(
   // Step: New Credentials - Enter the API Key
   // ***************************************************************
   async function inputAPIKey(input: MultiStepInput, state: MultiStepState) {
-    if (newCredentialByAnyMeans()) {
+    if (newCredentialByAnyMeans() && serverType !== ServerType.SNOWFLAKE) {
```
marcosnav, line 675: Same observation here as in `newCredential`, can this checks be done earlier and determine which method from `inputAPIKey` or `inputSnowflakeConnection` should be called. You could also benefit from returning early, instead of having big `if` blocks.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `DESCRIPTION` file and `renv.lock` file. The detected version
   will fill the `requires_r` field in configuration (#2636)
 
+- Added support for deploying to Connect in Snowflake using key-pair auth. (#2632)
+
```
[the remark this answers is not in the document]
christierney: doh! good catch, thanks

extensions/vscode/src/multiStepInputs/newCredential.ts
christierney: To expand a bit: I'm following the style used in `newDeployment.ts` where every step gets called but some of them get skipped based on various filters.

marcosnav: Just added a comment about `newDeployment.ts` too. I'm not fully on board with this pattern because snowflake and API Keys are different types of auth types and concepts, and I find it confusing that API Key related methods call Snowflake related methods and viceversa. I'd vote on re-working the pattern for clarity if needed.

christierney: I think the idea is that, as you can go backwards as well as forwards in the multistep flows, each possible step gets run along with a filter to skip it as needed. For example I might start adding a non-snowflake host, then hit the back button and change to a snowflake host.

marcosnav:
@christierney Finished taking a look. This is in general great and thorough work! 

I have some concerns, already raised. I understand that there are time constraints so I'll approve for you to have a clear path to present the demo. Still it'd be great to have follow up issues to address the comments. 

It is a bit embarrassing for me to ask of you to follow better or different practices in areas where we don't have them, but that's something we want to do and we have to start doing it and enforcing it, otherwise it'll never happen. 🙇

marcosnav (approved):

extensions/vscode/src/multiStepInputs/newCredential.ts
sagerb: Yes, @christierney, that is the pattern implemented in the multi-steps. They are a cascading list of possible steps, which dynamically have the responsibility of deciding if they need to be shown or if not, simply call the next step without adding anything to the promise stack. This is required in order to allow the dynamic branching you describe as a user changes a prompt along the way, which inserts, eliminates, or completely branches into other steps. The infrastructure can absolutely be improved, but that is a large undertaking. I would recommend proceeding with your current changes.

extensions/vscode/src/multiStepInputs/newCredential.ts
sagerb: In the existing pattern, each step needs to implement the logic to suppress itself, and shouldn't rely upon previous steps to determine it.

extensions/vscode/src/multiStepInputs/newCredential.ts
sagerb: In the existing pattern, each step needs to implement the logic to suppress itself, and shouldn't rely upon previous steps to determine it.

extensions/vscode/src/multiStepInputs/newCredential.ts
christierney: filed #2650 to track this

christierney:
I'm going to merge this due to time constraints but I've assigned myself #2650.

Issue #2625, opened 2025-04-17
Issue with initial set up - error saying failed to connect to api


iainwallacebms:
Hi,

I get an error when trying to create a new deployment (i.e. the first step) when using the latest publisher extension with the latest version of Positron on my companies Mac.

"Unable to connect with a failed api response"

<img width="1715" alt="Image" src="https://github.com/user-attachments/assets/1ebe9757-7463-470b-a60a-76fc463888cb" />

Any suggestions to trouble shoot greatly appreciated.

Thanks

Iain

sagerb:
Hi @iainwallacebms.

Can you provide the log text from the output pane after selecting Posit Publisher? This might help us narrow down the problem you are encountering.

While it shouldn't be a problem, do you know what chip architecture your Mac is? (We run this extension regularly on both Apple and Intel silicon Macs.)

iainwallacebms:
Hi,

It is an M1 Mac. Below are the screenshots of the output pane for the window and Posit publisher after pressing the add new deployment. Is there somewhere that I need to set up a connection to our Connect server before I start? https://report.pri.bms.com/connect/

Thanks

Iain

<img width="976" alt="Image" src="https://github.com/user-attachments/assets/5de6ea7f-5690-424f-8057-565927ed8534" />
<img width="849" alt="Image" src="https://github.com/user-attachments/assets/263f8572-6a2a-44c1-86f3-7dc96622cf7c" />

sagerb:
This issue is the same as #2609, although I won't be closing as duplicate.

iainwallacebms:
Thanks - I tried uninstall and reinstalling it as per the other issue, but still have the same issue.

sagerb:
Thank you for trying that @iainwallacebms.

Can you please grab the URL address listed in the output pane and try the following command at a terminal (replacing the `127.0.0.1:61079` with that address)? 

```
curl http://127.0.0.1:61079/api/credentials 
```

It should be an empty array if I understand your installation status. 

Here's my sample output, which shows what I see with credentials

```
 > curl http://127.0.0.1:61079/api/credentials 
[{"guid":"46e6994e-53be-4f2e-ac30-2cf47e5419c2","name":"local","url":"https://connect.localtest.me/rsc/dev-password","apiKey":"xxxxxxxxxxxxxxxxxxxxxxx"}]
```

If that doesn't work, could you try replacing the `127.0.0.1` with `localhost`, and retrying?

[...]

PR #2635, opened 2025-05-05
Parse R requirements from Depends: and renv.lock and improve .python-version parsing.

### LINKED ISSUE #2628: .python-version is not necessarily a version specifier (per PEP 440) and thus cannot necessarily be used for `environment.python.requires`
See <https://github.com/posit-dev/rsconnect-python/issues/659>

### LINKED ISSUE #2636: Publisher should set the R version constraint in the following order
* DESCRIPTION file (i.e. Depends: R (>= 3.5.0)
* renv.lock specifies R -> Version (debatable, we should discuss)

The version constraint should be set within `environment` -> `r` -> `requires` in the `manifest.json`.  If there is no version constraint found, do not specify anything in `environment` -> `r` -> `requires`. Either way, we should continue to populate `platform` with the "current" R version.

Example `manifest.json`:

```
{
  "version": 1,
  "locale": "en_US",
  "platform": "4.3.1",
  "metadata": {
    "appmode": "api",
    "primary_rmd": null,
    "primary_html": null,
    "content_category": null,
    "has_parameters": false
  },
  "environment": {
    "r": {
      "requires": ">=4.2.0, <5.0"
    }
  },
  "packages": {
    ...
  }
}
```

amol-:
## Intent

Add support for detecting R interpreter version requirement like https://github.com/posit-dev/publisher/pull/2617 did for Python.

Also improve the Python support for cases where the `.python-version` file does not specify any constraint operator. In those cases we use `~=` to provide the closest available version on Connect.

- fixes #2628
- closes #2636

## Type of Change

- - [x] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

A `r_requires` equivalent to `python_requires` is added and used by the Manifest generation code to detect the project requested interpreter version.

## User Impact

If the user had provided a interpreter version constraint in any of the 2 supported files, the constraint will now be propagated to Connect.

## Automated Tests

- [x] Added a suite of tests for RProjectRRequires
- [x] Added an test for a newly generated manifest out of a project with requirements

## Checklist


internal/config/types.go
```diff
@@ -158,9 +158,10 @@ func (p *Python) FillDefaults(
 }
 
 type R struct {
-	Version        string `toml:"version,omitempty" json:"version"`
-	PackageFile    string `toml:"package_file,omitempty" json:"packageFile"`
-	PackageManager string `toml:"package_manager,omitempty" json:"packageManager"`
+	Version          string `toml:"version,omitempty" json:"version"`
+	PackageFile      string `toml:"package_file,omitempty" json:"packageFile"`
+	PackageManager   string `toml:"package_manager,omitempty" json:"packageManager"`
+	RequiresRVersion string `toml:"requires_r_version,omitempty" json:"requiresRVersion"`
```
Lytol, line 164: Ultra nit: since we're already within an `R` struct, and golang tends to favor terse variable naming in general, what about just `Requires`?

internal/bundles/manifest.go
```diff
@@ -171,6 +176,16 @@ func NewManifestFromConfig(cfg *config.Config) *Manifest {
...
+		// If the configuration specifies a specific R version constraint
+		// (e.g. ">=3.8"), declare the environment requires that version.
+		if cfg.R.RequiresRVersion != "" {
+			if m.Environment == nil {
+				m.Environment = &Environment{}
+			}
+			m.Environment.R = &EnvironmentR{
+				Requires: cfg.R.RequiresRVersion,
+			}
+		}
```
marcosnav, line 188: It'd be nice to have a `NewManifestFromConfig` unit test that includes `Environment` being populated with this info

internal/inspect/r.go
```diff
@@ -78,10 +78,18 @@ func (i *defaultRInspector) InspectR() (*config.R, error) {
 		i.log.Debug("Error retrieving R package lock file", "error", err)
 	}
 
+	rProjectRequires := interpreters.NewRProjectRRequires(i.base)
+	r_requires, err := rProjectRequires.GetRVersionRequirement()
```
marcosnav, line 82: nit: Use camelCase for `r_requires`

internal/interpreters/python_requires_test.go
```diff
@@ -59,5 +59,5 @@ func (s *PythonRequiresSuite) TestGetPythonRequiresPythonVersion() {
 	pythonRequires, err := pyRequires.GetPythonVersionRequirement()
 	s.NoError(err)
 	s.NotEmpty(pythonRequires)
-	s.Equal(">=3.8, <3.12", pythonRequires)
+	s.Equal(">=3.8,<3.12", pythonRequires)
```
marcosnav, line 62: It'd be nice to also include tests as you did for `r_requires.go` where a plain `3.8` is set as `~=3.8`, maybe test the error cases too

internal/interpreters/r.go
```diff
@@ -526,3 +527,13 @@ func (i *defaultRInterpreter) GetPackageManager() string {
 func (i *defaultRInterpreter) GetPreferredPath() string {
 	return i.preferredPath.String()
 }
+
+func (i *defaultRInterpreter) GetRRequires() string {
+	rProjectRequires := NewRProjectRRequires(i.base)
+	r_requires, err := rProjectRequires.GetRVersionRequirement()
```
marcosnav, line 533: same nit, camelCase

internal/schema/schemas/draft/posit-publishing-schema-v3.json
```diff
@@ -146,6 +146,12 @@
           "default": "renv",
           "description": "Package manager that will install the dependencies. If package-manager is none, dependencies will be assumed to be pre-installed on the server.",
           "examples": ["renv", "none"]
+        },
+        "requires_r_version": {
```
marcosnav, line 150: The python version of this is `requires_python` with no `..._version`, WDYT about making both consistent?

internal/interpreters/python_requires.go
```diff
@@ -110,3 +127,33 @@ func (p *PyProjectPythonRequires) readSetupCfg() (string, error) {
...
+
+	// Otherwise it should be a version string
+	if !validVersion.MatchString(constraint) {
+		return "", fmt.Errorf("invalid python version: %s", constraint)
+	}
+
+	// strip trailing zeros and dots
+	// e.g. 3.8.0 -> 3.8
+	// and prefix with ~=
+	trimmed := strings.TrimRight(constraint, "0.")
```
Lytol, line 157: What if I have a version like `3.11.10`?  I think I'd end up with `3.11.1` with this implementation.

Regardless, shouldn't the compatible release clause include the patch in order to match to major.minor? Meaning, if `3.8` is specified in the `.python-version`, we ultimately want `~=3.8.0` (which is equivalent to `>=3.8.0, ==3.8.*`), right?

internal/interpreters/r_requires.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Join the dependencies into a single string and look for R.
+	// The regex will match "R (ANYTHING)"
+	all := strings.Join(deps, " ")
+	re := regexp.MustCompile(`R\s*\(([^)]+)\)`)
+	match := re.FindStringSubmatch(all)
+	if len(match) >= 2 {
+		return match[1], nil
+	}
+	return "", nil
```
Lytol, line 75: I'm not sure that this acts as a true DCF parser (see <https://www.debian.org/doc/debian-policy/ch-controlfields.html>), but maybe it does... and even if it doesn't, maybe it's "good enough" if there are no off-the-shelf golang DCF parsers that we can use. :shrug:

internal/interpreters/python_requires.go
```diff
@@ -53,7 +54,23 @@ func (p *PyProjectPythonRequires) readPythonVersionFile() (string, error) {
...
+			continue
+		}
+		adaptedPart, err := adaptPythonRequires(part)
+		if err != nil {
+			return "", err
+		}
+		adapted = append(adapted, adaptedPart)
+	}
+
+	return strings.Join(adapted, ","), nil
```
Lytol, line 73: It would be really nice to see tests around the various cases that we expect and how they should be transformed.

Lytol (dismissed):
Generally, seems very reasonable and most of my comments are just nits (so take them or leave them, whatever you prefer).

One thing that I think we should address before approval though:

- I'm not sure the `adaptPythonRequires` is doing what we expect, both in terms of the way it parses versions (removing trailing zeros and periods), but also the expected output. It would be really nice to see some tests (table-driven?) on inputs and outputs to verify.

marcosnav:
@amol- In the description of this PR it says this fixes #2625, I think the proper one is #2628 am I right?

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.X.X]
```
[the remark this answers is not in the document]
amol-: Renamed 👍

internal/bundles/manifest.go
amol-: Added in `TestNewManifestFromConfigVersionRequirements`

internal/interpreters/r.go
amol-: 👍

internal/schema/schemas/draft/posit-publishing-schema-v3.json
amol-: removed `_version` 👍

internal/bundles/manifest.go
```diff
@@ -53,6 +53,10 @@ type Metadata struct {
 	HasParameters   bool            `json:"has_parameters,omitempty"`   // True if this is content allows parameter customization.
 }
 
+type EnvironmentR struct {
+	Requires string `json:"requires"` // The R version to use for the content environment
+}
+
 type EnvironmentPython struct {
 	PythonRequires string `json:"requires"` // The Python version to use for the content environment
 }
```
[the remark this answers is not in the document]
amol-: Renamed both to `Requires`

internal/interpreters/python_requires.go
amol-: Extended the testsuite with most cases in `TestGetPythonRequiresPythonVersionAdapt`

internal/interpreters/python_requires.go
amol-: You are right. The right call was meant to be `TrimSuffix` not `TrimRight` 🫣

Version numbers are always padded by zero according to PEP440, see

```
As with version matching, the release segment is zero padded as necessary to ensure the release segments are compared with the same length.
```

For that reason I think it's much more readable and clear the intent when the "piece you don't care about" (for lack of a better definition) is omitted.

internal/interpreters/python_requires_test.go
amol-: Added a comprehensive test suite in `TestGetPythonRequiresPythonVersionAdapt`

internal/interpreters/r_requires.go
amol-: I'm open to suggestions here, I'm not an expert on `DESCRIPTION` syntax.

The overall idea here is that we just look for `Depends:` and consider part of it anything that is indented with a `\t` or `" "` following it. The DCF specification states `"The lines after the first are called continuation lines and must start with a space or a tab"`.
Anything != "\t" or " " signals the end of file or a new section and thus ends the dependencies list.

To be "fully correct" we would need to split the `deps` on commas to convert them to a proper list, but in reality we don't care much for the dependencies, we only care if there is a `R (VERSION)` package in the list.

I create a `TestGetRRequiresDESCRIPTIONDepends` testsuite with some corner cases, let me know if you see any that it makes sense to add.

marcosnav (approved):
Thank you for the additional changes! Looks great!

PR #2637, opened 2025-05-07
Quote paths to support windows

### LINKED ISSUE #2598: R command problem in positron when publishing from Windows
I tried the publisher in Positron (2025-03.0+116) and it fails at renv scanning 
````
Could not scan R packages from lockfile: renv.lock, exit status 1. See [Troubleshooting docs](https://github.com/posit-dev/publisher/blob/main/docs/troubleshooting.md) for help.
````

Looking at the log I see this 
```
time=2025-03-11T18:10:44.855+01:00 level=DEBUG msg="Collecting manifest R packages" local_id=2wZVlqVUa0bAxIjh event_op=publish/getRPackageDescriptions lockfile=c:\Users\chris\Documents\DEV_OTHER\00-TESTS\test-quarto\renv.lock
time=2025-03-11T18:10:44.856+01:00 level=DEBUG msg="Running command" local_id=2wZVlqVUa0bAxIjh event_op=publish/getRPackageDescriptions cmd="C:\\Program Files\\R\\bin\\R.bat" args="-s -e cat(.libPaths(), sep=\"\\n\")"
time=2025-03-11T18:10:44.900+01:00 level=ERROR msg="Error running command" local_id=2wZVlqVUa0bAxIjh event_op=publish/getRPackageDescriptions command="C:\\Program Files\\R\\bin\\R.bat" error="exit status 1"
'C:\Program' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.
````

The interesting part is 
```
command="C:\\Program Files\\R\\bin\\R.bat" error="exit status 1"
'C:\Program' n’est pas reconnu en tant que commande interne
ou externe, un programme exécutable ou un fichier de commandes.
```

which is French because I work on French Windows. 

It means 
````
command="C:\\Program Files\\R\\bin\\R.bat" error="exit status 1"
'C:\Program' is not recognized as an internal
or external command, an executable program or a batch file.
````

and this is usually a quoting problem when running the command on Windows. 

I figured I should report here. If you prefer direct slack message in our Posit slack, tell me the channel to use. Thanks !

### PR DESCRIPTION
**Pull request by jonkeane**
## Intent

Added functional tests for calling into R from Windows (and particularly when R is installed by rig). Those confirmed the failure. Due to the [way quotes work in windows](https://learn.microsoft.com/en-gb/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way), when the R executable was a `.bat`, we needed to do more + different quoting than we already were for the scripts we shelled out to. This PR changes those to write the script portions out to files and then use `R... -f ...` to run them. This is more robust since we don't have to worry about quoting within the contents of the script.

Resolves #2598

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

_last sentence of intent_

## User Impact

Windows users who have installed rig are able to use the publisher.

## Automated Tests

Added functional tests that confirm the bug and assert that it is resolved.

## Directions for Reviewers

The first commit shows the test failures, and the second shows the code the resolves the issue. If you want to review the tests separately from the fix, that is the way.

## Checklist

Not yet, but will:

### DIFF
--- a/.github/workflows/agent.yaml
+++ b/.github/workflows/agent.yaml
@@ -13,6 +13,28 @@ jobs:
         with:
           fetch-depth: 0
       - uses: extractions/setup-just@v2
+
+      # Install R with rig to ensure we are compatible
+      - name: Install rig with apt
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          sudo curl -L https://rig.r-pkg.org/deb/rig.gpg -o /etc/apt/trusted.gpg.d/rig.gpg
+          sudo sh -c 'echo "deb http://rig.r-pkg.org/deb rig main" > /etc/apt/sources.list.d/rig.list'
+          sudo apt update
+          sudo apt install -y r-rig
+      - name: Install rig with brew
+        if: ${{ runner.os == 'macOS' }}
+        run: |
+          brew tap r-lib/rig
+          brew install --cask rig
+      - name: Install rig with Choco
+        if: ${{ runner.os == 'Windows' }}
+        run: choco install rig
+
+      - name: Install R with rig
+        run: rig add release
+      - name: Install renv package
+        run: Rscript -e "install.packages('renv')"
       - uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
--- a/.github/workflows/agent.yaml
+++ b/.github/workflows/agent.yaml
@@ -14,7 +14,9 @@ jobs:
           fetch-depth: 0
       - uses: extractions/setup-just@v2
 
-      # Install R with rig to ensure we are compatible
+      # Install R with rig for our funcitonal tests to ensure we are compatible with it.
+      # We aren't used r-lib/actions here becuase rig, especially on windows, installs in
+      # slightly different ways (e.g. using a batch file on windows)
       - name: Install rig with apt
         if: ${{ runner.os == 'Linux' }}
         run: |
@@ -38,4 +40,6 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
-      - run: just test
+      # Run tests without `-short` to run functional tests
+      # TODO: https://github.com/posit-dev/publisher/issues/2649
+      - run: just test ./...
--- a/internal/executor/executor.go
+++ b/internal/executor/executor.go
@@ -72,8 +72,7 @@ func createTempRScript(content string) (string, func(), error) {
 }
 
 func (e *defaultExecutor) RunScript(executable string, args []string, script string, cwd util.AbsolutePath, log logging.Logger) ([]byte, []byte, error) {
-	log.Debug("Running command", "cmd", executable, "args", strings.Join(args, " "), "script", script)
-
+	log.Debug("Running script", "cmd", executable, "args", strings.Join(args, " "), "script", script)
 	// Write script contents to a file, clean it up, and append to args
 	tempScriptPath, cleanup, scriptErr := createTempRScript(script)
 	if scriptErr != nil {
--- a/internal/executor/executor.go
+++ b/internal/executor/executor.go
@@ -43,55 +43,27 @@ func (e *defaultExecutor) RunCommand(executable string, args []string, cwd util.
 	return stdout.Bytes(), stderrBuf.Bytes(), err
 }
 
-// createTempRScript creates a temporary file with R code and returns the file path.
-// The caller is responsible for removing the temporary file.
-func createTempRScript(content string) (string, func(), error) {
+// RunScript is helpful when running R scripts, they avoid needing to ensure that the contents of the script
+// is fully escaped and quoted.
+func (e *defaultExecutor) RunScript(executable string, args []string, script string, cwd util.AbsolutePath, log logging.Logger) ([]byte, []byte, error) {
+	log.Debug("Running script", "cmd", executable, "args", strings.Join(args, " "), "script", script)
+	// Write script contents to a file, clean it up, and append to args
 	tempFile, err := os.CreateTemp("", "*")
-	if err != nil {
-		return "", nil, fmt.Errorf("failed to create temporary R script: %w", err)
-	}
 
-	// Write the R code to the temporary file
-	if _, err := tempFile.WriteString(content); err != nil {
-		tempFile.Close()
-		os.Remove(tempFile.Name())
-		return "", nil, fmt.Errorf("failed to write to temporary R script: %w", err)
-	}
-	if err := tempFile.Sync(); err != nil {
-		tempFile.Close()
-		os.Remove(tempFile.Name())
-		return "", nil, fmt.Errorf("failed to flush temporary R script to disk: %w", err)
-	}
-
-	// Return the file path and a cleanup function
+	// defer cleaning up
 	cleanup := func() {
 		tempFile.Close()
 		os.Remove(tempFile.Name())
 	}
-	return tempFile.Name(), cleanup, nil
-}
-
-func (e *defaultExecutor) RunScript(executable string, args []string, script string, cwd util.AbsolutePath, log logging.Logger) ([]byte, []byte, error) {
-	log.Debug("Running script", "cmd", executable, "args", strings.Join(args, " "), "script", script)
-	// Write script contents to a file, clean it up, and append to args
-	tempScriptPath, cleanup, scriptErr := createTempRScript(script)
-	if scriptErr != nil {
-		return nil, nil, scriptErr
-	}
 	defer cleanup()
-	args = append(args, "-f", tempScriptPath)
-
-	cmd := exec.Command(executable, args...)
-	cmd.Dir = cwd.String()
-	var stdout bytes.Buffer
-	cmd.Stdout = &stdout
-	stderrBuf := new(bytes.Buffer)
-	cmd.Stderr = stderrBuf
-	err := cmd.Run()
 
 	if err != nil {
-		log.Error("Error running command", "command", executable, "error", err.Error())
-		os.Stderr.Write(stderrBuf.Bytes())
+		return nil, nil, fmt.Errorf("failed to create temporary R script: %w", err)
 	}
-	return stdout.Bytes(), stderrBuf.Bytes(), err
+
+	tempFile.WriteString(script)
+	tempFile.Sync()
+
+	args = append(args, "-f", tempFile.Name())
+	return e.RunCommand(executable, args, cwd, log)
 }
--- a/internal/executor/executor.go
+++ b/internal/executor/executor.go
@@ -4,6 +4,7 @@ package executor
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"os/exec"
 	"strings"
@@ -14,6 +15,7 @@ import (
 
 type Executor interface {
 	RunCommand(executable string, args []string, cwd util.AbsolutePath, log logging.Logger) ([]byte, []byte, error)
+	RunScript(executable string, args []string, script string, cwd util.AbsolutePath, log logging.Logger) ([]byte, []byte, error)
 }
 
 type defaultExecutor struct{}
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by dotNomad** [comment]
Removing myself as a reviewer and unsubscribing until this is ready. Feel free to re-request a review when you'd like me to take a look.

**Comment by jonkeane** [comment]
I also removed marcos and Bill, sorry fro the spam I'll re-add when I'm not fighting with windows CI 🫠

File: .github/workflows/agent.yaml [hunk 1]
```diff
@@ -13,6 +13,28 @@ jobs:
...
+          brew tap r-lib/rig
+          brew install --cask rig
+      - name: Install rig with Choco
+        if: ${{ runner.os == 'Windows' }}
+        run: choco install rig
+
+      - name: Install R with rig
+        run: rig add release
+      - name: Install renv package
+        run: Rscript -e "install.packages('renv')"
```
jonkeane, line 37: We could consider running these functional style tests in a separate action, though IMHO that's a refinement for later.

internal/executor/executor.go
```diff
@@ -40,3 +42,57 @@ func (e *defaultExecutor) RunCommand(executable string, args []string, cwd util.
...
+		return "", nil, fmt.Errorf("failed to flush temporary R script to disk: %w", err)
+	}
+
+	// Return the file path and a cleanup function
+	cleanup := func() {
+		tempFile.Close()
+		os.Remove(tempFile.Name())
+	}
+	return tempFile.Name(), cleanup, nil
+}
```
jonkeane, line 72: I was a little surprised I couldn't find a function that did this in the standard library / the filesystem packages we have. But maybe I just missed it. I'm happy to 🚮 this code if there's a better + more standard thing we can just use instead of not-invented-hereing this.

internal/interpreters/r_interpreter_functional_test.go
```diff
@@ -0,0 +1,213 @@
+package interpreters
```
jonkeane, line 1: Here and in the `r_functional_test.go` — I'm not sure if this is the right pattern for doing functional-style tests like this (e.g. comingled with the go unit tests). But I do think it is good to have them to catch circumstances like this one where our assumptions about how our calls get made hit operating system quirks.

.github/workflows/agent.yaml
dotNomad: Sounds good, would you mind creating a follow-up issue for that when we merge this?

I found it a bit confusing these were part of the unit tests at first blush, but understand it a bit more after digging in. The separation would make it a bit easier to understand right from the start.

internal/interpreters/r_interpreter_functional_test.go
```diff
@@ -0,0 +1,213 @@
...
+		return
+	}
+
+	// Create test project with renv.lock
+	s.createTestRenvLock()
+
+	// Get lockfile path
+	lockfilePath, err := s.defaultInterpreter.getRenvLockfilePathFromRExecutable(s.rExecutable.String())
+	if err != nil {
+		s.T().Logf("Error getting lockfile path: %v", err)
```
dotNomad, line 170: I ran `just test` locally and got a failure:

```
        r_functional_test.go:170: 
                Error Trace:    /Users/jordan/development/publishing-client/internal/publish/r_functional_test.go:170
                Error:          Expected nil, but got: &types.AgentError{Code:"renvPackageVersionMismatch", Err:(*errors.errorString)(0x1400003a2a0), Message:"Package renv: versions in lockfile '1.1.4' and library '1.0.7' are out of sync. Use renv::restore() or renv::snapshot() to synchronize.", Op:"", Data:types.ErrorData{"LibraryVersion":"1.0.7", "Lockfile":"/var/folders/wn/83k9zr2s791brfmfk90c_6vh0000gp/T/r-packages-test-2533733140/With Spaces/renv.lock", "LockfileVersion":"1.1.4", "Package":"renv"}}
                Test:           TestRPublishFunctionalSuite/TestGetRPackagesFunctional
    --- FAIL: TestRPublishFunctionalSuite/TestPublishWithClientFunctional (1.55s)
        r_functional_test.go:82: Using R executable: /usr/local/bin/R
        r_functional_test.go:96: renv is installed but has issues: the renv environment for this project is not in a healthy state. Run renv::status() for more details
        r_functional_test.go:304: 
                Error Trace:    /Users/jordan/development/publishing-client/internal/publish/r_functional_test.go:304
                Error:          Received unexpected error:
                                package renv: versions in lockfile '1.1.4' and library '1.0.7' are out of sync. Use renv::restore() or renv::snapshot() to synchronize
```

I have `R 4.4.2` installed locally with `rig`. Perhaps there is some additional setup I need to get the tests working as expected?

internal/interpreters/r_test.go
```diff
@@ -68,7 +68,7 @@ func (s *RSuite) TestInit() {
 
 	executor := executortest.NewMockExecutor()
 	executor.On("RunCommand", mock.Anything, []string{"--version"}, mock.Anything, mock.Anything).Return([]byte("R version 4.3.0 (2023-04-21)"), nil, nil)
-	executor.On("RunCommand", mock.Anything, []string{"-s", "-e", "renv::paths$lockfile()"}, mock.Anything, mock.Anything).Return([]byte(`[1] "/test/sample-content/shinyapp/renv.lock"`), nil, nil).Once()
+	executor.On("RunScript", mock.Anything, []string{"-s"}, "renv::paths$lockfile()", mock.Anything, mock.Anything).Return([]byte(`[1] "/test/sample-content/shinyapp/renv.lock"`), nil, nil).Once()
```
dotNomad, line 71: If I'm understanding correctly we are changing these from `RunCommand` to `RunScript` because `R... -f` can be used. For my own edification what is the distinction for when to run using `RunScript` and `RunCommand`?

dotNomad (commented):
Few questions and an error when running `just test`. I'll continue the review based on the new info

internal/interpreters/r_test.go
dotNomad: A comment explaining the difference would be fantastic 👍

internal/interpreters/r_interpreter_functional_test.go
jonkeane: Oh, actually, another possibility is to to update the `createTestRenvLock()` to actually use renv to create that (so that it matches). Also this is a flavor of #2646 it's annoying for us, for testing, but it also is annoying to our users!

internal/executor/executor.go
```diff
@@ -40,3 +42,57 @@ func (e *defaultExecutor) RunCommand(executable string, args []string, cwd util.
...
 	return stdout.Bytes(), stderrBuf.Bytes(), err
 }
+
+// createTempRScript creates a temporary file with R code and returns the file path.
+// The caller is responsible for removing the temporary file.
+func createTempRScript(content string) (string, func(), error) {
+	tempFile, err := os.CreateTemp("", "*")
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create temporary R script: %w", err)
+	}
```
tdstein, line 52: I would use defer here to ensure that the file is always deleted instead of handling it within the error blocks. https://go.dev/blog/defer-panic-and-recover

internal/interpreters/r_interpreter_functional_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+// skipIfNoR skips the current test if R executable is not available
+// unless we're running in CI where we should enforce that R tests run
+func (s *RInterpreterFunctionalSuite) skipIfNoR() bool {
+	if os.Getenv("GITHUB_ACTIONS") != "true" && s.rExecutable.String() == "" {
+		s.T().Skip("Skipping test: R executable not found")
+		return true
+	}
+	return false
+}
```
tdstein, line 75: I would approach these as integration tests and run them outside of the unit tests. That way you don't have to conditionally check if R exists on the system.

jonkeane: _nods_ let me see if I grok this before I implement it: if we took the `t.Short()` approach it would be adding something like 

```
   if testing.Short() {
        t.Skip("skipping functional tests in short mode.")
    }
```

And then in the Justfile add in `--short` by default (but in CI, when we want the functional tests, we don't use that). yeah?

internal/executor/executor.go
```diff
@@ -40,3 +42,57 @@ func (e *defaultExecutor) RunCommand(executable string, args []string, cwd util.
...
+		tempFile.Close()
+		os.Remove(tempFile.Name())
+		return "", nil, fmt.Errorf("failed to flush temporary R script to disk: %w", err)
+	}
+
+	// Return the file path and a cleanup function
+	cleanup := func() {
+		tempFile.Close()
+		os.Remove(tempFile.Name())
+	}
```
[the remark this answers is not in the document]
jonkeane: Ah, yeah that makes sense. And just to be clear (because I questioned this while I was writing this): I _can't_ just `defer` from inside of `createTempRScript()` because that would cleanup the file when the `createTempRScript()` finishes (and critically before our command can call it) yeah?

Maybe this is actually a sign that this should be all within the `RunScript()` method at that level and `defer` can do it all 🤔

internal/interpreters/r_interpreter_functional_test.go
tdstein: Yep, that's exactly how I was thinking about it.

internal/executor/executor.go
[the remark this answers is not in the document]
tdstein: Right. Moving to `RunScript` makes the most sense. An alternative approach could be to make createTempScript a closure, but that seems like overkill for this situation.

.github/workflows/agent.yaml
jonkeane: > This should be its own GitHub Action on the marketplace :)

There actually is [a set of actions provided by r-lib](https://github.com/r-lib/actions) which is great and does lots of good things *but* it doesn't use rig (and thus doesn't have hit the `R.bat` / `cmd.exe` quoting issues). 🫠 I'll add a comment about this just so it's clear _why_ we are doing this kind of install.

internal/executor/executor.go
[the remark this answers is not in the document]
jonkeane: Ok, I've done this, would you mind confirming that I have the defer in the right place?

jonkeane:
Ok, I've followed up on all of the comments. Would y'all mind giving this a look again?

dotNomad (approved):
I was able to run `just test` with no problems, and this is look very clean. Marking as approved, but I'd wait a moment for @tdstein to take a look for the Go specific review like answer the question around `defer` being in the right spot.

jonkeane:
> I was able to run `just test` with no problems

Would you also mind trying `just test ./...` which will (try to) run the functional tests too? I'm curious if you get errors — it's not the end of the world if you do, since now they don't run by default — but it would be good to know if you do!

dotNomad:
> Would you also mind trying `just test ./...` which will (try to) run the functional tests too? I'm curious if you get errors — it's not the end of the world if you do, since now they don't run by default — but it would be good to know if you do!

Absolutely. I just ran `just test ./...` at the root with no errors, ran great.

I did notice that when I tried to do the "Run Extension" local development in VS Code from the `/extensions/vscode` folder that I was getting unknown API errors on this branch. I suspect that is because this diverged from `main` a bit and there are schema updates, etc. It may be worth updating off `main` and I can re-test that to make sure there isn't something odd going on with this branch.

internal/interpreters/r_interpreter_functional_test.go
```diff
@@ -0,0 +1,179 @@
...
+	if testing.Short() {
+		t.Skip()
+	}
+
+	suite.Run(t, new(RInterpreterFunctionalSuite))
+}
+
+func (s *RInterpreterFunctionalSuite) SetupTest() {
+	// Set up the test directory
+	parentDir, err := os.MkdirTemp("", "r-interpreter-test-*")
```
tdstein, line 40: I think you can use `afero` to create this temporary directory. I haven't used `afero` in a while, so I could be mistaken.

justfile
```diff
@@ -233,8 +233,9 @@ run *args:
     pathname=`just executable-path`
     ${pathname} {{ args }}
 
-# Execute unit tests.
-test *args=("./..."):
+# Execute unit tests. -short disables our functional tests
+# To run all of the tests, use `just test ./...`
+test *args=("-short ./..."):
```
tdstein, line 238: Maybe for a different pull request, but some tooling to call the functional tests directly could be nice.

.github/workflows/agent.yaml
```diff
@@ -13,7 +13,33 @@ jobs:
...
+        if: ${{ runner.os == 'macOS' }}
+        run: |
+          brew tap r-lib/rig
+          brew install --cask rig
+      - name: Install rig with Choco
+        if: ${{ runner.os == 'Windows' }}
+        run: choco install rig
+
+      - name: Install R with rig
+        run: rig add release
```
tdstein, line 37: This would be a nice GitHub Action for the Marketplace. 😁 

```yaml
- uses: posit-dev/setup-r
  with:
    r-version: release
```

internal/interpreters/r_interpreter_functional_test.go
jonkeane: _nods_ I looked into `afero` when I started and didn't find a way to do this easily (or as easily as ^^^). Mind if we punt on this refinement? I can add a TODO, but I think it's better to get this fix out than it is to be perfect here. Thoughts?

PR #2642, opened 2025-05-09
Mimic major-minor Connect behavior

amol-:
## Intent

Adapt all version numbers that don't have a constraint to a compatible constraint that accepts any patch release.

## Type of Change

- - [x] New Feature 

## Approach

Change both python_requires and r_requires to apply a adaptToCompatibleConstraint function in charge of performing the correct padding of version number and applying the compatible operator.

## User Impact

Mimics the major-minor current behaviour of connect

## Automated Tests

Tested by the `TestGetPythonRequiresPythonVersionAdapt` test.

## Checklist

Should we? This is a change over a previous unreleased feature,
I guess the existing CHANGELOG still applies?


internal/interpreters/python_requires.go
```diff
@@ -162,9 +162,18 @@ func adaptPythonRequires(raw string) (string, error) {
 		return "==" + constraint, nil
 	}
 
-	if strings.Count(constraint, ".") == 1 {
-		// e.g. 3.8, what the user want is 3.8.*
-		constraint = constraint + ".0"
+	return adaptToCompatibleConstraint(constraint), nil
+}
+
+func adaptToCompatibleConstraint(constraint string) string {
```
marcosnav, line 168: It might be good to have this shared method as part of a new file (e.g: `interpreters.go`) within the same package, just so to be a bit more clear that it is not hardly associated with python

marcosnav (approved):
Looks good!, one minor suggestion

internal/interpreters/python_requires.go
amol-: I wanted to keep it close to the "requires" code,
It probably deserves a better packaging story for `r_require` and `python_require` as they are currently mixed in interpreters, but in reality are unrelated to a specific interpreter at all.

For the moment I'd keep it here close to the code that uses it, but I'm happy to revise it in a future PR if there are suggestions on how to refactor the package related to requirements parsing for R and Python

PR #2644, opened 2025-05-13
Release 1.14.0

marcosnav:
Prep for Publisher Release 1.14.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.14.0]
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Detection of required python interpreter version now will request a compatible
   version when there is `.python-version` file declaring a specific version number
   without any version requirement operator (#2628)
-
 - Fixed an issue where scanning for Python dependencies created an `renv.lock` file.
   Now the dependencies file is properly named `requirements.txt`. (#2639)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [1.14.0]
```
marcosnav, line 8: Only changing the heading, we've been keeping up to date the entries as PRs landed

docs/licenses.md
```diff
@@ -15190,7 +15190,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 ```
 
-### sass-embedded-darwin-arm64/dart-sass/src
+### sass-embedded-darwin-x64/dart-sass/src
```
sagerb, line 15193: I guess the platform is flipping back and forth based on the developer's platform, who is running the license generation?

I'm guessing this shouldn't impact the release, but should this be looked at?

sagerb (commented):
The CHANGELOG.md needs to be updated in two places, one included within this PR, and the other at the root of the repo.

I don't think you can get away from having to build up the changelogs. There are two PRs that did not update both of the change logs, but it seems like they should have:
- https://github.com/posit-dev/publisher/pull/2642
- https://github.com/posit-dev/publisher/pull/2635

You'll also want to make sure the issues have been assigned to the milestone.

marcosnav:
@sagerb 

> You'll also want to make sure the issues have been assigned to the milestone.
Thanks! good call, updated the milestone and tickets, bumped over remaining issues to next milestone too.

About CHANGELOG, both files were modified already, and the PRs you mention are related to the same entries, some features have multiple PRs related to them.

Thanks for checking!

docs/licenses.md
marcosnav: yup, it seems so 🤷‍♂️

dotNomad (approved):
Small note, the extra line breaks for the lists in the changelog cause rendering differences in the changelog on the site, and in VS Code. https://marketplace.visualstudio.com/items/Posit.publisher/changelog

Markdown is expecting just the one line break.

---

Other than that, descriptions in the changelogs look good.

PR #2645, opened 2025-05-14
Bug and feature GH issue templates

marcosnav:
It'd be good to have better guidance when reporting bugs or feature requests with new issue templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,24 @@
+---
+name: Report a bug
+about: Found something? Thank you for telling us!
+labels: ["bug"]
+---
+
+## System details
+
+    Publisher Version        :
+    Connect Version          :
+    Your OS                  :
+    Content Type to Deploy   :
+
+## Publisher Output and Logs
+
+<!--
+Publisher output helps us to identify the problem,
+it is located in the "Output" window under the "Posit Publisher Deployment" option.
+See https://github.com/posit-dev/publisher/blob/main/docs/troubleshooting.md#publisher-output
+-->
+
+## Problem Statement
+
+## Steps to Reproduce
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,24 @@
...
+    Connect Version          :
+    Your OS                  :
+    Content Type to Deploy   :
+
+## Publisher Output and Logs
+
+<!--
+Publisher output helps us to identify the problem,
+it is located in the "Output" window under the "Posit Publisher Deployment" option.
+See https://github.com/posit-dev/publisher/blob/main/docs/troubleshooting.md#publisher-output
```
marcosnav, line 19: This is how the new section in troubleshooting will look like

---

Screenshot 2025-05-14 at 12 06 10 p m

docs/troubleshooting.md
```diff
@@ -116,6 +116,19 @@ be displayed:
 Posit Connect does not allow the content type of a deployment to be changed. If you encounter this
 error, your best option is to create and deploy a new deployment.
 
+## Publisher Output
+
+Publisher output may help to track down problems, it is located in the "Output" window under the "Posit Publisher Deployment" option.
+
+The Output Panel in VSCode or Positron can be opened with ⇧+⌘+U in Mac, Ctrl+⇧+U in Windows.
```
dotNomad, line 123: Another way to show it which may be a bit easier is doing

`> Output: Show Output Channels...` in the command palette then selecting "Posit Publisher Deployment"

dotNomad (approved):
Better templates are great. I added a comment about a slightly easier (to me) way to get to the "Posit Publisher Deployment" output channel which we may want to document instead. It would remove a screenshot too I think.

docs/troubleshooting.md
marcosnav: Yes, that's helpful! I'll include this option too 👍

PR #2655, opened 2025-05-23
emit client factory errors so publish flow will halt

### LINKED ISSUE #2654: NewConnectClient error should halt deployment
Most errors in the course of a deployment will halt the publishing process and surface an error in the UI. However, an error from the client factory will not:

https://github.com/posit-dev/publisher/blob/709a1a534bf1b5b32695ca484e4b2d4885b2b2c2/internal/publish/publish.go#L197-L200

If the client creation fails, the UI will hang forever in the "Publishing" phase.

To reproduce:
- add a valid snowflake credential
- modify your snowflake `connections.toml` so that it is no longer valid
- try to deploy

christierney:
## Intent

An error from constructing a new Connect client will now emit an event so the UI can react.

fixes #2654

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Noticed via debugging that this specific error gets logged but is not surfaced to the UI. Added a call to emit the error event.

This was sufficient to get the actual error message to surface in the UI and to halt the publishing flow, so I stopped there. Possibly it would be better to create a specific Agent Error code? I'm not sure what that gets us.

## User Impact

Users who break their snowflake configuration will now get a useful error message instead of a hung process.

## Automated Tests

Added a unit test.

## Directions for Reviewers

Assume a valid snowflake connections.toml:

```toml
[default]
account = "the-account"
user = "user@posit.co"
authenticator = "SNOWFLAKE_JWT"
private_key_file = "/Users/me/.snowflake/rsa_key.p8"
```

Add a new credential for Connect running in snowflake, using this connection.

Edit the connection to break it:

```toml
...
private_key_file = "/Users/me/.snowflake/rsa_key.NOW_INVALID"
```

Attempt to deploy content. Observe that the deployment now halts and reports the error.

Screenshot 2025-05-23 at 2 39 14 PM

Screenshot 2025-05-23 at 2 39 19 PM

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Detect and report auth client errors at deploy time. (#2654)
+
 ## [1.16.0]
 
 ### Added
--- a/extensions/vscode/CHANGELOG.md
+++ b/extensions/vscode/CHANGELOG.md
@@ -6,6 +6,12 @@ file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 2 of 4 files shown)


dotNomad (approved):
Nice catch, and clean fix. 

An Agent Error code could give us the ability in the frontend to capture it and display a different error than one that is bubbled up from the backend, but from the examples you gave I think the errors are pretty darn good.

PR #2662, opened 2025-06-09
Use afero instead of utils

jonkeane:
Cleanup from #2637

I made this change, but apparently didn't push it up. cc @tdstein

### DIFF
--- a/internal/interpreters/r_interpreter_functional_test.go
+++ b/internal/interpreters/r_interpreter_functional_test.go
@@ -20,7 +20,6 @@ type RInterpreterFunctionalSuite struct {
 	suite.Suite
 	testProjectDir     util.AbsolutePath
 	log                logging.Logger
-	fs                 afero.Fs
 	interpreter        RInterpreter
 	rExecutable        util.AbsolutePath
 	defaultInterpreter *defaultRInterpreter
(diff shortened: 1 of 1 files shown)


internal/interpreters/r_interpreter_functional_test.go
```diff
@@ -37,11 +36,12 @@ func TestRInterpreterFunctionalSuite(t *testing.T) {
 
 func (s *RInterpreterFunctionalSuite) SetupTest() {
 	// Set up the test directory
-	parentDir, err := os.MkdirTemp("", "r-interpreter-test-*")
+	// We can't use an in memory filesystem here because we will end up chdiring to it
+	fs := afero.NewOsFs()
```
jonkeane, line 40: @tdstein So this actually was causing errors using the in memory file system since we can't ch into it. This might actually mean these changes are silly anyway?

tdstein: Ah. This might not work since the source code isn't also using afero.

jonkeane: Yeah, I'm going to close this then, we can come back and cleanup the operative code later.

Issue #2669, opened 2025-06-23
API supports adding Connect Cloud credential


rodriin:
### Description

Add the POST API request to add a credential.

Issue #2668, opened 2025-06-23
Store and retrieve Connect Cloud credentials


rodriin:
### Description

The API should support storing Connect Cloud credentials and returning them at the account level.

- Create a v2 credential object.
- Add the following fields:
  - serverType - connect or connect_cloud
  - accountId (Cloud only)
  - refreshToken (Cloud only)
  - accessToken (Cloud only)
- The name (nickname) field will contain the account name.
- When Publisher reads a v1 schema, automatically rewrite it to a v2 schema.

PR #2684, opened 2025-06-24
add support for adding/retrieving Connect Cloud credentials

mslynch:
## Intent

Resolves https://github.com/posit-dev/publisher/issues/2668.
Resolves https://github.com/posit-dev/publisher/issues/2669.

## Type of Change

This code adds new fields to the credential passed to and returned from the API:
- `serverType`
- `accountId`
- `accountName`
- `refreshToken`
- `accessToken`

These fields are also saved in the keyring or credential file. The credential version is bumped to 2.

Empty fields are also now omitted from the credential file.

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

I introduced a new CredentialV2 struct. When we read a credential with a version under 2, we'll automatically populate its ServerType.

To avoid adding too many arguments to `CredentialService.Set()`, I created a new struct `CreateCredentialDetails` to pass instead. This struct is initialized in the API handler.

This PR also includes a factor to extract some common validation logic from the `CredentialsService.Set()` implementations into `CreateCredentialDetails.ToCredential()`.

I think `fileCredentialsService` and `keyringCredentialsService` may be due for even further refactoring. I think both services should reuse the same code for validation and to match on credential version for version upgrades, but I think this PR is big enough as it is right now.

I had to extract `server_type.go` into its own package to be able to use it in `accounts` and `credentials`.

## User Impact

The only user impact should be the format change for credential storage.

## Automated Tests

Unit tests were added to cover the new server type and validation for associated fields.

## Directions for Reviewers

Add a credential from the UI as normal and ensure it can still be used to publish.

## Checklist


dotNomad:
It looks like this is still in draft and not ready for review. I'm going to unsubscribe and remove myself as a reviewer, but feel free to add me again when this ready for a look.

internal/accounts/provider_env_var.go
```diff
@@ -32,7 +33,7 @@ func (p *envVarProvider) Load() ([]Account, error) {
 	if apiKey == "" {
 		return nil, nil
 	}
-	serverType, err := ServerTypeFromURL(serverURL)
+	serverType, err := server_type.ServerTypeFromURL(serverURL)
```
rodriin, line 36: If we are saving the server type in the keyring or credential file, why are we still getting it from the URL? Ideally the server type will be passed in from the UI when saving a credential and then retrieved from the keyring or credential file when reading a credential.

I think you will need my changes in my branch to accomplish the above.

internal/credentials/credentials.go
```diff
@@ -75,6 +97,34 @@ type CredentialTable = map[string]CredentialRecord
...
+		GUID:                cr.GUID,
+		Name:                cr.Name,
+		URL:                 cr.URL,
+		ApiKey:              cr.ApiKey,
+		SnowflakeConnection: "",
+	}
+}
+
+func (cr *CredentialV1) toV2() (*CredentialV2, error) {
+	serverType, err := server_type.ServerTypeFromURL(cr.URL)
```
rodriin, line 111: Same thing here, I think the serverType should be passed in the credential object.

internal/credentials/errors.go
```diff
@@ -101,7 +101,7 @@ func NewIncompleteCredentialError() *IncompleteCredentialError {
 }
 
 func (e *IncompleteCredentialError) Error() string {
-	return "New credentials require non-empty Name, URL and either Api Key or Snowflake connection fields"
+	return "New credentials require non-empty Name, URL and either Api Key, Snowflake, or Connect Cloud connection fields"
```
rodriin, line 104: Does the Connect Cloud credential require the URL field too?

internal/credentials/file.go
```diff
@@ -21,15 +21,46 @@ var fsys = afero.NewOsFs()
...
 }
 
 func (cr *fileCredential) IsValid() bool {
-	return cr.URL != "" && (cr.ApiKey != "" || cr.SnowflakeConnection != "")
+	return cr.URL != "" && (cr.ApiKey != "" || cr.SnowflakeConnection != "" ||
+		(cr.AccountID != "" && cr.AccountName != "" && cr.RefreshToken != "" && cr.AccessToken != ""))
+}
+
+func (cr *fileCredential) toCredential(name string) (*Credential, error) {
+	serverType, err := server_type.ServerTypeFromURL(cr.URL)
```
rodriin, line 48: I think server type should be read from the fileCredential object here.

internal/server_type/server_type.go
```diff
@@ -10,10 +10,11 @@ import (
...
 
 const (
-	ServerTypeConnect     ServerType = "connect"
-	ServerTypeShinyappsIO ServerType = "shinyapps"
-	ServerTypeCloud       ServerType = "cloud"
-	ServerTypeSnowflake   ServerType = "snowflake"
+	ServerTypeConnect      ServerType = "connect"
+	ServerTypeShinyappsIO  ServerType = "shinyapps"
+	ServerTypeCloud        ServerType = "cloud"
+	ServerTypeConnectCloud ServerType = "connect_cloud"
```
rodriin, line 16: The list of server type options should be dwindled down to the 3 actual server types we are supporting or will be supporting so this list can match the UI and just use the server type sent from the UI instead of calculating it from the URL.

internal/services/api/post_credentials.go
```diff
@@ -45,7 +54,16 @@ func PostCredentialFuncHandler(log logging.Logger) http.HandlerFunc {
...
+		cred, err := cs.Set(credentials.CreateCredentialDetails{
+			Name:                body.Name,
+			URL:                 body.URL,
+			ApiKey:              body.ApiKey,
+			SnowflakeConnection: body.SnowflakeConnection,
+			AccountID:           body.AccountID,
+			AccountName:         body.AccountName,
+			RefreshToken:        body.RefreshToken,
+			AccessToken:         body.AccessToken,
+		})
```
rodriin, line 66: This POST request should receive the server type in the credential so we can stop relying on calculating it from the server URL.

internal/credentials/errors.go
mslynch: Yeah. Maybe in the future we could refactor this to make it not required. Right now this keeps things consistent, though.

internal/accounts/provider_env_var.go
mslynch: This code is for environment variable configuration, which I didn't touch. It still needs to determine the server type, and there's no environment variable to set that.

internal/services/api/post_credentials.go
mslynch: This is blocked by your ongoing work, right? I filed https://github.com/posit-dev/publisher/issues/2689 so we can do this later.

dotNomad (approved):
Pulled this down and tested with a new piece of content, and deploying an existing project and both worked perfectly. I also re-created one of my credentials for a Connect instance which worked great with the current changes.

I also turned off keychain credentials and created a new file credential and that worked well.

It looks like this is just the backend implementation so I didn't attempt to create a Posit Cloud credential.

internal/accounts/accountstest.go
```diff
@@ -3,6 +3,7 @@ package accounts
 // Copyright (C) 2023 by Posit Software, PBC.
```
sagerb, line 3: I think it is probably a mistake that this file was named without an underscore between accounts and test. I know this was ahead of your changes, but do you mind changing it? I don't think it gets executed as part of the test suite, if it isn't named correctly.

sagerb (approved):
Changes look good, thanks so much for giving me a chance to review.

I've only got one small request for a file rename to correct a pre-existing condition, but should hopefully make sure your new test code is executed.

PR #2687, opened 2025-06-25
add support for Snowflake "oauth" authenticator

christierney:
This works with the connection automatically configured by Posit Workbench when Snowflake session credentials are enabled.

Uses the alternative auth approach mentioned at
https://docs.snowflake.com/en/developer-guide/snowpark-container-services/working-with-services#ingress-authentication. That is, we issue a login request using the oauth token from the configured connection, then retrieve a session token from the response.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist


internal/api_client/auth/snowflake/connection_test.go
```diff
@@ -193,24 +198,32 @@ func (s *SnowflakeSuite) TestList() {
...
 				Account:        "path-acct",
 				User:           "path-user",
 				PrivateKeyFile: "/tmp/path/rsa_key.p8",
 				PrivateKeyPath: "/tmp/path/rsa_key.p8",
+				Authenticator:  "SNOWFLAKE_JWT",
+			},
+			"workbench": {
+				Account:       "workbench-acct",
+				Token:         "workbench-token",
+				Authenticator: "oauth",
```
zackverham, line 220: nit: is there a reason you're not capitalizing the authenticator parameter here like you do elsewhere?

christierney:
## validation

I downloaded the artifacts from the Github Action and used [wormhole](https://magic-wormhole.readthedocs.io/) to send it to a running VS Code workbench session. I used the New Credential flow to add https://bf2oiaib-duloftf-posit-software-pbc-dev.snowflakecomputing.app and the extension automatically detected and used the workbench-provided connection. Successfully deployed a shiny app: https://bf2oiaib-duloftf-posit-software-pbc-dev.snowflakecomputing.app/connect/#/apps/85e7a832-b6c1-4fc2-8573-a50488018578/access

internal/api_client/auth/snowflake/connection_test.go
christierney: Yes! It's because the file generated by workbench does it like this. Meanwhile, our own docs (https://musical-doodle-y61w93q.pages.github.io/) use all-caps `SNOWFLAKE_JWT`. So I opted for ignoring case, and this test reflects both styles.

christierney: (good question)

christierney: This also caused me to go and check the snowflake docs and they use both styles on the same page!

https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#using-key-pair-authentication-and-key-pair-rotation
https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-with-oauth

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -779,35 +779,27 @@ export async function newDeployment(
...
-
-        if (!pick || !isQuickPickItemWithIndex(pick)) {
-          return;
-        }
+      const pick = await input.showQuickPick({
+        title: state.title,
+        step: 0,
+        totalSteps: 0,
+        placeholder:
+          "Select the Snowflake connection configuration you want to use to authenticate.",
```
tdstein, line 787: nit; 
```suggestion
          "Select the Snowflake connection to use for authentication.",
```

internal/api_client/auth/snowflake/token_provider.go
```diff
@@ -0,0 +1,235 @@
...
+	user string,
+	privateKeyFile string,
+	role string,
+) (*JWTTokenProvider, error) {
+	pemData, err := os.ReadFile(privateKeyFile)
+	if err != nil {
+		return nil, fmt.Errorf("error loading private key file: %w", err)
+	}
+
+	block, _ := pem.Decode(pemData)
```
tdstein, line 55: Is this an ignored error?

internal/api_client/auth/snowflake/token_provider.go
```diff
@@ -0,0 +1,235 @@
...
+	if block == nil || block.Type != "PRIVATE KEY" {
+		return nil, fmt.Errorf("decoding PEM data failed")
```
tdstein, line 57: nit; 
```suggestion
		return nil, fmt.Errorf("failed to decode PEM block containing private key")
```

internal/api_client/auth/snowflake/token_provider.go
```diff
@@ -0,0 +1,235 @@
...
+		"exp": expiration.Unix(),
+	}
+
+	// Create the token
+	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+
+	// Sign the token using the RSA private key
+	signedToken, err := token.SignedString(p.privateKey)
+	if err != nil {
+		return "", fmt.Errorf("failed to sign token: %w", err)
```
tdstein, line 121: nit;

```suggestion
		return "", fmt.Errorf("failed to sign JSON Web Token (JWT): %w", err)
```

internal/api_client/auth/snowflake/token_provider.go
```diff
@@ -0,0 +1,235 @@
...
+
+// GetToken logs in to snowflake using the OAuth token and retrieves a session
+// token.
+func (p *OAuthTokenProvider) GetToken(host string) (string, error) {
+	reqData := map[string]any{
+		"data": map[string]string{
+			"ACCOUNT_NAME": p.account,
+			// workbench is not providing a user name,
+			// but it seems to not be actually required anyway
+			// "LOGIN_NAME": user,
```
tdstein, line 186: Should this be removed?

tdstein (approved):
Overall, this looks good to me. If you have time for another pass, I added some nits about error text and test structure.

internal/api_client/auth/snowflake/token_provider.go
christierney: Some background: we are basically cloning the behavior of https://github.com/snowflakedb/snowflake-connector-python to accomplish this login. That connector code supplies a `LOGIN_NAME` but in our case we don't have that; I found that it didn't seem to matter, but I wanted a note to future maintainers explaining the discrepancy.

christierney: (I'll update the comment to reflect this better)

PR #2691, opened 2025-07-01
support snowflake config.toml and environment variable overrides

christierney:
[note: this is layered on top of #2687 which should land first]

## Intent

Support more snowflake connection configs:

- `config.toml` (if this and `connections.toml` both exist, `connections.toml` wins)
- per-connection field env var overrides

Fixes posit-dev/connect#31773

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Followed the behavior in https://github.com/posit-dev/snowflakeauth/, except for the "default connection" support, which doesn't make a ton of sense in the publisher paradigm (we present the user with a list of connection options to choose from, rather than the user entering a connection name or falling back on a default connection name.)

Also ignoring unnamed connection env vars (e.g. `SNOWFLAKE_PRIVATE_KEY_PATH`) because which connection would it apply to? All of them? I think this would be a confusing behavior for the publisher flow, especially as we are not handling default connection options either.

- Look for `config.toml` in the same set of directories as `connections.toml`, but only if we have not found a `connections.toml`.
- Handle parsing the different file formats.
- For each connection, check the environment for named overrides and apply if found.

## User Impact

Snowflake users will have better automatic support for more varieties of snowflake configuration.

## Automated Tests

Added new tests to `connection_test.go`.

## Directions for Reviewers

I tested by launching publisher and adding a new credential, with https://bf2oiaib-duloftf-posit-software-pbc-dev.snowflakecomputing.app as the server URL. Publisher loads the connections and tests them and presents a list of valid connections, so I looked for the expected list of connection names.

### Config.toml support 

Tested with a `config.toml` in my snowflake config home. Verified I saw the right connection names.

Added a `connections.toml` alongside with different connections. Verified that I saw the connections from that file instead (connections takes precedence over config).

### Env var support

Given a TOML file with a working `default` connection and a non-working `workbench` connection (`workbench is missing a valid token):

```toml
[default]
account = "duloftf-posit-software-pbc-dev"
user = "chris@posit.co"
authenticator = "SNOWFLAKE_JWT"
private_key_file = "/Users/chris/.snowflake/rsa_key.p8"

[workbench]
account = "duloftf-posit-software-pbc-dev"
authenticator = "oauth"
```

I verified that Publisher offered me the `default` connection but not the `workbench` connection.

Closed Code, set env vars, relaunched code:

```shell
❯ export SNOWFLAKE_CONNECTIONS_WORKBENCH_TOKEN="[token from workbench]"
❯ export SNOWFLAKE_CONNECTIONS_DEFAULT_ACCOUNT=foo
❯ code
```

I verified that Publisher now offered me the `workbench` connection (it has a valid token now) but no longer offered me the `default` connection (it has an invalid account name now.) This proves that Publisher is using the env var overrides.

## Checklist

We are avoiding public documentation of snowflake features for the time being.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the naming of "Shiny for Python" and "Shiny for R" content types.
   (#2664)
 
+### Added
+
+- Improved support for Posit Connect deployments hosted in Snowpark Container
+  Services. (#2687, #2691)
+
 ## [1.16.1]
 
 ### Fixed
--- a/extensions/vscode/CHANGELOG.md
+++ b/extensions/vscode/CHANGELOG.md
@@ -6,6 +6,13 @@ file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Improved support for Posit Connect deployments hosted in Snowpark Container
+  Services. (#2687, #2691)
+
 ## [1.16.1]
 
 ### Fixed


marcosnav (approved):
Thank you for the comprehensive tests!

PR #2702, opened 2025-07-10
Update docs for running e2e tests

sagerb:
No user impact, this is purely an update to the development e2e flow and documentation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -163,6 +163,8 @@ deactivate
 
 #### Running E2E Tests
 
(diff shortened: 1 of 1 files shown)


sagerb:
> @sagerb These look great, I would just ask you to add the `just dev` command in line 177 where it mentions the `just e2e` command to build and run the cypress tests. Additionally I would document the use of the `just stop` command in-between repeated runs of the cypress tests to do an accurate reset of the docker images.

I have made the changes you requested, please let me know if there was anything else. Thanks!

PR #2703, opened 2025-07-11
decouple connect-specific publishing code from generic publishing code

mslynch:
## Intent

Refactor to decouple publishing and Connect code.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

I extracted the connect-specific code from `publish` into its own module. Later I'll add a similar module with a `ServerPublisher` to publish to Connect Cloud.

I wouldn't be surprised to encounter more code in need of refactoring, but this should at least be a good starting point to write the integration.

## User Impact

None (hopefully)

## Automated Tests

No new tests needed.

## Directions for Reviewers

No special steps - automated test should still pass.

## Checklist

- [ ] I have updated [CHANGELOG.md](../CHANGELOG.md) to cover notable changes. - **n/a**


marcosnav:
@mslynch I'm in the process of reviewing the PR, so far the approach you are taking looks great, I think it is a great idea to decouple the publishing process and have multiple publishing providers, if you will. 

I haven't finished taking a look but there are a couple things I already noticed and want to bring up:
1. It'd be great to include unit tests for the code under the new `/internal/publish/*` files. I understand some of these are being mostly re-organized and that we didn't have tests prior to this changes, so it is not totally fair to come asking for those now 😅 ... but with the increase of added functionality, I think skipping tests should not be an option anymore.
2. Not suggesting to be addressed in this PR. We should start to figure out a way to have integration tests too. Maybe there have been conversations within the hosted team about this already, for now thinking here out loud, new workflows are being introduced for which we are needing checks as insurance practice for future changes. Happy to join your team on conversations about this to have better testing practices in place.

PR #2710, opened 2025-07-21
refactor schema loading to support multiple versions and backwards compatibility

mslynch:
## Intent

Extracted this from https://github.com/posit-dev/publisher/pull/2693.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

This PR changes schema loading to enable backwards compatibility with old schemas. When loading a TOML file, instead of loading it directly into a struct, we load it into a map, extract $schema, and validate against that schema. Then depending on the schema version, we apply changes to make the file data compatible with the current schema version.

Once we introduce the v4 schema, the functions in `conversion.go` will get filled out to migrate from v3.

## User Impact

none

## Directions for Reviewers

Exercise schema validation by deploying an existing piece of content.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improved support for Posit Connect deployments hosted in Snowpark Container
   Services. (#2687, #2691)
+- Added internal API endpoint to list Connect Cloud accounts. (#2695)
 
 ### Changed
 
--- a/extensions/vscode/src/api/resources/Credentials.ts
+++ b/extensions/vscode/src/api/resources/Credentials.ts
@@ -27,12 +27,14 @@ export class Credentials {
     url: string,
     apiKey: string,
     snowflakeConnection: string,
+    serverType: string,
   ) {
     return this.client.post<Credential>(`credentials`, {
       name,
       url,
       apiKey,
       snowflakeConnection,
+      serverType,
     });
   }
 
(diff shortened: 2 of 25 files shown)


mslynch:
> I may need a bit of a walkthrough of this one to understand all of the changes. There are a lot of `mapstructure` changes here, but it isn't clear to me why.
> 
> Could the same approach not be taken on structs? This may be a bit of my lack of Go knowledge showing.

I should have explained in the PR description... Previously, the approach was to unmarshal the toml file directly into a struct using `go-toml` and `toml` tags. The new approach goes like:
1. unmashal the toml file into a map
2. tweak the map data to be compatible with the latest schema
3. unmarshal the map into the struct

The data tweaking in step 2 could have been done with structs, but it's easier to understand with maps - a struct-based approach would involve creating separate versions of structs for each schema version, which could get confusing. In this approach, everything you need to understand the upgrade is in `conversion.go`. Also (not that it matters too much), it lets us remove a redundant file read since we already had to read the file to validate the JSON schema.

The new `mapstructure` tags should all be basically equivalent to the TOML tags.

marcosnav:
In order to understand better the intent for this PR (the open TODOs on the upgrading methods), I had to take a look to the subsequent PR and left a comment there, mainly requesting more info and details to better understand the reasoning for this changes https://github.com/posit-dev/publisher/pull/2713/files#r2229336570

PR #2713, opened 2025-07-22
add schema v4 with server_type field

mslynch:
## Intent

Adds new v4 schema files, copy-pasted from v3 with the following changes:
- publishing schema:
  - adds required `.server_type` field
  - draft:
    - changes `.access` field to `.connect.access_control`
- publishing record schema:
  - makes `.server_type` field required

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [x] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Builds on the changes from https://github.com/posit-dev/publisher/pull/2710 to introduce a new schema version with backwards compatibility with v3.

The next PR will include schema changes to the v4 files to support Connect Cloud. `server_type` will be used by the schema to tell which configuration values are valid for the target.

## User Impact

New publishes and published records will be created in the V4 format. V3 files are still supported.

## Automated Tests

Added tests for conversion from v3 schema to v4.

## Directions for Reviewers

- Publish something new - it should create files in the v4 format.
- Publish something old with v3 files - it should work, but a new deployment record should be created in the v4 format.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bumped the credential version to 2 to support Posit Connect Cloud credentials. (#2684)
-- Changed the publishing and publishing record schema versions to v4 to add Posit Connect Cloud fields. (#2693)
 
 ## [1.16.1]
 


internal/schema/conversion.go
```diff
@@ -2,13 +2,44 @@ package schema
...
 // UpgradePublishingSchema upgrades a publishing schema instance to the current version.
 func UpgradePublishingSchema(data map[string]interface{}) {
-	// TODO: add logic here to upgrade the schema when needed.
+	currentSchema, _ := data["$schema"].(string)
+	if currentSchema != ConfigSchemaURL {
+		// Update the $schema to the latest version
+		data["$schema"] = ConfigSchemaURL
+
+		// Assume server_type is connect. It might actually be snowflake, but we don't have enough information to
+		// determine that, and it shouldn't matter regardless.
```
marcosnav, line 17: I'm not fully on-board with the changes to the schema handling, this includes the previous PR #2710 that refactors some areas to use mapstructures.

Backwards compatibility feels like a good idea, but at the same time I think this work maybe is introducing more complexity than needed, specially when a refactor is being introduced and it seems by these comments, that we cannot guarantee data integrity between schema versions.

Some questions:
**Is there a design doc that explains in more detail the need to have `server_type` in the schema?**
Especially when `ServerTypeFromURL` already serves the purpose to identify the type of server.

**Is backwards compatibility worth it? Is this something worth to maintain? What is the worst case if we don't have backwards compatibility?**
We already use `omitempty` when parsing TOML files to configuration or deployment structs as needed. Normally an older deployment record still works on newer versions of Publisher when features are additive, if the user wants to make use of a new feature, creating a new deployment can be done.

I'm not trying to impose that this work shouldn't land, I can totally be mistaken and I'm mostly writing down thoughts as they come by while reviewing these PRs. Providing more context, design docs and answering above questions might help to understand better.

mslynch: > Is there a design doc that explains in more detail the need to have server_type in the schema?

There isn't a design doc. I added `server_type` to the schema because there are certain configuration fields that are valid for Connect but not Connect Cloud and vice versa. [These changes](https://github.com/posit-dev/publisher/blob/d44523f9608fa32f79dee925e8da9a8d2b4c8983/internal/schema/schemas/posit-publishing-schema-v4.json#L497-L520) (from https://github.com/posit-dev/publisher/pull/2717) in the v4 publishing schema are the key here - the connect and connect_cloud configs are defined separately under `$defs` and will error if an invalid key for that `server_type` is used.

> Is backwards compatibility worth it? Is this something worth to maintain? What is the worst case if we don't have backwards compatibility?

In my opinion, yes; I'd rather impose the cost on our code rather than our users. The worst case for our users is that they see an error the next time they attempt to publish, aren't able or willing to understand it, and give up on the extension.

___

Looking back, I think I may have gone overboard with the mapstructure loading changes...

With omitempty, we should be able to load the file just fine. The backfill in `conversion.go` shouldn't be necessary, either:
- Setting `account_name` to nil served to populate a new required field for the JSON schema, but that can be worked around by setting a default value of null in the schema instead.
- setting `server_type` to `connect` didn't actually effect any change, since no code references the value after it's loaded and the JSON schema already specifies a default of `connect`. That default is a little sketchy from a data integrity perspective, but I'm ok with it given that `connect` and `snowflake` are equivalent in terms of configuration.

___

I do still think backwards compatibility is worth it. If we remove the mapstructure loading changes, the surface of the change becomes much smaller. We would still have to retain the files for both schema versions, but I don't think that's much of a maintenance or complexity burden.

sagerb:
>Is backwards compatibility worth it?

It may have been lost, but the decisions we made around versioning the configuration and deployment files was that we would be "compatible" in that we could read old versions, but we would only write out the latest versions. So we would convert upward, but we did not maintain compatibility in the way that I think of the definition, in that if a file is version 1, and we update it, we write it out in version 1. In the direction we have implemented so far, if a file is version 1, it is converted to version 2, and then used internally as a version 2 structure, and written out as a version 2. The burden that this placed on us was that we needed to be able to read old versions and convert them into new versions, but once those configuration files passed a certain point, they were always the latest version internally and written out as the newest version. This satisfied the customer needs of not losing deployments because of an upgrade, while also satisfying our desire to keep the code as simple as possible.

@mslynch I'm sorry I haven't had a chance to review this particular PR, but I hope these thoughts give you some background and at least my take on what the historical decisions and direction have been so far. Please reach out if you want to discuss further.

PR #2726, opened 2025-07-24
check for existing credential by combination of URL and account name

mslynch:
## Intent

Currently we have checks in place to prevent users from creating new credentials for the same server URL. However, we want users to be able to create multiple credentials for Connect Cloud (but not with the same account). That means we have to compare both the credential URL and account instead of just the URL.

We'll also have to make changes in the extension code, which performs a similar validation.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Updated the `ConflictCheck` method to check AccountID in addition to URL.

## User Impact

none.

## Automated Tests

Added assertions for:
- When a user creates a duplicate cloud credential
- When a user creates a non-duplicate cloud credential with the same URL

## Directions for Reviewers

Run the following curls:
```
curl -X POST localhost:9001/api/credentials -d '{"name": "foo", "serverType": "connect_cloud", "url": "https://api.connect.posit.cloud", "accountId": "123", "accountName": "my account", "refreshToken": "blah", "accessToken": "blah"}'

^ should succeed

curl -X POST localhost:9001/api/credentials -d '{"name": "foo", "serverType": "connect_cloud", "url": "https://api.connect.posit.cloud", "accountId": "123", "accountName": "my account", "refreshToken": "blah", "accessToken": "blah"}'

^ should fail

curl -X POST localhost:9001/api/credentials -d '{"name": "foo 2", "serverType": "connect_cloud", "url": "https://api.connect.posit.cloud", "accountId": "456", "accountName": "my account 2", "refreshToken": "blah", "accessToken": "blah"}'

^ should succeed

```

## Checklist

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -79,8 +79,11 @@ type CredentialV0 struct {
 }
 
 func (c *Credential) ConflictCheck(compareWith Credential) error {
-	if compareWith.URL == c.URL {
-		return NewURLCollisionError(c.Name, c.URL)
+	if compareWith.URL == c.URL && compareWith.AccountID == c.AccountID {
+		// Connect/Snowflake credentials have unique URLs and empty AccountID.
+		// Connect Cloud credentials may have duplicate URLs, but their account IDs must be unique. AccountName is also
+		// unique, but may become unstable in the future if we allow users to change it.
+		return NewIdentityCollisionError(c.Name, c.URL, c.AccountName)
 	}
 	if compareWith.Name == c.Name {
 		return NewNameCollisionError(c.Name, c.URL)
--- a/internal/credentials/errors.go
+++ b/internal/credentials/errors.go
@@ -54,18 +54,23 @@ func (e *NotFoundError) Error() string {
 	return fmt.Sprintf("credential not found: %s", e.GUID)
 }
 
-// URL is used by another credential
-type URLCollisionError struct {
-	Name string
-	URL  string
+// Credential has already been defined
+type CredentialIdentityCollision struct {
+	Name        string
+	URL         string
+	AccountName string
 }
 
-func NewURLCollisionError(name, url string) *URLCollisionError {
-	return &URLCollisionError{name, url}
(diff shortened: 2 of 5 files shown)


internal/credentials/errors.go
```diff
@@ -54,18 +54,23 @@ func (e *NotFoundError) Error() string {
...
 }
 
-func (e *URLCollisionError) Error() string {
-	return fmt.Sprintf("URL value conflicts with existing credential (%s) URL: %s", e.Name, e.URL)
+func (e *CredentialIdentityCollision) Error() string {
+	msg := fmt.Sprintf("URL value conflicts with existing credential (%s) URL: %s", e.Name, e.URL)
+	if e.AccountName != "" {
+		msg += fmt.Sprintf(", account name: %s", e.AccountName)
+	}
+	return msg
```
rodriin, line 73: I wonder if we should be printing the account name or display name? We will be using the display name in the extension. Just want to make sure the user can tie the message back to what they see in the extension. Also, should we print the actual `url` value?

internal/credentials/credentials.go
```diff
@@ -79,8 +79,11 @@ type CredentialV0 struct {
 }
 
 func (c *Credential) ConflictCheck(compareWith Credential) error {
-	if compareWith.URL == c.URL {
-		return NewURLCollisionError(c.Name, c.URL)
+	if compareWith.URL == c.URL && compareWith.AccountID == c.AccountID {
+		// Connect/Snowflake credentials have unique URLs and empty AccountID.
+		// Connect Cloud credentials may have duplicate URLs, but their account IDs must be unique. AccountName is also
+		// unique, but may become unstable in the future if we allow users to change it.
+		return NewIdentityCollisionError(c.Name, c.URL, c.AccountName)
```
rodriin, line 86: I like the simplicity of this!

rodriin (approved):
Looking real good!

PR #2719, opened 2025-07-23
switch to a single Connect-Cloud-Environment header to determine cloud API URLs

mslynch:
## Intent

Replaces the existing cloud URL headers with a new header, `Connect-Cloud-Environment`. If the header is missing, the agent will default production.

It will be easier for the extension to send a single header for the environment rather than two to determine the URLs we need.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

Added a new `CloudEnvironment` type to represent the environment. The API client constructors now accept `CloudEnvironment` instead of `baseURL`.

## User Impact

none

## Automated Tests

Updated API tests to use the new header.

## Directions for Reviewers

Exercise both APIs using these curls:
```
curl -X POST localhost:9001/api/connect-cloud/device-auth -H 'Connect-Cloud-Environment: staging'

curl -X POST localhost:9001/api/connect-cloud/oauth/token -H 'Connect-Cloud-Environment: staging' -d '{"deviceCode": "xxx"}'
```

## Checklist

- [x] I have updated [CHANGELOG.md](../CHANGELOG.md) to cover notable changes.

PR #2728, opened 2025-07-24
respect Connect-Cloud-Environment header when creating cloud credential

mslynch:
## Intent

Per https://github.com/posit-dev/publisher/pull/2719, the extension isn't required to have prior knowledge of the Connect Cloud URLs. Credentials need URLs, so the `POST /credentials` endpoint should respect the header that determines them for Cloud.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Used the existing header and URL determination from the Connect Cloud client for the server URL.

## User Impact

none.

## Automated Tests

Added a test case for `POST /credentials` that passes in a cloud credential with the environment header and asserts the URL on the response is correct.

## Directions for Reviewers

Verify the following curl command works:

```
curl -X POST localhost:9001/api/credentials -d '{"name": "foo 2", "serverType": "connect_cloud", "accountId": "456", "accountName": "my account 2", "refreshToken": "blah", "accessToken": "blah"}' -H 'Connect-Cloud-Environment: staging'
```

## Checklist


internal/services/api/post_credentials.go
```diff
@@ -58,6 +59,11 @@ func PostCredentialFuncHandler(log logging.Logger) http.HandlerFunc {
 			return
 		}
 
+		if body.URL == "" && body.ServerType == server_type.ServerTypeConnectCloud {
+			environment := types.CloudEnvironment(req.Header.Get(connectCloudEnvironmentHeader))
+			body.URL = connect_cloud.GetBaseURL(environment)
+		}
```
rodriin, line 65: I would suggest to make the `if` statement just check for:
```
body.ServerType == server_type.ServerTypeConnectCloud
```
No matter what, we should be taking the url from `connect_cloud.GetBaseURL(environment)`. It should not matter whether the extension sent an empty string for the url or an actual value. The url should be overwritten when the server type is Connect Cloud so it is consistent with the rest of the code.

rodriin (approved):
Looking good, had one suggestion.

PR #2749, opened 2025-07-30
Send product type from extension

mslynch:
## Intent

When we create the deployment, the agent needs to know what product type we're using so we can populate that field in the config file.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Passed product type as `connect_cloud` when the server type is `connect_cloud`; uses `connect` otherwise.

## User Impact

none

## Automated Tests

multi-stepper code is only covered by the e2e tests.

## Directions for Reviewers

I don't think everything for creating a Cloud deployment is hooked together yet - but when it is, this should set `product_type` correctly in the config file when creating a deployment.

Otherwise there should be no visible changes.

## Checklist

### DIFF
--- a/extensions/vscode/src/api/types/configurations.ts
+++ b/extensions/vscode/src/api/types/configurations.ts
@@ -4,6 +4,7 @@ import { AgentError } from "./error";
 import { ConnectConfig } from "./connect";
 import { SchemaURL } from "./schema";
 import { InterpreterDefaults } from "./interpreters";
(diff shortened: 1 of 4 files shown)


mslynch:
this is necessary for the e2e tests to pass; merging

PR #2764, opened 2025-08-01
Re-introducing the Connect Cloud feature as an environment variable

rodriin:
## Intent

Re-introducing the Connect Cloud feature flag as an environment variable to make the Connect Cloud credential feature available in the Publisher extension. The intent is to make this feature available to the QA team and others for testing so we can gather feedback before we release it as a default feature.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The `CONNECT_CLOUD_ENV` variable will be leveraged to turn on and off the Connect Cloud support extension changes, rather than having it hardcoded in the `extension.ts` file. The only valid values for the `CONNECT_CLOUD_ENV` variable are:
- `development`
- `staging`
- `production`

I have set the default value of the `CONNECT_CLOUD_ENV` variable to `production` and made the necessary code changes to only expose the Connect Cloud support for either the `development` or `staging` values. Therefore, keeping the Publisher extension releasable w/o having any risk of exposing incomplete functionality, while making it possible to add E2E tests using the `CONNECT_CLOUD_ENV` variable in CI.

## User Impact

None because the current functionality remains the same.

## Automated Tests

Existing E2E tests succeeded locally for this branch.

## Directions for Reviewers

There are no manual tests that can be verified for this change.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,20 +111,16 @@ an AMD64 architecture. If you are running on an ARM64 architecture (such as Appl
 
 ##### Connect Server License
 
-To run the end-to-end tests, you will need a valid Posit Connect server license set within your environment within the `CONNECT_LICENSE` environment variable. We higly recommend setting this variable in your shell profile (e.g., `.bashrc`, `.zshrc`, etc.) so that it is available in all terminal sessions.
+To run the end-to-end tests, you will need a valid Posit Connect server license.
+Set the `CONNECT_LICENSE` environment variable.
 
 ```bash
 export CONNECT_LICENSE="your-connect-license-key"
 ```
 
-To get an internal license, you can use the new [License Management app](https://connect.posit.it/support/license_management/)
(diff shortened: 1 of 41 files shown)


sagerb (commented):
Environment variables within the extension, as set when the process launched (especially when that process might be workbench), have caused some headaches in the past... That's why I'd suggest avoiding them.

Did you consider simply adding a configuration within the extension config definition of the Connect Cloud URL and info, filled by default to be production, but also able to be changed within tests, so that it goes to staging? This would allow some flexibility in the field as well, just in case the cloud addresses ever change (permanently or temporarily due to an outage).

rodriin:
@sagerb I did add a configuration within the extension and the down side there is that it would be visible and available to everyone using the extension today. We felt this was the safer route to keep Publisher releasable w/o having to compromise testing.

Issue #2768, opened 2025-08-03
feat: add rsconnect-style token authentication


jonkeane:
## Description
Onboarding should be as simple as possible. Though we should pursue a more modern authentication mechanism in Connect (and the publisher), just like Connect Cloud has, we should also make it easy for users of current versions for Connect to authenticate too.

PR #2769, opened 2025-08-03
Implement the old token auth method

jonkeane:
This implements the same token authentication method used by rsconnect. We want to implement something more modern, but this makes it so that this setup is able to be as smooth as it is with rsconnect.

resolves: #2768

## Intent

Implement the token-based auth that matches rsconnect

## Type of Change
- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Lots of Claude, used rsconnect as a base, implemented the token and signing as is from there.

## User Impact

New, smoother default authentication

## Automated Tests

I added some tests for this code, but it probably needs more.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added endpoints for performing OAuth Device Authorization Grant with Posit Cloud login. (#2692)
 - Added support to publishing schema for Connect Cloud. (#2729, #2747)
+- Added support for one-click token authentication with Connect. (#2769)
 
 ### Changed
 
--- a/internal/services/api/post_connect_token_test.go
+++ b/internal/services/api/post_connect_token_test.go
@@ -8,7 +8,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
-	
+
 	"github.com/posit-dev/publisher/internal/logging/loggingtest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
@@ -17,31 +17,31 @@ import (
 func TestPostConnectTokenHandler(t *testing.T) {
 	// This test is minimal because we can't easily mock the Connect server interaction
 	// We'll just verify that the handler correctly parses the request and returns a response
-	
+
 	logger := loggingtest.NewMockLogger()
 	// Set expectation for the Enabled method which is called by the HTTP client
 	logger.On("Enabled", mock.Anything, mock.Anything).Return(false)
 	// Mock the Error method which is called when there's an error during the HTTP request
 	logger.On("Error", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return()
 	handler := PostConnectTokenHandlerFunc(logger)
-	
+
 	// Create request body
 	requestBody, err := json.Marshal(PostConnectTokenRequest{
 		ServerURL: "https://example.com",
 	})
 	assert.NoError(t, err)
-	
+
 	// Create HTTP request
 	req := httptest.NewRequest("POST", "/api/connect/token", bytes.NewReader(requestBody))
 	req.Header.Set("Content-Type", "application/json")
-	
+
 	// Create response recorder
 	recorder := httptest.NewRecorder()
-	
+
 	// Call handler - note this will fail since we're not mocking the Connect client
 	// But we can test the error handling
 	handler(recorder, req)
-	
+
 	// Expect an internal server error since we can't connect to the example.com server
 	assert.Equal(t, http.StatusInternalServerError, recorder.Code)
 }
@@ -49,7 +49,7 @@ func TestPostConnectTokenHandler(t *testing.T) {
 func TestPostConnectTokenUserHandler(t *testing.T) {
 	// This test is minimal because we can't easily mock the Connect server interaction
 	// We'll just verify that the handler correctly parses the request and returns a response
-	
+
 	logger := loggingtest.NewMockLogger()
 	// Set expectation for the Enabled method which is called by the HTTP client
 	logger.On("Enabled", mock.Anything, mock.Anything).Return(false)
@@ -58,7 +58,7 @@ func TestPostConnectTokenUserHandler(t *testing.T) {
 	// Mock the Debug method which is called during token verification
 	logger.On("Debug", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return()
 	handler := PostConnectTokenUserHandlerFunc(logger)
-	
+
 	// Create request body
 	requestBody, err := json.Marshal(struct {
 		ServerURL  string `json:"serverUrl"`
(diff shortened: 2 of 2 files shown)


extensions/vscode/src/multiStepInputs/newCredential.test.ts
```diff
@@ -0,0 +1,191 @@
+// Copyright (C) 2025 by Posit Software, PBC.
```
jonkeane, line 1: Like the other test file, I'm not totally sure if these tests are necessary. They do encapsulate some of the behavior, but there aren't tests for the extant code already.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -455,28 +620,79 @@ export async function newCredential(
...
+    }
+  } else if (isSnowflake(serverType)) {
+    // For Snowflake, require snowflakeConnection
+    if (
+      state.data.snowflakeConnection === undefined ||
+      isQuickPickItem(state.data.snowflakeConnection)
+    ) {
+      return;
+    }
+  }
```
jonkeane, line 661: Is this checking helpful here?

jonkeane:
One note: the screen build into connect has `rsconnect` + `RStudio IDE` baked into it's message. We will need to change that (though it will only be changed for new installs of Connect.

> You're almost done configuring R to publish content to Posit Connect. Click the Connect button below to link your Posit Connect account (admin) to the rsconnect R package.
> 
> If you didn't reach this screen by initiating an account link from the RStudio IDE or in R via rsconnect::connectUser, click Cancel.

jonkeane:
@dotNomad after our chat last week, I was curious how far I could get "just" copying the rsconnect implementation of token auth. Turns out, all the way — at least from my poking at it and deploying locally + dogfood.

One thing that we will need to figure out is that this adds considerably to the data stored with the credential (especially the private key), and keychains on macos are relatively limited (and we will exceed this with more than one private key):

https://github.com/posit-dev/publisher/blob/ff6a884b63a592856aef2b9923a4595eb610ab9b/vendor/github.com/zalando/go-keyring/keyring.go#L14

extensions/vscode/src/multiStepInputs/newCredential.test.ts
marcosnav: Yup, these tests are lacking coverage but I think it is nice to have a starting point, since here we have a few mocks and helpers to extend the test suite.

internal/clients/http_client/http_client.go
```diff
@@ -64,6 +64,19 @@ func NewBasicHTTPClient(baseURL string, timeout time.Duration) *defaultHTTPClien
 	}
 }
 
+func NewAuthenticatedHTTPClient(baseURL string, timeout time.Duration, authMethod auth.AuthMethod) *defaultHTTPClient {
```
marcosnav, line 67: This one does not seem to be used

marcosnav (commented):
Overall it looks pretty good, a few comments and questions. There are areas here for which @rodriin might have more specific thoughts due to WIP on multistepper.

extensions/vscode/src/multiStepInputs/newCredential.ts
jonkeane: _nods_ in the service of moving this PR forward, I'm going to defer that, especially because, as you say, this is getting reworked soon already.

internal/clients/connect/client_connect.go
```diff
@@ -548,3 +548,29 @@ func (c *ConnectClient) ValidateDeploymentTarget(contentID types.ContentID, cfg
...
+	// Convert from UserDTO to User
+	userObj := connectUser.toUser()
+	*user = *userObj
+	return nil
+}
+
+// GetCurrentUser retrieves information about the current authenticated user using the users/current endpoint
+func (c *ConnectClient) GetCurrentUser(log logging.Logger) (*User, error) {
+	var connectUser UserDTO
+	err := c.client.Get("/__api__/users/current", &connectUser, log)
```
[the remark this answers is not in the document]
nealrichardson: If you wanted to use a proper v1 endpoint (please 🙏), it is `/v1/user`: https://docs.posit.co/connect/api/#get-/v1/user

I believe `/me` includes additional information, but you might not need that here?

internal/services/api/post_connect_token.go
```diff
@@ -0,0 +1,155 @@
...
+			PrivateKey: privateKey,
+		}
+
+		JsonResult(w, http.StatusCreated, response)
+	}
+}
+
+// PostConnectTokenUserHandlerFunc creates a handler for the POST /api/connect/token/user endpoint
+// This endpoint checks if a token has been claimed and returns the user information
+func PostConnectTokenUserHandlerFunc(log logging.Logger) http.HandlerFunc {
```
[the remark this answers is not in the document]
jonkeane: Done, thanks!

internal/clients/connect/client_connect.go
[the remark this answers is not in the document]
jonkeane: Done!

jonkeane:
@marcosnav @tdstein  @dotNomad this is ready to review again

internal/api_client/auth/token.go
```diff
@@ -0,0 +1,113 @@
...
+			return fmt.Errorf("failed to read request body: %w", err)
+		}
+
+		// Restore the request body for subsequent reads
+		req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+
+		// Standard calculation for other request bodies
+		md5Sum := md5.Sum(bodyBytes)
+		contentChecksum = base64.StdEncoding.EncodeToString(md5Sum[:])
+	} else {
```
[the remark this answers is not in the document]
jonkeane: Yeah, we still do — the hash for no-body is a hash of the emptystring and then the checksum is still used later and all of that.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -652,6 +656,71 @@ export async function newCredential(
...
+    // url should always be defined by the time we get to this step
+    const serverUrl = typeof state.data.url === "string" ? state.data.url : "";
+
+    try {
+      // Create and initialize the token activator
+      const tokenActivator = new ConnectAuthTokenActivator(serverUrl, viewId);
+      await tokenActivator.initialize();
+
+      // Activate the token
+      const result = await tokenActivator.activateToken();
```
marcosnav, line 709: I really like how this turned out, specially because the `activator` now has insurance (tests) 😄

internal/clients/http_client/http_client.go
marcosnav: This one still seems to not be used.

marcosnav (approved):
Looks pretty good. Thank you for the additional changes, I really like how pulling the token activation to it's own `class` turned out.

Just one last observation, `NewAuthenticatedHTTPClient` seems to not be used at all.

internal/clients/http_client/http_client.go
jonkeane: Aaah yes, I missed this in my last clean up, thanks for the (second) nudge 🙏 

I've gotten too used to some of the typescript helpers that yell at me when something is unused 🙈

Issue #2358, opened 2024-10-07
Re-consider how we deal with input errors


dotNomad:
Anytime we `await` a input and do validation the files can change out from underneath us. Our APIs will catch things like adding a Secret to a Configuration that is invalid, but would it be better to catch those cases in the input validation?

A follow-up: is it worth using `ignoreFocusOut: true` in a lot of our inputs since it exacerbates this problem.
(Some inputs should still have this if we expect a copy paste from somewhere else - like with an API key)

Original discussion: https://github.com/posit-dev/publisher/pull/2337#discussion_r1786878908

dotNomad:
Some other things to investigate:
- when do we want to do this?
  - mainly when the user is incentivized to edit files to fix the validation during the `await input`
- potentially show buttons to fix things for them

PR #2780, opened 2025-08-06
feat: Dismiss select deployment with focus out

dotNomad:
This PR addresses a piece of feedback from @wch while working through an issue where the expectation was that the "Select Deployment" input would dismiss when focus changed similar to other VS Code inputs.

This PR only changes the "Select Deployment" quick pick since it doesn't have any user input we expect outside of VS Code (like we do for API keys in the Credential flow).

## Intent

Part of #2358

## Approach

Removed `quickPick.ignoreFocusOut = true;` so it defaults to `false`

## User Impact

Users can now dismiss the "Select Deployment" input by clicking away from it.

## Checklist


marcosnav (approved):
piece of cake

PR #2782, opened 2025-08-06
fixes for credentials

mslynch:
## Intent

Addresses these issues:
- Credentials were not being created with the right environment because we weren't passing the header on the API call
- production Connect Cloud credentials were being interpreted as having server type of `connect`

Also adds some more cautious checking in `ConflictCheck`.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist

### DIFF
--- a/internal/services/api/post_credentials.go
+++ b/internal/services/api/post_credentials.go
@@ -61,7 +61,7 @@ func PostCredentialFuncHandler(log logging.Logger) http.HandlerFunc {
 		}
 
 		var environment types.CloudEnvironment
-		if body.ServerType == server_type.ServerTypeConnectCloud {
+		if body.ServerType.IsCloud() {
 			environment = cloud.GetCloudEnvironment(req.Header.Get(connectCloudEnvironmentHeader))
 			body.URL = cloud.GetFrontendURL(environment)
 		}


internal/credentials/credentials.go
```diff
@@ -81,21 +81,21 @@ type CredentialV0 struct {
 	ApiKey string `json:"apiKey"`
 }
 
-func (c *Credential) ConflictCheck(compareWith Credential) error {
-	if c.ServerType == server_type.ServerTypeConnectCloud {
+func (c *Credential) ConflictCheck(newCred Credential) error {
+	if newCred.ServerType == server_type.ServerTypeConnectCloud {
```
marcosnav, line 85: One suggestion, not necessarily to be done in this PR but since server type now determines code flows and will be compared in many places, WDYT of extending `ServerType` with a few methods like `ServerType.isCloud()`, `ServerType.isConnect()`?

Maybe something similar to creds? the comment below with `// this is a Connect Cloud credential` suggests that it might be possible to abstract that to have a `.isCloudCredential()` too.

PR #2786, opened 2025-08-07
First pass at multiple keyring support

### LINKED ISSUE #2783: Store each credential as it's own entry in the keychain
Follow on from #2769 

On macOS, keychain items have a maximum capacity of 3,000 bytes. With the addition of token-based auth that is quickly exhausted when we store all credentials in a single keychain file. Instead of serializing all of them into a single keychain entry, we should save each credential to a separate entry and then turn them into a list in our Go code.

jonkeane:
Resolves #2783

## Intent

Instead of storing all of the credentials in one keyring item, store each in it's own keyring item. This does mean that we need to have a separate keyring item that is a list of known items (the go-keyring package does not seem to have the ability to list and parse names of keyring items, you must get them by name).

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach
Mostly Claude

## User Impact

Be able to store more credentials (and more than _one_ legacy token-based auth method)

## Automated Tests

Added some tests, especially for the migration

## Checklist

TODO:

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -233,6 +233,7 @@ func (cr *CredentialRecord) ToCredential() (*Credential, error) {
 }
 
 type CreateCredentialDetails struct {
+	GUID       string // Optional, used for migration to preserve original GUID
 	Name       string
 	URL        string
 	ServerType server_type.ServerType
@@ -292,7 +293,12 @@ func (details CreateCredentialDetails) ToCredential() (*Credential, error) {
 		return nil, err
 	}
 
-	guid := uuid.New().String()
+	// If a GUID is provided, use it (for migration)
+	// Otherwise, generate a new GUID
+	guid := details.GUID
+	if guid == "" {
+		guid = uuid.New().String()
+	}
 	return &Credential{
 		GUID:                guid,
 		Name:                details.Name,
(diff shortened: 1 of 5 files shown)


jonkeane:
> After installation, my existing credentials are no longer visible. Should this be backwards compatible?

Yeah, the intention was to have these be migrated but retained. Do you happen to have any logs from the first time it ran?

marcosnav (commented):
I think the backwards compatibility should be left out from `keyringCredentialsService`, might be better to have this service to work only with individual credential records (the new approach) and have a migrator entity (or func) responsible for moving the old ones.

You could have the migrator entity run just before `NewKeyringCredentialsService` is called, that way you can be consistent and no need to have internal forks for `Get, Delete, Reset, etc` methods.

WDYT?

jonkeane:
Thanks for the feedback! Strong agree that this is a better approach and I've done that here.

I also confirmed that it migrated my credentials just fine locally (there is no migration on Positron Pro, but I did confirm that it works there too). I was also able to store more than one token too. 

Mind taking another look?

PR #2789, opened 2025-08-08
chore: Remove unused envVarProvider

dotNomad:
Noticed that this bit of code is unused after digging into a question about Publisher.

This PR just removed the unused `envVarProvider` and its tests.


mslynch (approved):

sagerb:
>It looks like the AccountSourceEnvironment type isn't completely unused, and is a more difficult clean-up. Do you think that its required for this PR?

No, I suppose not. It just seemed like it would complete the removal... ok, I'll approve it.

sagerb (approved):
I'm ok with just removing these for now and leaving the other removals for a future PR.

PR #2791, opened 2025-08-08
chore: Remove remnents of credentials from environment variable support

sagerb:
This PR follows on the heels of https://github.com/posit-dev/publisher/pull/2789#pullrequestreview-3102130529, as it is cleaning up the dead code related to supporting environment variables as a credential.

To verify, I have run the normal building steps and then executed the existing e2e tests, which include validation for credential support. If CI passes, we should be good with the removal of this code.

### DIFF
--- a/internal/state/state.go
+++ b/internal/state/state.go
@@ -61,13 +61,7 @@ func getDefaultAccount(accountList []accounts.Account) (*accounts.Account, error
 	if len(accountList) == 0 {
 		return nil, errNoAccounts
 	} else if len(accountList) > 1 {
-		// If an account was provided via environment variables, use it.
-		for _, acct := range accountList {
-			if acct.Source == accounts.AccountSourceEnvironment {
-				return &acct, nil
-			}
-		}
-		// Otherwise we don't have a way to choose
+		// we don't have a way to choose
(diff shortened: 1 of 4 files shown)


internal/state/state.go
```diff
@@ -61,13 +61,7 @@ func getDefaultAccount(accountList []accounts.Account) (*accounts.Account, error
 	if len(accountList) == 0 {
 		return nil, errNoAccounts
 	} else if len(accountList) > 1 {
-		// If an account was provided via environment variables, use it.
-		for _, acct := range accountList {
-			if acct.Source == accounts.AccountSourceEnvironment {
-				return &acct, nil
```
dotNomad, line 67: A bit of digging into this, it looks like the only place we make accounts is in the Credentials providers which loads the file or keychain credentials. Correct my if I'm wrong, but that should mean we never see `AccountSourceEnvironment` account types here.

dotNomad (approved):
Thank you for the additional cleanup!

internal/state/state.go
sagerb: Totally correct, that's why it was ok to remove this code.

PR #2816, opened 2025-08-14
fix keyring ForceSet implementation to not create duplicate creds

mslynch:
## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist

### DIFF
--- a/internal/credentials/keyring.go
+++ b/internal/credentials/keyring.go
@@ -96,20 +96,35 @@ func (ks *keyringCredentialsService) set(credDetails CreateCredentialDetails, ch
 		return nil, err
 	}
 
+	guidToUpdate := cred.GUID
 	if checkConflict {
 		err = ks.checkForConflicts(&table, cred)
 		if err != nil {
 			return nil, err
 		}
+	} else {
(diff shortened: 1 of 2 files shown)


internal/credentials/keyring.go
```diff
@@ -96,20 +96,35 @@ func (ks *keyringCredentialsService) set(credDetails CreateCredentialDetails, ch
...
+			}
+
+			if tableCred.Name == credDetails.Name {
+				guidToUpdate = tableCred.GUID
+				break
+			}
+		}
 	}
 
+	cred.GUID = guidToUpdate
```
rodriin, line 120: Is this reassignment really needed?

rodriin (approved):
Looks good, just had the one question.

internal/credentials/keyring.go
mslynch: I think so, otherwise the credential object will have a guid that doesn't match the table key.

PR #2826, opened 2025-08-15
chore: Improve bug template with descriptions

dotNomad:
I noticed that our Issue Template for bugs was a bit sparse.

It didn't include instructions on where to find some of the things we were requesting like the version of the extension.

It didn't ask whether the issuer was using VS Code or Positron.

This swaps the bug-report template to use the [GitHub `yaml` syntax](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema) so the fields are inputs specifying what is required and what isn't. The IDE option is even a dropdown ✨ 

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [x] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## User Impact

It is now much easier to fill out a bug report!

## Directions for Reviewers

You can view the `bug-report.yml` file in a rich way using the "View file" dropdown option:

CleanShot 2025-08-15 at 11 56 03

This gives a better indication of how the template will be seen.


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,89 @@
...
+        Please provide the following information to help us diagnose the issue.
+
+  - type: input
+    id: publisher-version
+    attributes:
+      label: Publisher Version
+      description: "Find this in Positron/VS Code: Extensions view → Posit Publisher → under Installation on the right"
+      placeholder: "e.g., 1.18.0"
+    validations:
+      required: true
```
jonkeane, line 18:
```suggestion
      required: false
```

IMHO we shouldn't _require_ any of these fields. We don't want to put up barriers to people submitting things (even if that information would be very helpful to us)

dotNomad: Done in https://github.com/posit-dev/publisher/pull/2826/commits/a29b22254ed81801a8e9ad5a2d0ddad4a2ce9766

Issue #1347, opened 2024-04-10
main CI doesn't use the tag when building


mmarchetti:
From the [main build of 1.0.alpha5](https://github.com/posit-dev/publisher/actions/runs/8637664088/job/23680619957#step:6:16)

```
Run just build
Command: ./cmd/publisher
Version: 1.0.alpha4-311-g0e99958e
Name: publisher
Mode: prod
Building: darwin/amd64
Executable: /work/bin/darwin/amd64/1.0.alpha4-311-g0e99958e/publisher
Building: darwin/arm64
Executable: /work/bin/darwin/arm64/1.0.alpha4-311-g0e99958e/publisher
Building: linux/amd64
Executable: /work/bin/linux/amd64/1.0.alpha4-311-g0e99958e/publisher
Building: linux/arm64
Executable: /work/bin/linux/arm64/1.0.alpha4-311-g0e99958e/publisher
Building: windows/amd64
Executable: /work/bin/windows/amd64/1.0.alpha4-311-g0e99958e/publisher.exe
```

The version number should be 1.0.alpha5.

PR #2863, opened 2025-08-20
Updating the polling interval default to 2 seconds

### LINKED ISSUE #2777: Connect Cloud device auth polling interval is slow
The device auth polling interval, currently five seconds, can create a pretty noticeable gap between clicking authorize and seeing the extension move to the next step. Maybe we could consider reducing it, or ask for the interval returned to us to be reduced.

rodriin:
## Intent

Resolves #2777 

This PR updates the polling interval default to be every 2 seconds to match the updated device auth polling interval in this PR: [#1347](https://github.com/rstudio/lucid-auth/pull/1347)

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The new polling interval will apply to all the polling done for Connect Cloud like the device auth token polling and the account polling.

## User Impact

More responsive multi-stepper experience when creating a new credential for Connect Cloud.

## Automated Tests

Existing E2E tests succeeded locally for this branch.

## Directions for Reviewers

Verify creating a new credential for Connect Cloud is faster.

## Checklist

### DIFF
--- a/extensions/vscode/src/multiStepInputs/multiStepHelper.ts
+++ b/extensions/vscode/src/multiStepInputs/multiStepHelper.ts
@@ -406,8 +406,8 @@ export class MultiStepInput {
         // eslint-disable-next-line no-async-promise-executor
         async (resolve, reject) => {
           let abortPolling = false;
-          // default the polling interval to 5 seconds
-          pollingInterval ||= 5000;
+          // default the polling interval to 1 second
+          pollingInterval ||= 1000;
           const input = window.createInputBox();
           input.title = title;
           input.step = step;
@@ -490,8 +490,8 @@ export class MultiStepInput {
             }
           };
 
-          // start polling (poll every 5 seconds at most 120 times === 10 minutes max time)
-          return pollApiFunction(pollingInterval, 120)
+          // start polling (poll every 1 second at most 600 times === 10 minutes max time)
+          return pollApiFunction(pollingInterval, 600)
             .then(resolve)
             .catch(reject);
         },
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/multiStepInputs/multiStepHelper.ts
```diff
@@ -406,8 +406,8 @@ export class MultiStepInput {
         // eslint-disable-next-line no-async-promise-executor
         async (resolve, reject) => {
           let abortPolling = false;
-          // default the polling interval to 5 seconds
-          pollingInterval ||= 5000;
```
hekhuisk, line 410: Per spec https://datatracker.ietf.org/doc/html/rfc8628#section-3.2:
>  interval
      OPTIONAL.  The minimum amount of time in seconds that the client
      SHOULD wait between polling requests to the token endpoint.  If no
      value is provided, clients MUST use 5 as the default.

So if you aren't given an interval this should remain 5 seconds.

rodriin: This is used for polling the device auth and the accounts endpoint in `vivid-api`. For the device auth there will be a value so the default will never apply. I added the default for the accounts polling. I don't think there is a harm in calling that one every 1 second, but I'll make it match whatever we decide the default to be for the device auth polling so the user experiences a similar polling time for both endpoints we are polling. I saw you proposed 3 seconds for the device auth. I'll adjust this one to 3 seconds if we agree to that.

Issue #2773, opened 2025-08-05
Generate the dependency portion of R manifests from renv.lock directly


jonkeane:
With https://github.com/rstudio/renv/pull/2059 the renv lockfiles should have all the information we need to construct manifests without needing to have a working install of that exact version in an active library. We should update our manifest generation code to do this directly. If we can't, we should raise issues with renv and resolve those too.

Related to (and would likely solve some of the frustration described in) #2646 but the discussion there took a different path (discussion around interpreters) and don't want to overload that issue with this idea.

Issue #2838, opened 2025-08-19
feat: reduce dependency on a renv managed library


amol-:
## Description

We should handle three cases:

1. A `renv.lock` exists, the `renv.lock` is used as is, independently from the state of the library of the selected interpreter
2. There is no `renv.lock` we should detect project dependencies automatically
3. There should be a button to Scan for dependencies and generate a `renv.lock` for the user based on **2**

PR #2898, opened 2025-08-21
feat: Support for parsing the Manifest from the Lockfile without a library

amol-:
## Intent

Implement direct conversion from `renv.lock` to `manifest.json` without requiring a working R library.
This is a building block for https://github.com/posit-dev/publisher/issues/2838

Closes https://github.com/posit-dev/publisher/issues/2773

## Type of Change

- - [x] New Feature 

## Approach

A new `GetManifestPackagesFromLockfile` function is added, the codebase still preserves the old `GetManifestPackages` so that the behaviour of the two can be compared for backward compatibility.

## User Impact

This should provide a more robust experience as the manifest generation shouldn't depend anymore from the state of the R library. 

## Automated Tests

A new `TestLockfilePackageMapperSuite` is added, with some tests that compare results from `GetManifestPackages`  to those from `GetManifestPackagesFromLockfile`

## Directions for Reviewers

Reviews can be started from `func (p *defaultPublisher) getRPackages` that's more or less the entry point to the feature

## Checklist


jonkeane:
Ok, sorry this took me a second to take a look. I think I see what's going on here. I believe the lockfile that is generated is coming from here, yeah?

https://github.com/posit-dev/publisher/blob/20bb103b6dc6065374b7d63dd44df52499214d1f/internal/publish/publish_test.go#L88-L111

That is the old-style of renv.lock file for renv.lock files before renv 1.1.0 where there are fewer fields in the lock file. For now (in this PR, at least), we can require that renv.lock files be made with renv>=1.1.0. We might need to come back and figure out what to do if we can do anything with renv<1.1.0. Maybe we should update that renv fixture there. Or maybe our legacy code is stripping away the old-style of renv lock files.

internal/interpreters/r.go
```diff
@@ -51,6 +51,7 @@ type RInterpreter interface {
 	GetPreferredPath() string
 	CreateLockfile(util.AbsolutePath) error
 	RenvEnvironmentErrorCheck() *types.AgentError
+	IsRenvInstalled(rexecPath string) *types.AgentError
```
amol-, line 54: I exposed this because the `ScanDependencies` of  `LockfilePackageMapper` does not require a working R Library, it only needs the `renv`  package to be isntalled.

internal/publish/interpreters.go
```diff
@@ -34,6 +34,10 @@ func (p *defaultPublisher) configureInterpreters() error {
 		if err != nil {
 			return err
 		}
+		// Enforce modern renv.lock format when publishing
+		if err := renv.ValidateModernLockfile(lockfile); err != nil {
+			return err
+		}
```
amol-, line 40: In the end this is not strictly required, but to make things more consistent it is better to restrict renv.lock format to >= 1.1.0 it's a safe trade-off, asking users to only update renv seems reasonable.

amol-:
@dotNomad ready for review 👍

internal/inspect/dependencies/renv/lockfile.go
```diff
@@ -55,6 +83,25 @@ func ReadLockfile(path util.AbsolutePath) (*Lockfile, error) {
...
+// ValidateModernLockfile enforces requirements for renv >= 1.1.0 lockfiles.
+// We require modern lockfiles because they provide resolvable repository URLs
+// in the top-level Repositories section, enabling consistent package deployment
+// across different environments without relying on local R configuration.
+func ValidateModernLockfile(lockfile *Lockfile) error {
+	if len(lockfile.R.Repositories) == 0 {
+		return fmt.Errorf("renv.lock is not compatible, missing Repositories section. Regenerate the lockfile with renv >= 1.1.0")
+	}
+	return nil
+}
```
dotNomad, line 95: Error came through to the frontend really well here.

dotNomad (commented):
Fantastic comments it helped tremendously with the review.

When enabling the feature manually using `true` in `publish.go` it worked as I expected it to. I tested with a workspace at the project level and with sub-directory setups.

I will note that I got _significant_ "Restoring Environment" times (30+ minutes) using this with Connect though, even when re-deploying the same project. I'm not sure if that is something that is occurring because of how the manifest is being generated, but its worth looking into. I didn't notice anything really specific in the code changes, but I did compare a `manifest.json` generated for a Gallery item (specifically: https://github.com/posit-dev/connect-extensions/tree/main/extensions/usage-metrics-dashboard) and the `manifest.json` generated from this feature and they were different.

Some differences I noticed:
- additional fields like `RemoteType`, `RemoteRef`, etc in this feature
- text fields like `Authors` included line breaks `\n` in this feature

Maybe we can pair a bit to identify why "Restoring Environment" takes so much longer, or perhaps you have an idea?

---

Also pointed this at Connect Cloud and it worked great.

PR #2904, opened 2025-08-21
feat: Supporting work for Quarto deployments to include an optional static type configuration

### LINKED ISSUE #2851: Quarto Deployments: Detect the configuration alternative for static/rendered content on project inspection
Provide the configuration alternative for Quarto when `POST inspect` is called by including a new alternatives field on the configuration:
```go
type Config struct {
    //...
    Alternatives      []Config      `toml:"-" json:"alternatives,omitempty"`
```

This new field should be populated by the `QuartoDetector` when there is a possible static version of the content. Such static version configuration should:
1. populate the `Files` attribute with the applicable rendered version of the project files.
2. `Type` should be `ContentTypeHTML` for the rendered version

For this work, it is important to rely heavily on what `quarto inspect` output provides for `output-dir` and `files` (the `files` JSON data is already filtered down by the possible presence of `_quarto.yml#project.render`).

If there is not an `output-dir` stated, an `.html` naming version of files listed under `files` should be used.

marcosnav:
This PR introduces supporting work to inspection of Quarto projects in the agent so it generates and provides an alternative configuration to deploy the static version (renderings) of Quarto projects.

## Intent

Resolves #2851 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Leverage on `quarto inspect` output metadata to determine which files are related to generated renderings.

## User Impact

None yet, this is supporting work that is not user facing at the moment.

## Automated Tests

Updated the Quarto inspection integration tests that read `quarto inspect` output from `.json` files and assert the resulting configuration that is passed to the extension is well formed.

## Directions for Reviewers

Cannot be verified manually, yet.

## Checklist


internal/inspect/detectors/quarto.go
```diff
@@ -42,166 +41,17 @@ func NewQuartoDetector(log logging.Logger) *QuartoDetector {
 	}
 }
 
-type quartoMetadata struct {
-	Title   string `json:"title"`
-	Runtime string `json:"runtime"`
-	Server  any    `json:"server"`
-}
-
-type quartoProjectConfig struct {
```
marcosnav, line 51: Moved related `quarto inspect` output objects to it's own file.

internal/config/types.go
```diff
@@ -74,6 +74,7 @@ func (t ContentType) IsAppContent() bool {
 
 type Config struct {
 	Comments            []string      `toml:"-" json:"comments,omitempty"`
+	Alternatives        []Config      `toml:"-" json:"alternatives,omitempty"`
```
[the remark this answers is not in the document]
dotNomad: On first glance I'm a bit confused by the need for the new key.

We already have support for alternative configurations through our `/inspect` API. I'm guessing these will be pulled out and presented in that array JSON response?

I'm guessing (since this is the first file in the PR) that our detectors currently return a single result, our `/inspect` API returns multiple if multiple inspectors return a valid config, and this is a way around that allowing the Quarto detector to return multiple.

Let me know if I'm on the right track understanding this.

internal/inspect/detectors/quarto.go
dotNomad: Thanks that was helpful 😄

internal/inspect/detectors/quarto_inspect_output.go
```diff
@@ -0,0 +1,236 @@
...
+	} `json:"website"`
+}
+
+type quartoFilesData struct {
+	Input           []string `json:"input"`
+	ConfigResources []string `json:"configResources"`
+}
+
+type quartoInspectOutput struct {
+	fs afero.Fs
```
dotNomad, line 40: This was an interesting inclusion. Feels like a good opportunity for me to learn more, can you explain why this was added?

dotNomad (approved):
I am really liking the file separation here. Reading through this seems like a great start to the quarto static support.

I did have a two questions to verify my understanding.

internal/config/types.go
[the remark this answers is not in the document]
marcosnav: > and this is a way around that allowing the Quarto detector to return multiple.

That's right, we can get multiple configurations already but the goal here is to display the alternatives in a subsequent step, after the user confirms they want to use Quarto as the project type, then we show the alternative to publish the static version of the project. 

It might be confusing and I wasn't 100% convinced on this, it works but still I'm open to suggestions or happy to change down the road how we surface the `static` version of the configuration.

internal/inspect/detectors/quarto_inspect_output.go
marcosnav: Sure thing, there are some methods in `quartoInspectOutput` that handle system paths in different ways and methods like `util.NewAbsolutePath` require the file system provider to be passed along. Although it can be `nil` and paths will pick a default `fs`, this gives flexibility specially to instantiate a fake memory file system in tests when needed.

Issue #2858, opened 2025-08-19
Scroll to failure in Publisher logs when opening


dotNomad:
When we open the Publisher logs it would be helpful to scroll to the failure if one exists to get users to the most relevant part of the logs as soon we can.

We will need to investigate if this is possible:

There is a [`TreeView.reveal`](https://code.visualstudio.com/api/references/vscode-api#TreeView.reveal) method that looks very promising.

We should be able to use this when clicking one of our "view logs" buttons, but we may not be able to get it when a user clicks on the "Publisher" logs view manually.

PR #2917, opened 2025-08-22
feat: Revealing first failing stage in Publisher logs

### LINKED ISSUE #2857: Auto expand failing section of Publisher logs
In the image below the "Validate Deployment Record" is expanded, but it is not _always_ expanded when the logs are opened.

The `LogsTreeStageItem` remembers if it has been expanded or collapsed previously. When a new deployment is kicked off and a section fails we should try to expand that content to show the most helpful logs to users as quickly as we can.

Image

dotNomad:
This PR introduces a new feature: when the "posit-publisher-logs" panel `TreeView` is revealed the first failing deployment stage (if one exists) will be revealed.

Revealing it:
- scrolls it into view
- expands other `TreeViewItems` if necessary
- expands the `LogsTreeStageItem` itself

## Intent

Resolves #2857 #2858

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach was to use the [TreeView `reveal()` method](https://code.visualstudio.com/api/references/vscode-api#TreeView.reveal) alongside an [`onDidChangeVisibility` listener](https://code.visualstudio.com/api/references/vscode-api#TreeView.onDidChangeVisibility) on the TreeView.

When the TreeView is shown we try to grab the first `LogsTreeStageItem` has the `LogStateStatus.failed` status.

If one exists we `reveal()` it.

To accomplish this the two `TreeItem` types (`LogsTreeItemLogItem` and `LogsTreeStageItem`) needed to track their `parent` `TreeItem`s to implement `TreeDataProvider`'s optional `getParent` method which is necessary for the `reveal()` method to function.

With all of this in place the Publisher Logs reveal the first failing stage.

## User Impact

When opening the Publisher Logs the first stage is shown quickly getting users to the most relevant bit of the logs (a failure) as quickly as possible

## Directions for Reviewers

Deploy a project without an error and check that the logs view behaves normally.

Deploy a project that causes a failure in the logs and open the Publisher logs using any method you'd like. The failing stage should be in view even if the logs are very small.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   easily copy system information - extension version, IDE, and platform details
   (#2835)
 - Added a delete button for Credentials in the Credentials View (#2862)
+- When opened, the Publisher logs scroll to and expand the first failed stage
+  if a stage is marked as failed (#2857, #2858)
 
 ### Changed
 
--- a/extensions/vscode/src/views/logs.ts
+++ b/extensions/vscode/src/views/logs.ts
@@ -411,15 +411,33 @@ export class LogsTreeDataProvider implements TreeDataProvider<LogsTreeItem> {
   }
 
   /**
-   * Returns the first LogsTreeStageItem with `LogStageStatus.failed`
-   * Returns undefined if no LogsTreeStageItems have the `LogStageStatus.failed`
+   * Returns either
+   * the first LogsTreeStageItem with `LogStageStatus.failed` or
+   * the last LogsTreeLogItem of that stage if the stage has events
+   *
+   * Returns `undefined` if no LogsTreeStageItems have `LogsStateStatus.failed`
    */
-  private findFirstFailureItem(): LogsTreeStageItem | undefined {
+  private getTreeItemToReveal(): LogsTreeItem | undefined {
     const root = new LogsTreeStageItem(this.publishingStage);
 
     for (const stage of this.publishingStage.stages) {
       if (stage.status === LogStageStatus.failed) {
-        return new LogsTreeStageItem(stage, root);
+        const stageItem = new LogsTreeStageItem(stage, root);
+
+        if (stage.events.length) {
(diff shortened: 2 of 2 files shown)


jonkeane:
I just tried this out in Positron on Workbench, and it does indeed open + scroll, though I'm not sure that the _first_ error is the one we want to scroll to. 

I uploaded one of our examples, but missing a sidecar file (a common mistake + also one that should be easy to ID with the logs).

This is the first view I got after clicking `View Logs` in the toast:

Screenshot 2025-08-25 at 08 05 06

A few comments: (1) most of the logs view is taken up with a long list of things that went fine. And the error that's there is not what I expected (something about our new NSS module). Which clearly wasn't a _catastrophic_ error, because if I scrolled down to the end I see what I was expecting (below). Would it be possible to get closer to this second view? Possibly by scrolling to the last error and not the first (or maybe those NSS errors are being mis-identified as errors?) and also possible to have the scroll be such that the error line is at the ~ top of the screen rather than the bottom? That way we get as much of the traceback / context up and visible as possible?

Screenshot 2025-08-25 at 08 05 31

jonkeane:
> > Would it be possible to get closer to this second view? Possibly by scrolling to the last error and not the first

> To clarify this scrolls the first failing stage into view, not the first error in the individual items inside the stages.

> > (or maybe those NSS errors are being mis-identified as errors?) and also possible to have the scroll be such that the error line is at the ~ top of the screen rather than the bottom? That way we get as much of the traceback / context up and visible as possible?

> I'll investigate to see if the reveal() API has some options for that flexibility and see what details we have in the individual log items that could allow a more targeted reveal()

Aaaah ok ok I got it now, we're not parsing the lines themselves, just the stages. Yeah ok, so what we actually want here is the failing stage to be in view _and_ scrolled to the bottom of that stage then, right?

jonkeane:
Aaah yes, this is so much nicer (and shows exactly the error that I should care about without scrolling).

https://github.com/posit-dev/publisher/issues/2922

Issue #2878, opened 2025-08-20
Connect Cloud: show application logs when app fails to start up


mslynch:
## Description

In Connect Cloud, our logs are tagged as either `build` or `runtime`. As part of the deployment process in Publisher, we stream the `build` logs while the content is deploying. One publishing failure scenario is that the app crashes immediately upon startup, which will cause a publish to fail. In that case, the app will likely output useful information about why it crashed to the `runtime` log. We should be surfacing that information in Publisher.

mslynch:
Since each new revision gets a new log channel, we should be able to show both build and runtime logs. Let's try that and see how it looks - if it's confusing, we can table this for now.

tyatposit:
verified.

Confirmed this with a few different project types and a variety of missing files. The messaging is clear and available for the user to find for all scenarios I tried.

Still don't love that we can't actually copy the info, but that could be revisited in the future.

PR #2923, opened 2025-08-25
display runtime logs and keep fetching for 5 seconds after publish completes

mslynch:
## Intent

Resolves https://github.com/posit-dev/publisher/issues/2878.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

We can show runtime logs just fine in addition to build logs... But since the logs in question happen at the very end of the publishing process, and our logs stream is independent of our publishing status updates, they can get cut off if the status updates before the logs can come through. Here I introduce a 5 second delay before stopping the logs client.

## User Impact

Users will see the runtime error that caused their app to fail, or if their app starts, the first few seconds of app logs.

## Automated Tests

Not sure how to test this... I'm guessing it's not worth the effort to figure out.

## Directions for Reviewers

Deploy an app with a critical file missing, like https://github.com/posit-hosted/examples-shiny-python with data/tips.csv missing. The logs should never be cut off at the end.

## Checklist

### DIFF
--- a/internal/publish/connect_cloud/publish_to_server.go
+++ b/internal/publish/connect_cloud/publish_to_server.go
@@ -88,7 +88,8 @@ func (c *ServerPublisher) PublishToServer(contentID internal_types.ContentID, bu
 	ctx, cancel := context.WithCancel(context.Background())
 	defer func() {
 		go func() {
-			// Allow up to 5 seconds for logs to flush before cancelling
+			// Allow up to 5 seconds for logs to flush before cancelling.
+			// Needs goroutine to avoid blocking the completion of the deployment.
 			time.Sleep(5 * time.Second)
 			cancel()
 		}()


rodriin (approved):
This seems to be working but I'm not sure how I feel about it. On one hand it is nice to see the error, but on the other hand that's a lot of logs to scroll through. I guess I need to see this once one of my PRs is merged that auto selects and highlights the error line in the Publisher logs. Maybe it will feel better then.

This is the other PR I'm referring to: Tweaking the wording on the cancel deploy info message and some other minor fixes #2931

This what the logs look like for me and the scrolling involved:

https://github.com/user-attachments/assets/57e1533c-ad5c-4605-9a2b-15c7ce40ea2e

tyatposit:
verified.

Confirmed this with a few different project types and a variety of missing files.  The messaging is clear and available for the user to find for all scenarios I tried.  

Still don't love that we can't actually copy the info, but that could be revisited in the future.

Issue #2912, opened 2025-08-22
bug: `python-fastapi` not support by Connect Cloud error


dotNomad:
### Publisher Version

(developer environment)

### IDE

VS Code

### Content Type to Deploy

FastAPI

### Problem Statement

When selecting a `simple.py` FastAPI entrypoint for deployment to Connect Cloud I get the error "Content type `python-fastapi` is not support by Connect Cloud"

I would expected our inspection to select the right type and be able to deploy this to Connect Cloud.

### Steps to Reproduce

1. Open our `sample-content/fastapi-simple/simple.py` as an entrypoint
2. Select a Connect Cloud credential for the deployment destination
3. Attempt to publish

<img width="914" height="218" alt="Image" src="https://github.com/user-attachments/assets/dc5a26bc-1e20-4eb5-9b54-66f4f1c0c7f3" />

rodriin:
We should be able to detect an unsupported content type while creating the deployment.

PR #2927, opened 2025-08-25
validate connect cloud type field on deployment creation

mslynch:
## Intent

Error on deployment config creation if using Connect Cloud with an unsupported product type, rather than waiting until deploying to error.

Resolves https://github.com/posit-dev/publisher/issues/2912.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Reuses an existing utility function to determine if the type is supported.

## User Impact

Users will see the error when they create their content rather than when they attempt to publish it.

## Automated Tests

Added a test case for the validation error.

## Directions for Reviewers

Attempt to create a fastapi deployment - it should error.

## Checklist

### DIFF
--- a/internal/clients/types/types.go
+++ b/internal/clients/types/types.go
@@ -1,6 +1,11 @@
 package types
 
-import "github.com/posit-dev/publisher/internal/types"
+import (
+	"fmt"
+
+	"github.com/posit-dev/publisher/internal/config"
+	"github.com/posit-dev/publisher/internal/types"
+)
 
 // Copyright (C) 2025 by Posit Software, PBC.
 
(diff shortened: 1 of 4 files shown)


rodriin (approved):
🚀

mslynch:
I'm starting to wonder if this is the right solution. If our app mode detection picks the wrong mode, and you should really be able to deploy, this will cause problems.

mslynch:
closing this PR - we're going to take the approach described in https://github.com/posit-dev/publisher/issues/2912#issuecomment-3225487878 instead.

PR #2930, opened 2025-08-25
feat: Use hostname as initial credential name

### LINKED ISSUE #2922: feat: take as the default name for a credential the server URL
Drop the https and just use it. Seems like we could streamline the new credential setup by not requiring the user to type something in here.

dotNomad:
When creating a Connect credential we ask the user to input their server URL.

This PR changes that making the default value to the hostname of the URL they gave us in the earlier input step to reduce the amount of user input necessary.

## Intent

Resolves #2922

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Grabbed the `url` from the `MultiStepState` and attempt to URL parse it to grab the `url.hostname`.

If that parsing doesn't work fallback to the current behavior where there isn't a default.

## User Impact

Reduces friction when creating a Connect credential by reducing user input.

## Directions for Reviewers

Create a Connect Credential and see the default value filled in for the name!

## Checklist

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
@@ -508,14 +508,13 @@ export async function newConnectCredential(
     input: MultiStepInput,
     state: MultiStepState,
   ) {
-    if (typeof state.data.url === "string") {
+    if (!state.data.name && typeof state.data.url === "string") {
       try {
         const url = new URL(state.data.url);
         // Default the name to the hostname of the given URL
         state.data.name = url.hostname;
       } catch {
-        // If URL parsing fails, fall back to empty string
-        state.data.name = "";
+        // If URL parsing fails, leave name unchanged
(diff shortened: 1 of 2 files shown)


extensions/vscode/src/multiStepInputs/newConnectCredential.ts
```diff
@@ -508,6 +508,16 @@ export async function newConnectCredential(
     input: MultiStepInput,
     state: MultiStepState,
   ) {
+    if (!state.data.name && typeof state.data.url === "string") {
```
dotNomad, line 511: Currently `state.data.name` never gets set, but I figured it was good practice to check for this in case another step is created or this step moves for some reason.

jonkeane (approved):
This is simple enough, but could this use a test?

PR #2950, opened 2025-08-27
cloud content type validation

mslynch:
## Intent

Resolves https://github.com/posit-dev/publisher/issues/2912.

Shows a validation error when choosing an unsupported content type for cloud in the homeview:

Screenshot 2025-08-27 at 13 19 51

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

Added the validation in FromFile using existing code to check the supported types. It's important that we don't call this when we write t

I had to shuffle some things around to get around circular import issues.

## Automated Tests

Added a new FromFile test that tests this behavior.

## Directions for Reviewers

Create a new fastapi deployment with a cloud credential and verify it shows the error.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -166,7 +166,7 @@ Run the following commands from the `test/e2e` subdirectory:
 source .venv/bin/activate
 ```
 
-To access the test user credentials for PCC, you must authenticate with the 1Password CLI before running E2E tests. Request access to this 1Password [publisher vault](https://positpbc.1password.com/app#/7qnaxd5rhjcigmkv5nunrx33p4).
+To access the test user credentials for PCC, you must authenticate with the 1Password CLI before running E2E tests. Request access to the 1Password Publisher vault if needed. A direct link to the vault is available in the #Publisher Slack channel bookmarks.
 
 If you have the 1Password CLI (`op`) installed globally (in your PATH), run:
 


internal/config/config.go
```diff
@@ -76,9 +78,26 @@ func FromFile(path util.AbsolutePath) (*Config, error) {
 		return nil, err
 	}
 	cfg.PopulateDefaults()
+
+	err = validate(cfg)
+	if err != nil {
+		return nil, err
+	}
+
```
mslynch, line 86: This is the important part.

internal/bundles/manifest.go
```diff
@@ -205,9 +206,9 @@ func NewManifestFromConfig(cfg *config.Config) *Manifest {
 		}
 	}
 	switch cfg.Type {
-	case config.ContentTypeRMarkdown, config.ContentTypeRMarkdownShiny:
+	case contenttypes.ContentTypeRMarkdown, contenttypes.ContentTypeRMarkdownShiny:
```
rodriin, line 209: This is very nit picky but my eyes were not digesting `contenttypes` easily instead I kept expecting `contentTypes`. Feel free to ignore this, you probably had a good reason for going this route.

internal/clients/types/types.go
```diff
@@ -136,3 +141,24 @@ type AuthorizationResponse struct {
...
+	case contenttypes.ContentTypePythonBokeh:
+		return ContentTypeBokeh, nil
+	case contenttypes.ContentTypePythonDash:
+		return ContentTypeDash, nil
+	case contenttypes.ContentTypePythonShiny, contenttypes.ContentTypeRShiny:
+		return ContentTypeShiny, nil
+	case contenttypes.ContentTypePythonStreamlit:
+		return ContentTypeStreamlit, nil
+	case contenttypes.ContentTypeQuartoDeprecated, contenttypes.ContentTypeQuarto, contenttypes.ContentTypeHTML:
+		return ContentTypeQuarto, nil
```
rodriin, line 159: Is this the case that will get split into Quarto and Static once `static` is introduced in `vivid-api`?

internal/config/config.go
```diff
@@ -76,9 +78,26 @@ func FromFile(path util.AbsolutePath) (*Config, error) {
...
 	return cfg, nil
 }
 
+// validate performs additional validation beyond schema validation. This validation is intended only to apply to an
+// existing config file, not to a new config file being created.
+func validate(cfg *Config) error {
+	if cfg.ProductType.IsConnectCloud() {
+		// Don't allow app modes that don't map to a Cloud content type
+		_, err := types.CloudContentTypeFromPublisherType(cfg.Type)
+		return err
```
rodriin, line 96: Nice looks like `err` will either be `nil` or an actual error! However, if we error out here do we still let them create the new deployment? Or is this just the bit that reports the error so we can show the warning in the UI?

rodriin (approved):
Wow big refactor for the renames.

internal/bundles/manifest.go
mslynch: Someone told me a long time ago that Go package names aren't supposed to have underscores or capital letters, but I never saw the official recommendation for myself... I went to dig it up and found that [Go's documentation](https://go.dev/blog/package-names#package-names) leans that way without making an explicit guideline, but [Google's Go style guide](https://google.github.io/styleguide/go/decisions.html#package-names) makes it a rule. I  took a look in the Connect codebase and found that it also follows this pattern.

internal/clients/types/types.go
mslynch: Yep!

internal/config/config.go
mslynch: We still let them create the deployment, yeah. This path is only executed when we read from the config file, which happens through a few endpoints - but not the one where we create the file.

rodriin: Very nice, great job!

PR #2952, opened 2025-08-27
feat: add celebratory message when cred token is connected

### LINKED ISSUE #2901: feat: more celebration when tokens are connected
## Description
At the end of adding a token (when you're adding a name), we should have more celebration than we have now.

fizzboop:
## Intent
Resolves #2901

## Type of Change
- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## User Impact
Gives users a small celebratory notification that they successfully connected their credential token.

<details>
<summary>Screencapture</summary>

https://github.com/user-attachments/assets/0eccc954-ad55-4a6e-9dee-c309bd7a4f92

</details>

## Directions for Reviewers
1. Create a new credential and follow the steps.
2. After naming the credential, a toast on the bottom right corner should appear.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   on a failed deploy (#2860)
 - Use server hostname as the default credential name for new Connect credentials
   (#2922)
+- Added a celebratory message when credential token is connected (#2901)
 
 ### Changed
 
--- a/extensions/vscode/src/multiStepInputs/newCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newCredential.ts
@@ -6,6 +6,7 @@ import {
   MultiStepInput,
   MultiStepState,
 } from "./multiStepHelper";
+import { window } from "vscode";
 import { Credential, ServerType, ProductName } from "src/api";
 import { extensionSettings } from "src/extension";
 import { platformList } from "src/multiStepInputs/common";
@@ -167,6 +168,11 @@ export async function newCredential(
     // navigating backwards and then forward in the multi-stepper steps
     throw new AbortError();
   }
+
+  window.showInformationMessage(
+    // To prevent a large refactor for the flow, we added the as Credential declaration
+    `Successfully added ${(credential as Credential).name} 🎉`,
+  );
 
   return credential;
 }


extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -168,5 +169,10 @@ export async function newCredential(
     throw new AbortError();
   }
 
+  window.showInformationMessage(
+    // To prevent a large refactor for the flow, we added the as Credential declaration
+    `Successfully added ${(credential as Credential).name} 🎉`,
```
dotNomad, line 174: I noticed that the video said "added a credential" and not the name?

Additionally if this is what we were thinking with the celebratory message lets refactor this a bit to avoid the type assertion here. Before we do that work lets make sure this is what @jonkeane was envisioning.

jonkeane: _nods_ yeah, I was intending #2901 to be a celebration after the token has been successfully claimed, but before someone has put the name in. You can see in the video when you're on the setting-the-nickname-screen there's a toast that says "Successfully authenticated as byonca_honaker" today. That's a little too bland: most users won't realize that that means the connection was successful and therefore they are on the next step of making a nickname. Also, because it is a toast and so far away from the text input box is not great — that's a lot of distance for someone to connect the dots between these two things. 

I was hoping the celebration would happen in the nickname picker (or at least is closer to there visually speaking) so it's clear "you've done the thing, now you just need to pick a nickname". The VScode UI here isn't doing us any favors because the box of typing in the URL, the box that is waiting for the token to be activated, and then the nickname box are all *very similar* so it's not clear that you're making progress through the process. 

Having a celebration at the end when the credential is created might be a nice to have (though frankly I think we have too many toasts and vscode generally has too many toasts so adding another toast to that fire isn't great). But we need something the step before to really satisfy #2901

jonkeane: Yeah that's better than nothing. What about "successfully connected" instead of "authenticated"? Maybe include the URL (and the username is nice too imho)

jonkeane: Hmm, so _for cloud_ we probably want it to say someting like "Connect Cloud" when I mentioned the URL I was thinking about that for Connect, basically: the URL that the person had just entered. So that it's clear that's what they have connected to?

fizzboop: thank you for clarifying!

fizzboop:
@jonkeane:
After chatting with Jordan, we came up with this possible solution. Is this more aligned to your vision? We paired on trying to add a new line, but the `prompt` of the `InputBox` does not support this.


https://github.com/user-attachments/assets/0fc86a9f-5475-4613-be4f-4a097bd0e993

dotNomad:
> Is this more aligned to your vision?

Specifically this bit:

CleanShot 2025-09-03 at 12 40 08

It uses the `prompt` for entering a nickname, the step after authenticating.

CHANGELOG.md
```diff
@@ -7,11 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
```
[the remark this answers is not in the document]
fizzboop: i was surprised too, but then i saw [this](https://github.com/posit-dev/publisher/blob/main/CHANGELOG.md) on `main`.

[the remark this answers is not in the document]
fizzboop: i think the reason why it looks like it's an addition is because i relocated the newest `Added` bullet point under the oldest one and then added my bullet point underneath.

Issue #2845, opened 2025-08-19
Handling of missing .renv or manifests files on publish attempts


tyatposit:
## System details

Publisher Version        : publisher-1.18.1-41-g96e48fdb-darwin-arm64.vsix
Connect Version          : 2025.02.0
Your OS                  : MacOS 14.7.7 (23H723)
Content Type to Deploy   : rmarkdown - https://github.com/posit-hosted/examples-rmarkdown

## Publisher Output and Logs

missing renv.lock
`/Repos/publisher-examples/examples-rmarkdown/renv.lock: no such file or directory. See [Troubleshooting docs](https://github.com/posit-dev/publisher/blob/main/docs/troubleshooting.md) for help.`

## Problem Statement

It would be a better user experience to handle cases where the user does not have the appropriate files to be able to successfully publish their repo and potentially help them with the steps in the extension when common problems are hit.

eg. know that if they don't have an renv.lock file to run `renv::init()`

The docs show that this specific scenario used to be covered, but we don't see that same button/error anymore

<img width="469" height="307" alt="Image" src="https://github.com/user-attachments/assets/916e5248-c825-4183-91b5-3f29a3b59e77" />

## Steps to Reproduce

1. Deploy an R project repo that is missing the renv.lock file

dotNomad:
@mslynch just to clarify this issue is not something that is impacting customers right now (because I assume it hasn't been released), but will if we do release without addressing it?

dotNomad:
> Right, it's not affecting customers now. I mean that we should fix this one way or another before we do release, preferably through [#2838](https://github.com/posit-dev/publisher/issues/2838).

I labelled this as a `release-blocker` given the impact it has and that you identified it as an actual bug. Thanks for investigating.

PR #2953, opened 2025-08-27
bring back renv errors

mslynch:
## Intent

Resolves https://github.com/posit-dev/publisher/issues/2845.

Brings back the old "Project requires renv initialization "renv::init()" to be deployed" error message when your renv.lock file is missing and you try to deploy.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I moved the interpreter code back to where it was originally happening.

## User Impact

Users should see the proper error again.

## Automated Tests

Updated automated tests to include an renv.lock and make sure it gets read from.

## Directions for Reviewers

Verify this error appears when you try to deploy a project that's missing renv.lock:
```
Project requires renv initialization "renv::init()" to be deployed. See [Troubleshooting docs](https://github.com/posit-dev/publisher/blob/main/docs/troubleshooting.md) for help.
```

## Checklist

### DIFF
--- a/internal/publish/interpreters.go
+++ b/internal/publish/interpreters.go
@@ -8,8 +8,8 @@ import (
 	"github.com/posit-dev/publisher/internal/interpreters"
 )
 
-// readInterpreterDetails reads interpreter-specific dependency files and updates the Target deployment record for recordkeeping.
-func (p *defaultPublisher) readInterpreterDetails() error {
+// addInterpreterDetailsToTarget reads interpreter-specific dependency files and updates the Target deployment record for recordkeeping.
+func (p *defaultPublisher) addInterpreterDetailsToTarget() error {
 	if p.Config.Python != nil {
 		filename := p.Config.Python.PackageFile
 		if filename == "" {
--- a/internal/publish/publish.go
+++ b/internal/publish/publish.go
@@ -282,11 +282,6 @@ func (p *defaultPublisher) doPublish() error {
 	defer bundleFile.Close()
 	defer os.Remove(bundleFile.Name())
 
-	err = p.readInterpreterDetails()
-	if err != nil {
-		return err
-	}
-
 	err = p.serverPublisher.PublishToServer(contentID, bundleFile)
 	if err != nil {
 		return err
--- a/internal/publish/bundle.go
+++ b/internal/publish/bundle.go
@@ -18,7 +18,7 @@ func (p *defaultPublisher) createBundle(manifest *bundles.Manifest) (*os.File, e
 	p.emitter.Emit(events.New(op, events.StartPhase, events.NoError, createBundleStartData{}))
 	prepareLog.Info("Preparing files")
 
-	err := p.readInterpreterDetails()
+	err := p.addInterpreterDetailsToTarget()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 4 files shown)


rodriin (commented):
The changes look straight forward and I tested it for Connect Cloud and it all works as expected. However, it does not work as expected for Connect. We should try to make this work for Connect as well.

mslynch:
Pushed another commit to make things work with Connect too. This should be the same ordering of steps that we originally had.

dotNomad:
Noticed that this failed on linting. Noting that as I get started with the review 👀

internal/publish/publish.go
```diff
@@ -270,18 +265,28 @@ func (p *defaultPublisher) doPublish() error {
...
 
-	bundleFile, err := p.createBundle()
+	bundleFile, err := p.createBundle(manifest)
 	if err != nil {
 		return err
 	}
 	defer bundleFile.Close()
 	defer os.Remove(bundleFile.Name())
 
+	err = p.readInterpreterDetails()
```
dotNomad, line 285: I see that `p.readInterpreterDetails` is don't here and inside of `p.createBundle` above. Can we remove this if `p.createBundle` errors on this?

internal/publish/manifest.go
dotNomad, line 1: Really like that you split this out 💯

internal/publish/publish.go
```diff
@@ -253,11 +253,6 @@ func (p *defaultPublisher) doPublish() error {
 		return err
 	}
 
-	err = p.configureInterpreters()
-	if err != nil {
-		return err
-	}
```
dotNomad, line 259: Do we need to configure interpreters prior to creating the manifest? I see that `createManifest` calls `getRPackages` which uses the interpreters.

Further look into `readInterpreterDetails` it doesn't look like `getRPackages` depends on the changes to `defaultPublisher.Target` that it sets.

Just thinking through this out loud to make sure I'm not missing something.

internal/publish/interpreters.go
```diff
@@ -8,7 +8,8 @@ import (
 	"github.com/posit-dev/publisher/internal/interpreters"
 )
 
-func (p *defaultPublisher) configureInterpreters() error {
+// readInterpreterDetails reads interpreter-specific dependency files and updates the Target deployment record for recordkeeping.
+func (p *defaultPublisher) readInterpreterDetails() error {
```
dotNomad, line 12: The `read` rename here felt a bit confusing to me. Not a big deal, but I expected `read` to not set anything in `defaultPublisher`.

`configure` wasn't super clear, but made it more clear. The additional comment helps a lot though good add 👍

dotNomad (commented):
I had one question about the duplicate call to `readInterpreterDetails` that I think is worth discussing.

internal/publish/publish.go
mslynch: Yep - just removed it; we only want to call it inside createBundle.

internal/publish/interpreters.go
mslynch: Just renamed this to `addInterpreterDetailsToTarget`. Naming things is difficult 😔

dotNomad (approved):
Good choice on the new name that is much better 💯 

With the removal of the extra call this is looking perfect. I pulled it down and tested it - got the `renv` setup notification exactly as I expected.

Issue #2948, opened 2025-08-27
Help out on rendering Quarto documents if there are no existing renderings


marcosnav:
**Edit:**

The Quarto Extension wasn't really suited for our use case, instead we will have to call `quarto render` directly.

---

**Original:**

For Quarto deployments, when opting to "Publish rendered document only" and there are no renderings, call the Quarto Extension (if available) to pre-render the document so the resulting configuration is accurate with the static files to be deployed.

The available commands from the Quarto Extension that can help with this, depending on the project, are:
- `quarto.renderProject` (If there is a `_quarto.yml` present, we want to render the project)
- `quarto.renderDocument` (No `_quarto.yml`? render the doc passed as entrypoint)

marcosnav:
> We were thinking about trying something like this: Adding a button above that is "render your project" when it's a render -> static deploy.

The latest helper that landed today https://github.com/posit-dev/publisher/pull/2965 does most of the heavy-lifting for this.

I can start experimenting with the button tomorrow.

> One thing that we will need to add is tracking the source file (notice the pink/red circle in the config) so that we know what to call.

- What do you mean by source file? and by "know what to call"?
- The new helper tracks if it is a complete Quarto project or just a standalone doc and handles both.

dotNomad:
If someone has a Quarto project and they were on the `_quarto.yml` for example that would be the source file - so we then know that this `html` type Configuration needs to render with Quarto.

If you are an individual file like `index.qmd` then that would be the source file.

marcosnav:
@jonkeane sounds good, thank you for clarifying!

PR #2965, opened 2025-08-28
feat: Quarto helper to handle rendering Quarto projects

marcosnav:
## Intent

Introducing a new `class QuartoProjectHelper` to help out on:
- Identifying if a Quarto project has existing renderings
- Check if there is Quarto available on the system
- Attempt to render the project
- Fallback to attempt to render a single document when it isn't a project

Not in use at the moment, it is a building block that then will be pulled to the multistepper.

Related to #2948 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

New `class QuartoPorjectHelper` that does multiple checks and calls `quarto` bin as needed.

## User Impact

None at the moment.

## Automated Tests

New unit tests for this helper.

## Directions for Reviewers

These changes cannot be tried manually. 

## Checklist


extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -619,7 +620,19 @@ export async function newDeployment(
...
+        api.files,
+        entrypoint,
+        newDeploymentData.entrypoint.inspectionResult!.configuration.entrypoint!,
+        newDeploymentData.entrypoint.inspectionResult!.projectDir,
+      );
+      try {
+        await quartoHelper.verifyRenderedOutput();
+      } catch (err) {
+        console.log(err);
+      }
```
rodriin, line 635: I thought you said this wasn't in use? Looks in use to me, unless this condition has zero chance of being `true`: `if (pick.label === useRenderedDoc)`.

.vscode/launch.json
rodriin, line 1: Did you mean to remove this file? I thought we needed it to run locally against the local agent.

extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -619,7 +620,19 @@ export async function newDeployment(
 
     // If wants to push rendered code, update the inspectionResult to use that config
     if (pick.label === useRenderedDoc) {
+      const entrypoint = newDeploymentData.entrypoint.filePath || "";
       useAlternativeConfig();
+      const quartoHelper = new QuartoProjectHelper(
+        api.files,
+        entrypoint,
+        newDeploymentData.entrypoint.inspectionResult!.configuration.entrypoint!,
+        newDeploymentData.entrypoint.inspectionResult!.projectDir,
```
rodriin, line 629: I'm not loving the use of the guaranteed `!` operator here. Is it possible to refactor the code so we don't have to do this?

extensions/vscode/src/utils/quartoProjectHelper.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    // No renderings, we'll help out rendering, if possible
+    try {
+      await this.isQuartoAvailable();
+    } catch (_) {
+      // Quarto is not available on the system,
+      // let the user continue, nothing we can do
+      return;
+    }
```
rodriin, line 47: I think all of these `catch` statements can be:
```
} catch {
  // ...
}
```
I don't think you need the: `(_)` necessarily.

extensions/vscode/src/utils/quartoProjectHelper.ts
```diff
@@ -0,0 +1,118 @@
...
+    try {
+      await this.renderDocument();
+    } catch (_) {
+      return Promise.reject();
+    }
+  }
+
+  isQuartoAvailable(): Promise<void> {
+    return runTerminalCommand("quarto --version");
+  }
```
rodriin, line 69: How will we know if the quarto version is actually available if the return type here is `Promise<void>`? Wouldn't `Promise<boolean>` be better or even `Promise<string>` for that matter?

extensions/vscode/src/multiStepInputs/newDeployment.ts
marcosnav: gah sorry, I pushed some in test work ignore this changes for now

.vscode/launch.json
```diff
@@ -1,17 +0,0 @@
-{
-  // Use IntelliSense to learn about possible attributes.
```
marcosnav, line 2: gah sorry, I pushed some in test work ignore this changes for now

extensions/vscode/src/utils/quartoProjectHelper.ts
marcosnav: Because it is a terminal command that runs, the promise rejects if the command fails, so if Quarto is available is a `resolve` otherwise is a `reject`, there is no need for a value within the promise.

Although I see that it might be confusing and handling only the promise value might be better 🤔 , I'll wrap the terminal resolution to have that.

rodriin: Oh ok I understand now.

PR #2976, opened 2025-09-02
feat: Re-arrange inputConfigAlternatives deployment step so the chosen inspection result check is not hardcoded

marcosnav:
## Intent

Re-arrange `inputConfigAlternatives` deployment step so the chosen inspection result check is not hardcoded

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

Check the `inspectResult` configuration type and do not rely on label strings.

## Directions for Reviewers

Choosing to publish the rendered version of a Quarto doc should still work.

## Checklist

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -146,6 +146,16 @@ export async function newDeployment(
     return Boolean(alternatives && alternatives.length);
   };
 
+  const inspectionResultAlternative = (
+    insRes: ConfigurationInspectionResult,
+  ): ConfigurationInspectionResult => {
+    // At the moment the only alternative we handle is the rendered version alternative
+    return {
+      projectDir: insRes!.projectDir,
+      configuration: insRes!.configuration!.alternatives![0],
+    };
+  };
+
   const useAlternativeConfig = () => {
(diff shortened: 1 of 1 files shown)


rodriin (approved):
I like the cleanup, very nice. I have a general reservation against the use of the guaranteed `!` operator. It usually indicates that further refactoring could be done to either not make things nullable/undefined or maybe switch it up to use the optional `?` operator instead?

marcosnav:
> I like the cleanup, very nice. I have a general reservation against the use of the guaranteed `!` operator. It usually indicates that further refactoring could be done to either not make things nullable/undefined or maybe switch it up to use the optional `?` operator instead?

I agree, but hesitant to change it. I do not like the need for `!` but at this stage of the code we are 100% sure these properties are present but we cannot change the object type notation. I'll spend a couple mins to see if I can add a quick change to this.

dotNomad:
> I like the cleanup, very nice. I have a general reservation against the use of the guaranteed `!` operator. It usually indicates that further refactoring could be done to either not make things nullable/undefined or maybe switch it up to use the optional `?` operator instead?

I'm in agreement. I took a quick look, and there wasn't a really easy to way to accomplish a full removal.

PR #2983, opened 2025-09-03
feat: Prevent users to have unknown content configurations for Quarto projects

### LINKED ISSUE #2890: Quarto Deployment Determines an `unknown` Type
## System details

    Connect Cloud Account Name   : chip (has since been deleted)
    Connect Cloud Account Type   : individual
    Connect Cloud Deployment URL :
    Content Type to Deploy       : Quarto

## Publisher Output and Logs
Output is blank. 🤷 

## Problem Statement
When generating a deployment for https://github.com/posit-hosted/examples-quarto-r, the `type` in the toml file was determined to be `unknown`. I needed to update it to `quarto-static`.

The current theory is because I do not have Quarto installed, that the proper `type` could not be determined when I choose to deploy the `.qmd` file.

## Steps to Reproduce
- Open Positron (without Quarto installed on your machine)
- Clone https://github.com/posit-hosted/examples-quarto-r
- Deploy `report.qmd` to Connect Cloud
- The `.toml` file that is generated will have a `type` of `unknown`

marcosnav:
## Intent

Resolves #2890 

Getting a configuration with `type= "unknown"` is painful. Users that do not have a Quarto binary on the system but that the project can be easily identified as a Quarto project by `.qmd` files are landing in this state.

This PR aims to fix that.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

If there is no Quarto binary on the system, pay attention to the files so that for `.qmd` files or if there is a `_quarto.yml` present we determine the project is a Quarto project.

The resulting configuration will add the adjacent `.qmd` files, we cannot do more than that since `quarto inspect` won;t work. Even so, it is a much better state that can be fixed with the files picker, rather than having `type= "unknown"`.

## User Impact

Users with no Quarto binary on the system can get Quarto type configurations.

## Automated Tests

Expanded unit tests to cover the described scenario.

## Directions for Reviewers

This one is a bit cumbersome to verify manually because quarto should not be available on the system. Options are:
- Get rid (hide, rename, move temporarily from `/usr/local/bin`) of your quarto installation.
- Modify the code, on `/inspect/detectors/quarto.go L51` change `RunCommand("quarto"` to not be `"quarto"` and use an unavailable command name. Then package and install.

Try publishing a Quarto project, even if there is no quarto binary available, you should be able to get a Quarto type configuration.

## Checklist

### DIFF
--- a/internal/inspect/detectors/quarto.go
+++ b/internal/inspect/detectors/quarto.go
@@ -28,6 +28,10 @@ var specialYmlFiles = []string{
 	"_brand.yaml",
 }
 
+// When there is no quarto binary or version detected,
+// this is the default we will use on deployment configurations.
+var defaultQuartoVersion = "1.7.34"
+
 type QuartoDetector struct {
 	inferenceHelper
 	executor executor.Executor
@@ -153,8 +157,18 @@ func (d *QuartoDetector) configFromFileInspect(base util.AbsolutePath, entrypoin
 	inspectOutput, err := d.quartoInspect(entrypointPath)
 	if err != nil {
 		// Maybe this isn't really a quarto project, or maybe the user doesn't have quarto.
-		// We log this error and continue checking the other files.
 		d.log.Warn("quarto inspect failed", "file", entrypointPath.String(), "error", err)
+
+		// If the inspection fails, but we have a quarto.yml or .qmd file, we can still generate a basic config.
+		d.log.Warn("attempting to identify Quarto project by files context", "file", entrypointPath.String())
+
+		config, err := d.genNonInspectConfig(base, entrypointPath)
+		if err == nil && config != nil {
+			d.log.Info("Generated Quarto configuration without Quarto binary inspection", "file", entrypointPath.String())
+			return config, nil
+		}
+
+		d.log.Warn("Could not identify Quarto project by files context", "file", entrypointPath.String())
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


internal/inspect/detectors/quarto.go
```diff
@@ -28,6 +28,10 @@ var specialYmlFiles = []string{
 	"_brand.yaml",
 }
 
+// When there is no quarto binary or version detected,
+// this is the default we will use on deployment configurations.
+var defaultQuartoVersion = "1.7.34"
```
marcosnav, line 33: @jonkeane @dotNomad Thoughts on this? 

Since there is no Quarto binary to check the version, I'm adding a fallback because the deployment configuration is not valid without a version and we also need it to generate the manifest.

dotNomad: My understanding of this is the `[quarto]` section requires a version unlike the `[python]` and `[r]` sections can use a selected/installed interpreter. With that being the case I'm fine with this hard code for now as well.

dotNomad (approved):
> Modify the code, on /inspect/detectors/quarto.go L51 change RunCommand("quarto" to not be "quarto" and use an unavailable command name. Then package and install.

Thank you for adding this in the "Directions for Reviewers" this was very helpful to get going quickly on this 💯 

---

This is such a _drastically_ better experience for those without Quarto installed. Well done!

PR #3000, opened 2025-09-03
feat: machinery to set up renv in a directory

toph-allen:
## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist

### DIFF
--- a/internal/inspect/dependencies/renv/project_dependencies.go
+++ b/internal/inspect/dependencies/renv/project_dependencies.go
@@ -74,8 +74,8 @@ func (s *defaultRDependencyScanner) ScanDependenciesInDir(paths []string, rExecu
 
 	// Run the script (use the temporary project as the working directory)
 	// Ensure working directory is provided as an AbsolutePath
-	absTmp := util.NewAbsolutePath(targetPath.String(), nil)
-	stdout, stderr, err := s.rExecutor.RunScript(rExecutable, []string{"-s"}, script, absTmp, s.log)
+	absTarget := util.NewAbsolutePath(targetPath.String(), nil)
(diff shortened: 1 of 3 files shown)


internal/services/api/post_packages_r_scan.go
```diff
@@ -52,24 +57,18 @@ func (h *PostPackagesRScanHandler) ServeHTTP(w http.ResponseWriter, req *http.Re
...
-	rPath := util.NewPath(b.R, nil)
-
-	rInterpreter, err := h.rInterpreterFactory(projectDir, rPath, h.log, nil, nil, nil)
-	if err != nil {
-		InternalError(w, req, h.log, err)
-		return
-	}
-	err = rInterpreter.CreateLockfile(lockfileAbsPath)
+	dependencyScanner := h.rDependencyScannerFactory(h.log)
+	_, err = dependencyScanner.SetupRenvInDir(projectDir.String(), lockfileRelPath.String(), b.R)
```
amol-, line 67: This seems to be necessary only for tests, but adding complexity to the implementation to make tests easier does seem something that will hit back on long term.

The `PostPackagesRScanHandler` could have the scanner as an attribute instead of the factory, the scanner is primarily stateless, it only relies on the Executor and the Executor has no state. So we can just pass a scanner to `NewPostPackagesRScanHandler` and avoid the factory.

To be honest, I think I'd make the scanner an internal detail of `PostPackagesRScanHandler` and have the `NewPostPackagesRScanHandler` create one by itself. At that point the tests could just change the attribute of the resulting post handler instead of passing arguments to the constructor.

amol- (approved):
LGTM

PR #3001, opened 2025-09-04
feat: Allow user to explicitly pick python package manager or server default

amol-:
## Intent

At the moment, if package_manager is not specified, the publisher will default to enforcing "pip".
Instead we want:
1. If not specified -> server default
2. if pip -> enforce pip
3. If uv -> enforce uv

## Type of Change

- - [x] New Feature 

## Approach

If `python.package_manager` option is not specified, it will now default to an `"auto"` value.
This value gets then translated to `"pip"`+`"allow_uv=null"` when generating the manifest.
That means the server will be able to choose what to do based on its `DefaultPackageManager` option.

The `allow_uv=null` is because older servers didn't understand or accept the
`allow_uv` option. By setting it at null it gets omitted from the `manifest.json`
and thus retains backward compatibility.

If instead the user specified a value, that value is used and the server can accept
or refuse the deploy based on its `AllowUv` server side option.

## User Impact

When deploying against a server that defaults to `uv`, the user can now
switch back to `pip` by specifying `package_manager=pip`, previously
doing so, would have yet deployed with `uv` if that was the default server
package manager.

## Automated Tests

A test confirms that the json generated given the 3 cases is the expected one.

## Directions for Reviewers

`TestNewManifestFromConfig_PythonPackageManagerVariants_JSON`  more or less explains the expectations

## Checklist

### DIFF
--- a/internal/bundles/manifest.go
+++ b/internal/bundles/manifest.go
@@ -86,8 +86,9 @@ type Jupyter struct {
 
 type PythonPackageManager struct {
 	Name        string `json:"name"`
-	Version     string `json:"version,omitempty"` // Package manager version
-	PackageFile string `json:"package_file"`      // Filename listing dependencies; usually "requirements.txt"
+	Version     string `json:"version,omitempty"`  // Package manager version
+	PackageFile string `json:"package_file"`       // Filename listing dependencies; usually "requirements.txt"
+	AllowUv     *bool  `json:"allow_uv,omitempty"` // Whether the package manager can be "uv"
 }
 
 type PackageMap map[string]Package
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -27,8 +27,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - R manifest package generation now uses only `renv.lock` file by default. To use the
-  local `renv` library instead, set `[r].packages_from_library = true` in the
+  local `renv` library instead, set `r.packages_from_library = true` in the
```
amol-, line 30: Unrelated, but it was requested in a previous PR to change this and it didn't value a dedicated PR.

marcosnav (approved):
Nice and clean work as always 🚀

PR #3002, opened 2025-09-04
wip nightly prerelease workflow

dotNomad:
Work in Progress workflow.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist

### DIFF
--- a/.github/workflows/nightly-prerelease.yaml
+++ b/.github/workflows/nightly-prerelease.yaml
@@ -1,0 +1,179 @@
+name: Nightly Pre-release
+on:
+  # TODO: Enable schedule after testing
+  # schedule:
+  #   # Run at 7 AM UTC daily
+  #   - cron: "0 07 * * *"
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Dry run"
+        required: false
+        default: true
+        type: boolean
+
+jobs:
+  check:
+    name: Check if release needed
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-prerelease.yaml
```diff
@@ -0,0 +1,179 @@
...
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          ref: main
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "lts/*"
+
+      - name: Calculate version
```
dotNomad, line 69: This step is wild, and a first pass using Claude. I'm sure it can be greatly, greatly simplified.

.github/workflows/nightly-prerelease.yaml
```diff
@@ -0,0 +1,179 @@
...
+  # schedule:
+  #   # Run at 7 AM UTC daily
+  #   - cron: "0 07 * * *"
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Dry run"
+        required: false
+        default: true
+        type: boolean
```
dotNomad, line 13: This is currently a dry-run only workflow because we will have to setup a GitHub app to actually make the tag rather than using a `GITHUB_TOKEN`

> If you do want to trigger a workflow from within a workflow run, you can use a GitHub App installation access token or a personal access token instead of GITHUB_TOKEN to trigger events that require a token.

PR #3014, opened 2025-09-04
Clean up contributing docs

jonkeane:
Remove some references to internal tools.


tyatposit:
Thanks.  fwiw, we also reference our aws console in the schema section a little further down.

PR #3016, opened 2025-09-04
Integration requests

joshyam-k:
DRAFT

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist


extensions/vscode/src/constants.ts
```diff
@@ -88,6 +88,8 @@ const homeViewCommands = {
   AddCredential: "posit.publisher.homeView.addCredential",
   DeleteCredential: "posit.publisher.homeView.deleteCredential",
   RefreshCredentials: "posit.publisher.homeView.refreshCredentials",
+  AddIntegrationRequest: "posit.publisher.homeView.addIntegrationRequest",
+  DeleteIntegrationRequest: "posit.publisher.homeView.deleteIntegrationRequest",
```
mconflitti-pbc, line 92: question for @dotNomad , when does it make sense to add commands to the command palette?

extensions/vscode/webviews/homeView/src/components/views/Integrations.vue
mconflitti-pbc, line 1: Let's change this to IntegrationRequests.vue

extensions/vscode/webviews/homeView/src/stores/home.ts
```diff
@@ -42,6 +45,26 @@ export const useHomeStore = defineStore("home", () => {
...
+  //     return config.configuration.integrationRequests || [];
+  //   },
+  //   set: (requests: IntegrationRequest[]) => {
+  //     const config = selectedConfiguration.value;
+
+  //     if (config !== undefined && !isConfigurationError(config)) {
+  //       config.configuration.integrationRequests = requests;
+  //     }
+  //   }
+  // });
```
mconflitti-pbc, line 66: following up with @dotNomad and @marcosnav : assuming we get the watcher stuff wired up properly, computed() actually causes problems here since this makes the integrationRequests value depend upon the selectedconfiguration. When state is refreshed and new data is added to the homestore, there is a time where selectedConfiguration is undefined while it is waiting to be set which was causing the blip in the UI. By making this is a ref value instead we can set it and let it maintain its value until we want it to change.

test/e2e/tests/integration-requests.cy.js
```diff
@@ -0,0 +1,158 @@
...
+    cy.resetConnect();
+    cy.resetCredentials();
+    cy.visit("/");
+  });
+
+  afterEach(() => {
+    cy.clearupDeployments("static");
+  });
+
+  it("Connect Cloud should not display IntegrationRequests", () => {
```
mconflitti-pbc, line 12: is it possible for us to run this same test for a test connect instance?

internal/schema/schemas/posit-publishing-schema-v3.json
```diff
@@ -469,6 +469,40 @@
       },
       "description": "Names of secrets required by the application. Injected as environment variables.",
       "examples": ["API_KEY", "DATABASE_PASSWORD"]
+    },
+    "integration_requests": {
```
mconflitti-pbc, line 473: @dotNomad we added this here but do we need this?

marcosnav (commented):
Looks pretty good in general! A few questions and comments

internal/clients/connect/client_connect.go
```diff
@@ -562,3 +562,21 @@ func (c *ConnectClient) GetCurrentUser(log logging.Logger) (*User, error) {
...
+}
+
+func (c *ConnectClient) GetServerSettings(log logging.Logger) (*ServerSettings, error) {
+	var settings ServerSettings
+	err := c.client.Get("/__api__/server_settings", &settings, log)
+	if err != nil {
+		return nil, err
+	}
+	return &settings, nil
+}
```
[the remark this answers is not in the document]
mconflitti-pbc: doh! must have missed it. Will flip this to public, add the fields we need to use and reuse this method. Thanks!

joshyam-k:
@marcosnav thanks for the helpful feedback! I had to make the `getSettings` method and some of it's fields public which required changing a number of tests (hence the large-ish diff), but everything seems to be working smoothly.

extensions/vscode/src/constants.ts
dotNomad: Apologies for the late response here, but I think it makes sense anytime the command could be run without arguments or specific state. For example `DeleteIntegrationRequest` requires knowing which integration request to remove so that doesn't belong in the command palette in my mind.

Issue #2996, opened 2025-09-03
feat: Open connect logs in file viewer


jonkeane:
## Description

The logs viewer is really great in that it curates logs, but it would be nice to be able to open the text of logs in the file viewer itself. This would make searching, copying, pasting, etc. much easier *and* it would let someone use this as input in positron assistant too.

Came up in user testing with @blairj09

PR #3017, opened 2025-09-05
Implementation of Publisher raw logs toggle

### LINKED ISSUE #2875: There is no ability to right click or easily copy the contents of the Publisher Tab Logs
## Steps to Reproduce

1. Attempt a publish
2. Click the Publisher Tab logs
3. Expand one of the sections
4. Try to right click
5. Try to copy the actual contents of the logs

actual:  No way to easily copy the details
expected:  Should be able to copy details like error ids, etc

rodriin:
## Intent

Resolves #2875 

The intent is to introduce a toggle view on the Publisher logs where the user can easily toggle between log summary view (tree view) and the log raw view (text view) so they can easily copy the logs for more in-depth investigations. The are also a couple of handy buttons "Copy Logs" and "Save & Open Logs", which respectively either copies all the logs to the clipboard or saves all the logs to a file and opens it. The "Save & Open Logs" button, which generates the log file and opens it in the editor will provide the capability for searching and sharing that file with other extensions such as the Positron Assistant, which can be used to help analyze errors and provide potential solutions.

<details>
  <summary>VS Code screen video</summary>

https://github.com/user-attachments/assets/cde32f23-e25d-484f-b866-f846f0c85a14
</details> 

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

The approach was to use a container panel and have a couple of views that belong to that panel but only one view is visible at any given point using a view mode that gets persisted in the context global state. There are also a couple of commands for the toggling. The reason for this is to be able to change the icon when we toggle so the list icon changes to the tree icon and so on. The other thing to notice is since we have a couple of views now for the logs panel and any of them can be hidden or visible at any given time, we have to then focus on both views whenever we want to focus on the logs panel view. Whichever view is visible at the moment will work as expected and it will be a no-op for whichever view is currently hidden. This could have been done is a more complex way where we figure out which view is currently visible and only focus on that one view but that felt like unnecessary extra work.

## User Impact

User will now be able to toggle between the summary logs view (tree) and the raw logs view (text). They will also be able to copy and save the raw logs.

## Automated Tests

My branch passes E2E tests in CI

## Directions for Reviewers

- [ ] Verify that you are able to toggle between the summary logs view (tree) and the raw logs view (text) for both Connect and Connect Cloud deployments.
- [ ] Verify that you are able to copy and save the raw logs for both Connect and Connect Cloud deployments in either the tree view or the plain view of the logs panel.

## Checklist


mslynch (approved):
very cool!

jonkeane:
I'm curious if the raw view is available to positron assistant? That was one of the big drivers for folks asking for #2996

rodriin:
@jonkeane Great question!
> I'm curious if the raw view is available to positron assistant? That was one of the big drivers for folks asking for https://github.com/posit-dev/publisher/issues/2996

I'm not sure but I can check that and let you know. Maybe I can even throw in a button in there to open it in an editor tab? I'll let you know what's possible.

extensions/vscode/src/views/logs.ts
```diff
@@ -88,6 +98,283 @@ const createLogStage = (
 
 const RestoringEnvironmentLabel = "Restoring Environment";
 
+const stages = new Map([
```
rodriin, line 101: This was refactored out of `LogsTreeDataProvider` to avoid code duplication.

jonkeane:
> The reason I didn't keep it that way was because we would need a new place to launch it from and I could not think of where to put that initially. Looking at it now, the argument could be made that the toggle button could launch it into the editor instead of toggling the view. 
 
 Yeah, I like this idea — that location for "open this in the file viewer" is nice.
 
> However, you would loose the ability to see the logs moving live while the publishing is underway because it would write to the file whatever has been logged until that exact point and time when the user clicks the button. 

_nods_, I'm not actually sure that folks want to see the logs streaming in this way — for most people they are going to let it run and then look at the logs at the end. That being the case: IMHO not seeing them moving isn't a big deal. But also, I'm sure there's someway in the vscode API/UI for us to update the text of the logs in the file viewer (I know it's conceptually very different, but it does this when it's editing code in LLM agent mode!). So maybe there's a way we can actually even have that happen.

> Therefore, I think it would be more intuitive to keep it the way it is implemented now, but I'm open to other suggestions/ideas.

IMO, the way it is now feels pretty clunky. If someone were to want to interact with the logs via positron assistant they would need to know to click the expanded log view and then also click the save button before that becomes available. Neither are things that folks are likely to intuit without being shown / told how to do. And having them in two different places will make folks wonder: why are these both here? What is the difference? etc.

PR #3030, opened 2025-09-08
fix: Avoid trying to include n bundles the default renv.lock when it doesn't exist.

amol-:
## Intent

When the `renv.lock` file doesn't exist the publisher sets it by default anyway and tries to include it in the bundle,
this fails as there is no renv.lock to include. As in that case we scan dependencies and generate the manifest without
any renv.lock, we should just avoid including in the bundle any lockfile.

## Type of Change

- - [x] Bug Fix 

## Approach

When the lockfile isn't found, keep `Config.R.PackageFile` empty to signal that there is no lockfile at all.
Previously it was set to the default value in all cases.
Then the code can use `PackageFile == ""` as a guard to know if the renv.lock must be included in the bundle or not.

## User Impact

Deploying without any lockfile should work

## Automated Tests

Added tests to confirm that the lock is included when it exists and that is not used when it does not exist.

## Checklist

Does it need an update?

### DIFF
--- a/internal/publish/interpreters.go
+++ b/internal/publish/interpreters.go
@@ -32,20 +32,20 @@ func (p *defaultPublisher) addInterpreterDetailsToTarget() error {
 		p.Target.Requirements = requirements
 	}
 
-	if p.Config.R != nil && p.Config.R.PackageFile != "" {
-		filename := p.Config.R.PackageFile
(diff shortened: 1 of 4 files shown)


jonkeane:
I also checked with an `renv.lock` file too and that works well as well! ❤️

PR #3031, opened 2025-09-08
chore(ci): Create initial nightly-prerelease.yaml

cgraham-rs:
## Intent

Create initial workflow and supporting actions to trigger a prerelease.

Fixes: #3032

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

- Leverage https://github.com/posit-dev/publisher/pull/3002 as a springboard
- Tried to include plenty of code comments and logging so humans can review and debug easily
- Leveraged native semver package functionality as much as possible

## User Impact

None, other than the potential for more frequently nightly prerelease builds in the future

## Automated Tests

None, we will have to test this in dry_run mode manually after merging

## Directions for Reviewers

- Review the code as usual
- Double-check that there's nothing that will be creating and pushing tags to the repo right now so we can safely test the workflow using dry_run once merged

## Checklist

### DIFF
--- a/.github/workflows/nightly-prerelease.yaml
+++ b/.github/workflows/nightly-prerelease.yaml
@@ -138,14 +138,21 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          # Use a PAT to ensure the tag push triggers the release workflow
-          # GITHUB_TOKEN doesn't trigger other workflows by default
-          token: ${{ secrets.RELEASE_PAT || secrets.GITHUB_TOKEN }}
           ref: main
+
+      - name: Generate token for GitHub App
+        id: generate-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.POSIT_CONNECT_PROJECTS_APP_ID }}
+          private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
+          # Tags are part of repository contents, so we need this permission
+          permission-contents: write
 
       - name: Create tag
         env:
           NEXT_VERSION: ${{ needs.version.outputs.next-version }}
+          GH_TOKEN: ${{ steps.generate-token.outputs.token }}
         run: |
           # Configure git
           git config user.name "github-actions[bot]"
@@ -154,8 +161,8 @@ jobs:
           # Create the tag
           git tag -a "$NEXT_VERSION" -m "Nightly pre-release $NEXT_VERSION"
 
-          # Push the tag (this will trigger the release workflow)
-          git push origin "$NEXT_VERSION"
+          # Use the token for authentication when pushing
+          git push "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}" "$NEXT_VERSION"
 
           echo "✅ Created and pushed tag: $NEXT_VERSION"
 


cgraham-rs:
> Looks great in general, really happy to see this work.
> 
> Just one thing while trying the tests, there is one case that is a bit confusing and I'm not sure if it is accurate, want to confirm first.
> 
> ```
> === Running test: Real world tags for next patch version ===
> Inputs:
>   release-type: patch
>   all-tags: v1.19.1,v1.19.0,v1.18.1,v1.18.0,v1.16.1,v1.16.0,v1.14.0,v1.12.1,v1.12.0,v1.10.0
> 
> Results:
>   Outputs:
>     next-version: v1.20.0
>     latest-version: v1.19.1
>     current-release: v1.18.1
>     current-prerelease: v1.19.1
>   ✅ All outputs match expected values
> ```
> 
> This test is bumping a `patch`, the last release in here is `v1.18.1` I would expect it to be `v1.18.2` but it bumps the `minor` to `v1.20.0`. Is that how we want it?, I think it makes sense with the mental model that if there is a pre-release and we want to include everything so we bump the minor to be a well cut release.
> 
> But at the same time I'm thinking that we do not always want that? E.g: A release `1.18.0` just went out, a week passes and there is a new feature in the works worth a pre-release so there you have `1.19.0`. Then, a couple bugs are fixed a week forward, the feature being worked on `1.19.0` is not ready for the prime time, we want to release `1.18.1` with small bug fixes that do not deserve a major nor even a minor bump, so when trying that it'll jump to `1.20.0`? Am I right?

This took some thinking to make sure I fully understood it. Please double check all of this.

The trick is the latest version is actually `latest-version: v1.19.1` which is an odd-minor version, thus a _prerelease_ version. But we called for `patch` which is asking for a _release_ patch which **must** use an _even-minor_ version. Thus the next logical version is `next-version: v1.20.0` I think this is functionally correct for honoring the VS Code extension versioning scheme.

Here's a theoretical real world scenario. 
- We're producing nightly prerelease builds 
- Let's say we've produced 5 bug-fix-only prereleases thus far
- We then decide to release a production bug-fix-only `patch` release
- Since that has to be an even-minor version number we have no choice but to increment the minor version number even though we're only intending to release a patch release
- As I write this I'm not thinking of an alternative. It doesn't make any sense to me that we'd go backwards in minor version numbers from `latest-version: v1.19.1` to `next-version: v1.18.2` because the last production release was `current-release: v1.18.1` Production users would get an upgrade as they would have had v1.18.1` installed. But this would likely cause prerelease users some potential problems

.github/actions/calculate-next-version/README.md
```diff
@@ -0,0 +1,105 @@
+# Calculate Next Version
```
jonkeane, line 1: I'm attaching this here but, it's really about the whole action here: this seems like a lot of code for bumping the minor or patch version. I suspect we've abstracted a lot and tried to handle a bunch of extra cases we don't need to (i.e. I suspect the completeness of this is YAGNI compared to the relatively smaller chunk of code doing the same thing in the branch: https://github.com/posit-dev/publisher/pull/3002/files#diff-443ddd6d357ddc79538a1cbaf01097a05bd139a4e7468c28f21f972cd091605cR72-R123)

Can we cut this down to the bare minimum we need to get automatic pre-releases out today and then we can work on a full version later if that is helpful for us? (which might be never, and that's ok!)

marcosnav (approved):
All clear with today's discussion! 🙌 This is great!

PR #3037, opened 2025-09-09
stage lockfile

amol-:
## Intent

Alternative to https://github.com/posit-dev/publisher/pull/3030 that always ensures a lockfile is in the generated bundle.
This would allow more robust bundles to future evolutions where Connect Server uses the lockfile directly.

## Type of Change

- - [x] Bug Fix 

## Checklist


jonkeane:
Ok, I've tested this in Positron on Workbench + locally and it's wroking well: both generating the ephemeral `renv.lock` files and reading from those when they are available. 🎉 Thank you for this!

PR #3048, opened 2025-09-10
feat: add "set up renv" button

### LINKED ISSUE #2986: feat: add button to set up renv in project directory
Add a button to the R packages list that invokes the method added in #2928 to set up an renv environment in a project directory.

toph-allen:
## Intent

Add a button, "Set Up renv", that sets up an renv environment, installs dependencies, and creates an renv lockfile. Exposes functionality added in #3000.

<details>
  <summary>Screen recording of the button in action</summary>

  [video](https://github.com/user-attachments/assets/32423526-9922-4999-aecd-cf69ba0a5dfc)
</details>

Resolves #2986

## Type of Change

- - [x] New Feature 

## Approach

- Adds a button to the R packages section, which is hidden when an R lockfile is detected.
- The button is disabled when clicked. The button will not re-enable till the extension restarts, to avoid problems like a race condition in the case of lengthy installed.

## User Impact

- Allows users with an R project without `renv` to set up renv in that directory.

## Automated Tests

Back-end tests were added in #3000. Adding the button is a front-end component-level change, and those aren't tested right now. I don't see a place to hook in for e2e tests -- perhaps in `test/e2e/tests/deployments.cy.js`, where tests are currently skipped.

## Directions for Reviewers

- Front-end changes are unremarkable.
- The back end code uses `InterpretersFromRequest` to get the R interpreter path. I don't know enough about how this works to feel _certain_ that it will always pick the correct R interpreter for a user's project. It's used like this elsewhere, but those cases might already have an `renv` environment to specify the R version, and I'm not sure how it falls back.

## Checklist


jonkeane:
> As currently implemented, this button shows when there is no renv.lock, not when there is no renv infra at all. This means it could technically overwrite and renv project if when there's no lockfile present (though this is probably an edge case). It feels as though there is a loose mapping from the extension's representation of the lockfile to renv's structure, and this probably isn't something to worry about at this point — I do think that this implementation is "good enough" in that regard.

Agreed this is good enough for now. There might be a circumstance where we want to be smarter here, but let's have someone ask us for that.

marcosnav (approved):
Tried this out and it is working nicely! Code looks pretty good too!

PR #3075, opened 2025-09-12
refactor(ci): Move generate-licenses workflow job to a step

cgraham-rs:
## Intent

Refactor license workflow validate-inputs job to a step. This will cleanup the job list so only one job will be reported like our other workflows.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

Move the job to a step so the entire workflow is reported under a single job instead of two like it currently is.

Before
image

After
image

## Checklist

### DIFF
--- a/.github/workflows/generate-licenses.yaml
+++ b/.github/workflows/generate-licenses.yaml
@@ -30,8 +30,10 @@ permissions:
   contents: write
 
 jobs:
-  validate-inputs:
+  license-check:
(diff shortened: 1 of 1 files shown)


marcosnav (approved):
Good call 👍

PR #3090, opened 2025-09-15
Fix nightly-prelease "Create tag" job

cgraham-rs:
## Intent

Attempt to fix the nightly-prelease create tag creation. The tag is being correctly created, but it is not triggering release.yaml which suggests some issue creating the tag.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

- Move `generate-token` as the first step so theoreatically all subsequent steps will use that token
- Remove unnecessary `generate-token` permission that our token definitely does not have (and should not need)
- Simpify the tag creation letting it automatically use the token created during `generate-token`

We will have to test post-merge using a manual dispatch.

## Checklist

### DIFF
--- a/.github/workflows/nightly-prerelease.yaml
+++ b/.github/workflows/nightly-prerelease.yaml
@@ -132,8 +132,6 @@ jobs:
           private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
           # Tags are part of repository contents, so we need this permission
           permission-contents: write
-          # Add workflow permission to allow triggering other workflows
-          permission-workflows: write
 
       - uses: actions/checkout@v5
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-prerelease.yaml
```diff
@@ -137,23 +132,26 @@ jobs:
           private-key: ${{ secrets.POSIT_CONNECT_PROJECTS_PEM }}
           # Tags are part of repository contents, so we need this permission
           permission-contents: write
-          # Add workflow permission to allow triggering other workflows
-          permission-workflows: write
+
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          ref: main
```
nealrichardson, line 139: If you're trying to use the App token, do you need to pass it in here?

cgraham-rs: Probably. 

From [the documentation](https://github.com/actions/checkout?tab=readme-ov-file#usage).

```
    # Personal access token (PAT) used to fetch the repository. The PAT is configured
    # with the local git config, which enables your scripts to run authenticated git
    # commands. The post-job step removes the PAT.
    #
    # We recommend using a service account with the least permissions necessary. Also
    # when generating a new PAT, select the least scopes necessary.
    #
    # [Learn more about creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
    #
    # Default: ${{ github.token }}
    token: ''
```

cgraham-rs: Done.

PR #3095, opened 2025-09-15
remove alert from R Packges header

### LINKED ISSUE #3088: bug: When I don't have any renv infrastructure, I shouldn't be given a warning symbol
This is a follow on / was missed in #3029. The warning triangle on the R Packages section shouldn't be there (now that it's no longer an issue that there is no `renv.lock` file):

Image

toph-allen:
Remove alert icon from R Packages header that showed when there was no `renv` infrastructure.

## Intent

# Resolves #3088 

## Type of Change

- - [x] Bug Fix 

## Approach

Remove the alert from `RPackages.vue` and `home.ts`.

## Automated Tests

As far as I can tell, no automated tests cover this alert.

## Directions for Reviewers

- Run the extension in an R project which isn't using `renv`, confirm there is no alert icon in the R Packages header.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added more prominent status messages when rendering Quarto documents. #2940
 
+### Fixed
+
+- The "R Packages" section no longer shows you an alert if you aren't using renv. (3095)
+
 ## [1.22.0]
 
 ### Added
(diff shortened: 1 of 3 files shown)


marcosnav (approved):
Ezpz. Just confirming, we do not show the alert because
**a)** users have their own renv setup and that's ok, 
**b)** users not having renv setup is ok too now that it is automatically handled for them
Am I right?

toph-allen:
> Ezpz. Just confirming, we do not show the alert because
> 
> **a)** users have their own renv setup and that's ok, 
> 
> **b)** users not having renv setup is ok too now that it is automatically handled for them
> 
> Am I right?

Yeah, exactly. In the past, using R without renv was a problem state because you couldn't publish, but now using renv is optional. :)

PR #3169, opened 2025-10-02
feat: Detection support for plumber2 projects

### LINKED ISSUE #3150: feat: Detect plumber2 projects
## Description

Add detection for `plumber2` driven projects.

`rsconnect` had this done recently with https://github.com/rstudio/rsconnect/pull/1172, we can use this work as reference for changes needed to infer `plumber2` content.

marcosnav:
## Intent

Resolves #3150

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

- Look for `_server.yml` file presence and it's `engine:` field.

## User Impact

Publisher should be able to detect and create proper deployment configurations for Plumber2 projects.

## Automated Tests

Expanded existing plumber tests.

## Directions for Reviewers

Attempt to deploy a Plumber2 based project.

## Checklist

### DIFF
--- a/internal/publish/r_package_descriptions.go
+++ b/internal/publish/r_package_descriptions.go
@@ -72,8 +72,9 @@ func (p *defaultPublisher) recordExtraDependencies() (*util.AbsolutePath, error)
 		return nil, nil
 	}
 
-	// Create the file
-	depsPath := p.State.Dir.Join("__publisher_deps.R")
+	// Create the file __publisher_deps.R under .posit dir
+	// so it does not make the files tree list to jump or blink while deploying.
+	depsPath := p.State.Dir.Join(".posit", "__publisher_deps.R")
 	depsFile, err := depsPath.Create()
 	if err != nil {
 		return nil, err
@@ -116,9 +117,7 @@ func (p *defaultPublisher) getRPackagesWithPath(scanDependencies bool) (bundles.
 
 		extraDepsPath, err := p.recordExtraDependencies()
 		if err != nil {
-			agentErr := types.NewAgentError(types.ErrorRenvLockPackagesReading, err, lockfileErrDetails{Lockfile: p.Dir.String()})
-			agentErr.Message = fmt.Sprintf("Could not scan R packages from project: %s", err.Error())
-			return nil, util.AbsolutePath{}, agentErr
+			log.Error("Could not record extra dependencies file", "error", err.Error())
 		}
 		if extraDepsPath != nil {
 			log.Debug("Including extra dependencies file for scanning", "file", extraDepsPath.String())
--- a/internal/publish/r_package_descriptions_test.go
+++ b/internal/publish/r_package_descriptions_test.go
@@ -6,14 +6,14 @@ import (
 	"errors"
 	"testing"
 
-	"github.com/posit-dev/publisher/internal/publish/publishhelper"
-
 	"github.com/posit-dev/publisher/internal/bundles"
 	"github.com/posit-dev/publisher/internal/config"
+	"github.com/posit-dev/publisher/internal/contenttypes"
 	"github.com/posit-dev/publisher/internal/events"
 	"github.com/posit-dev/publisher/internal/inspect/dependencies/renv"
 	"github.com/posit-dev/publisher/internal/logging"
 	"github.com/posit-dev/publisher/internal/logging/loggingtest"
+	"github.com/posit-dev/publisher/internal/publish/publishhelper"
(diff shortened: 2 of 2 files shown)


internal/inspect/detectors/plumber.go
```diff
@@ -3,38 +3,151 @@ package detectors
...
+				cfg.Files = append(cfg.Files, fmt.Sprint("/", routesFileString))
+			}
+		}
+	}
+}
+
+func (d *PlumberDetector) inferByServerFile(base util.AbsolutePath, entrypoint util.RelativePath) (*config.Config, error) {
+	for _, serverFile := range possiblePlumberServerFiles {
+		// Does a server file exist?
+		serverFilePath := base.Join(serverFile)
```
karawoo, line 109: If they have both `_server.yml` and `_server.yaml` it looks like this will use just the first one, yes? Hopefully this is an unlikely scenario, but in rsconnect we error if we detect both.

internal/inspect/detectors/plumber.go
```diff
@@ -3,38 +3,151 @@ package detectors
...
 
-func NewPlumberDetector() *PlumberDetector {
+type plumberServerMetadata struct {
+	Engine      string `yaml:"engine"`
+	Routes      any    `yaml:"routes"` // string when single file or []string when multiple files
+	Constructor string `yaml:"constructor"`
+}
+
+// rsconnect looks for these specific entrypointPath filenames and server files.
+var possiblePlumberEntrypoints = []string{"plumber.R", "entrypoint.R"}
```
karawoo, line 32: should these maybe also include lower case extensions `plumber.r`/`entrypoint.r`?

internal/inspect/detectors/plumber.go
marcosnav: Yes, this will only use the first one that is found. With this, I'm not convinced it is worth checking if both are present and erroring out. Still happy to re-consider if there is something you think I'm missing.

marcosnav:
@karawoo @jonkeane This one is ready for another look! The implicit dependencies check is in now, so required dependencies for some content types that are not imported with `library(...)` will be found this time.

extensions/vscode/src/constants.ts
```diff
@@ -22,6 +22,8 @@ export const ENTRYPOINT_FILE_EXTENSIONS = [
   ".qmd",
   ".r",
   ".rmd",
+  ".yml",
+  ".yaml",
```
marcosnav: Allowing YAML files to be picked as entrypoint.

internal/inspect/r.go
```diff
@@ -82,7 +82,7 @@ func (i *defaultRInspector) InspectR() (*config.R, error) {
 	rProjectRequires := interpreters.NewRProjectRRequires(i.base)
 	rRequires, err := rProjectRequires.GetRVersionRequirement()
 	if err != nil {
-		i.log.Warn("Error retrieving required R version", err)
+		i.log.Warn("Error retrieving required R version", "error", err.Error())
```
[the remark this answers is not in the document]
jonkeane: More morbid curiosity, but what do the logs look like with this change?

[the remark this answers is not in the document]
jonkeane: Ah ok, hmmm for this one it's not too bad (though having `...Error ...: error=...` a little silly. Though also: warn + error seem contradictory here, no? Is it warn-level or is it error-level?

For the manifest one, that's going to make something like the following right?

```
level=DEBUG msg="Generated manifest:" manifest="some path"
```

And that seems a little extra silly having manifest twice `..manifest:" manifest=...` Maybe instead we should do something like inserting the manifest string into the first string and have no named arguments there? Is that `msg` part parsed out by something that we want to keep it short?

Very much agreed that `!BADKEY` is bad, but I think we could improve these. And of course, this is definitely outside of the scope of this PR, but something for us to clean up.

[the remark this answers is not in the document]
marcosnav: The R version error is shown as `WARN`, it is indeed contradictory but the reason (I think) is that not able to get the version data should not halt the process. Maybe it can still be `WARN` level but the message on lines of `"Could not retrieve R version" details="details here"`? 🤔 

The manifest one shows the whole resulting manifest content
```
level=DEBUG msg="Generated manifest:" manifest="{"R",.. "Version".. "Dependencies".....}"
```

> Maybe instead we should do something like inserting the manifest string into the first string

Yes, the manifest key there is pointless and everything can be under `msg`.

PR #3179, opened 2025-10-08
check for error earlier in matchingWalker.Walk to avoid nil deref

mslynch:
## Intent

Related to https://github.com/posit-dev/publisher/issues/3180.

When deploying in this situation, it results in an internal server error due to a nil pointer dereference. We're not checking the error we get from walking the filesystem soon enough.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Moved the error handling up so it can catch this case.

## User Impact

Users won't hit an internal server error in this case; the filesystem walk will complete successfully and an error will be logged about the file that caused an error.

## Automated Tests

Added a test to verify this case.

## Directions for Reviewers

Use steps in linked ticket and hit deploy.

## Checklist

### DIFF
--- a/internal/bundles/matcher/walker_test.go
+++ b/internal/bundles/matcher/walker_test.go
@@ -23,9 +23,8 @@ import (
 type WalkerSuite struct {
 	utiltest.Suite
 
-	fs     afero.Fs
-	cwd    util.AbsolutePath
-	subdir util.AbsolutePath
+	fs  afero.Fs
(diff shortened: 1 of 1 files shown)


dotNomad:
I'm going to close this PR up as a stale. We can re-open if we want to take another attempt at it.

PR #3249, opened 2025-11-03
Dependabot bumps

jonkeane:
Consolidating a few, checking the failures

### DIFF
--- a/test/e2e/tests/deployments.cy.js
+++ b/test/e2e/tests/deployments.cy.js
@@ -144,12 +144,10 @@ describe("Deployments Section", () => {
       // Set public access via helper and then deploy
       cy.getPublisherTomlFilePaths("examples-shiny-python").then(
         ({ config }) => {
-          // return cy.savePublisherFile(config.path, {
-          //   connect_cloud: { access_control: { public_access: true } },
-          // });
+          // Write both Python version and public access in one operation
           return cy.writeTomlFile(
             config.path,
-            "[connect_cloud]\n[connect_cloud.access_control]\npublic_access = true",
+            "version = '3.11'\n[connect_cloud]\n[connect_cloud.access_control]\npublic_access = true",
           );
         },
       );


tyatposit:
> The e2e failures are real. Though I kinda suspect that the first failure is then causing the others. What I'm seeing (on [the shiny PCC test](https://github.com/posit-dev/publisher/blob/1576ba090fe70cff5d7d524cff08c43a5a21bacb/test/e2e/tests/deployments.cy.js#L118-L223) is:
> 
> > Unable to resolve dependency: unable to resolve dependencies from requirements.txt: This content's dependencies could not be resolved. Check your requirements.txt.. Please check the logs for more information. See Troubleshooting docs for help.
> 

I was able to see the full error when i finally got it to fail locally:

`2025-11-03T23:39:5573761Z Error resolving dependencies: Because the current Python version (3.10.18) does not satisfy Python>=3.11 and scipy==1.16.3 depends on Python>=3.11, we can conclude that scipy==1.16.3 cannot be used. And because you require scipy==1.16.3, we can conclude that your requirements are unsatisfiable.`

I added to the deployment test, a field that sets the python version on PCC, and it seemed to help it pass for me locally.  I couldn't reproduce the other failures locally, but I guess you were right and were related somehow since the rerun passed.  🤷‍♂️

jonkeane:
Thanks for finding that + fixing it 🙏 . It hadn't occurred to me that we were seeing a different version of Python on the server and _that_ was the issue — but that makes sense now! Do you know if we mark this test item as requiring a higher python with a `.python-version` file or the like, would that work too? (see: #2617 ) It _should_ but it would be good to confirm this is the case on the cloud path too.

> I added to the deployment test, a field that sets the python version on PCC, and it seemed to help it pass for me locally. I couldn't reproduce the other failures locally, but I guess you were right and were related somehow since the rerun passed. 🤷‍♂️

Yeah, when I looked at the screenshots, these all looked to be because we (re)tried the failing test ^^^ each of those added a new deployment to the picker, and the subsequent e2e tests expect to find deployments without scrolling. Those two additional deployments meant that although the deployments were there, one needs to scroll the picker to find them.

PR #3263, opened 2025-11-10
chore: Bundle a few dependabot updates

jonkeane:
Trying to overcome the CI failures the individual PRs had.

This bundles #3259, #3258, and #3257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


nealrichardson:
Sounds like you want some dependabot groups: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

jonkeane:
> Sounds like you want some dependabot groups: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

Yeah this would also be good to do for sure, though I was bundling here more to prove/disprove that the CI failure is separate from the external PR

marcosnav (approved):

PR #3321, opened 2025-12-03
chore: Setup TypeScript bundler moduleResolution 

### LINKED ISSUE #3304: Investigate updated esbuild + TypeScript setup
With the work to upgrade Vitest to version 4 in #3303 there was a workaround needed to handle sub-path imports. This makes me think there is a better way to handle our `tsconfig` setup alongside `esbuild`. Ideally our dependencies shouldn't cause TypeScript errors that become irrelevant with our `esbuild` bundler setup.

Some things that will be helpful:
- [investigate the latest VS Code extension template created with `yo`](https://code.visualstudio.com/api/get-started/your-first-extension)
- [investigate the latest `esbuild` and `webview` VS Code extension examples](https://github.com/microsoft/vscode-extension-samples)

dotNomad:
This PR does a handful of small things related to our `esbuild`, TypeScript, and VS Code tooling set ups in an effort to use `moduleResolution: bundler` to handle non-CommonJS dependencies more easily. I'll try to detail them as best I can:

1. `esbuild` has some recommended TypeScript configuration options
    - https://esbuild.github.io/content-types/#isolated-modules
        - We weren't using this one so it was added
    - https://esbuild.github.io/content-types/#es-module-interop
        - This one already existed in our config
2. Since we use `esbuild` we can use `module: preserve` and `moduleResolution: bundler`
    - There are more details in [my comment in the parent issue](https://github.com/posit-dev/publisher/issues/3304#issuecomment-3564980555)
    - This let's TypeScript know we are using a bundler. Previously it thought we were using `Node10` since `module` was set to `commonjs`. Some dependencies were using things like sub-path imports that were causing issues with this setup.
    - Ultimately this new setup is more correct since we are using `esbuild` instead of the TypeScript compilation.
3. I setup a new `tsc-watch` npm script and watch task for VS Code. Now when you run the extension a `esbuild` and `tsc-watch` are run to show `esbuild` and typing errors. Makes working on the extension just a bit easier.
4. Moved our TypeScript target up to `ES2023`.
    - Some of the [VS Code extension samples](https://github.com/microsoft/vscode-extension-samples) use 2023-2024, so figured we could get access to slightly newer APIs. Especially since we use `esbuild`.
5. Removed the `tsconfig.json` `paths` workarounds we had. Those handled some of the import issues that non-`commonjs` dependencies had with TypeScript. We no longer need those with `moduleResolution: bundler`.

---

The way we do our extension tests also needed to be updated. Previously we used `tsc` to compile the extension tests, but with the new setup we need to use a bundler - `esbuild` - similar to our extension setup.

To accomplish this I setup a `esbuild.tests.mjs` script and a `tsconfig.test.json`. From there we can use the newer [`vscode-test-cli`](https://github.com/microsoft/vscode-test-cli) to do what our `src/test/runTest.ts` file was doing.  This greatly simplifies VS Code extension setup requiring only the test suite files. `vscode-test-cli` did need a configuration file so I created it `.vscode-test.mjs`.

## Intent

Resolves #3304

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -634,7 +634,7 @@
   },
   "scripts": {
     "vscode:prepublish": "npm run --prefix webviews/homeView build && npm run esbuild-base -- --minify && npm run build-font",
-    "esbuild-base": "tsc --noEmit && esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
+    "esbuild-base": "tsc --noEmit --project tsconfig.json && esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
     "esbuild": "npm run esbuild-base -- --sourcemap",
     "esbuild-watch": "npm run esbuild-base -- --sourcemap --watch",
     "tsc-watch": "tsc --noEmit --watch --project tsconfig.json",


dotNomad:
Getting some consistent e2e test errors. I'm not sure why since the esbuild of the extension is largely the same. I'm running the tests of `main` as well to compare.

dotNomad:
> Getting some consistent e2e test errors. I'm not sure why since the esbuild of the extension is largely the same. I'm running the tests of `main` as well to compare.

`main` is also having the same e2e test errors.

sagerb (approved):
I was able to run and build everything locally, and the changes to the files make sense to me.

Issue #3089, opened 2025-09-14
bug: ephemeral `renv.lock` file creation sometimes doesn't get all of the dependencies when it is listed in `[files]`


jonkeane:
This is similar to #3063 when deploying a shiny app, I tried creating renv infrastructure first (publishing) then deleting the `renv.lock` file. Because we don't (yet) remove the `renv.lock` file form the config file's `files` array, it is still there.  I found that this would create the ephemeral lockfile just fine, but the lockfile (and thus the manifest) would be missing critical packages (and most critically shiny itself!) 

I'm attaching the ephemeral lockfiles that show up in .posit/publish/deployments to help show this, but I've been able to recreate it with the steps above very reliably.

The `renv.lock` file when I've deleted the lockfile, but it is still listed in the `files` array: 
[renv.lock-in-files.txt](https://github.com/user-attachments/files/22319039/renv.lock-in-files.txt)

The `renv.lock` file when I've deleted the lockfile *and* removed it from the `files` array too:
[renv.lock-not-in-files.txt](https://github.com/user-attachments/files/22319042/renv.lock-not-in-files.txt)

This might seem like an edge case, but it is exactly the state that someone would get to if they tried once with renv, and then came back to a project later and weren't able to update the lockfile with renv and so want to start from scratch.

PR #3358, opened 2025-12-17
Be more robust when passing the R scanner nonexistent files

jonkeane:
Don't fail to add required dependencies when we pass a non-existent path to the scanner. Will supersede #3357, but want to do TDD, so starting with a failing test

## Intent

Resolves: #3089

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   its validation: So long as there is a GUID in the string, it will be attempted. This
   means that one can use the in-app URL, the standalone URL, or simply the GUID again. (#3341)
 
+- R dependency detection succeeds even if non-existent files are included in the configuration. (#3089)
+
 ## [1.26.0]
 
 ### Added


.github/workflows/agent.yaml
```diff
@@ -35,8 +35,8 @@ jobs:
 
       - name: Install R with rig
         run: rig add release
-      - name: Install renv package
-        run: Rscript -e "install.packages('renv')"
+      - name: Install R packages
+        run: Rscript -e "install.packages(c('renv', 'glue'))"
```
jonkeane: We needed one extra dependency to check if it exists, glue is low-dependency so should be just fine + quick to install

internal/inspect/dependencies/renv/project_dependencies.go
```diff
@@ -85,10 +85,16 @@ func (s *defaultRDependencyScanner) ScanDependenciesInDir(paths []string, target
...
+	deps <- character()
+	for (path in rPathsVec) {
+		tryCatch({
+			d <- renv::dependencies(path = path, progress = FALSE)
+			deps <- c(deps, d$Package[!is.na(d$Package)])
+		}, error = function(e) {
+			# Silently skip paths that cause errors (e.g., non-existent files, directories)
+			invisible()
+		})
+	}
```
jonkeane: I opted to iterate through the files provided around the tryCatch here. This _might_ be less-than-ideal performance-wise if there are a very large number files, but it means that we are robust to even files getting deleted during this process.

PR #3377, opened 2026-01-08
feat: open connect remote content

amol-:
## Intent

Allow opening a remote bundle deployed on a connect server.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Register a custom url handler for the `"vscode://posit.publisher/connect`" url
That accepts the server and content guid as arguments and fetches those.

## User Impact

The user can open remote content without having to manually fetch it first.

## Automated Tests

TBD

## Directions for Reviewers

I tried to isolate the core logic into `OpenConnectContent.ts` and `post_open_connect_content.go`

## Checklist

### DIFF
--- a/internal/clients/connect/client.go
+++ b/internal/clients/connect/client.go
@@ -46,4 +46,6 @@ type APIClient interface {
 	GetCurrentUser(logging.Logger) (*User, error)
 	GetIntegrations(logging.Logger) ([]Integration, error)
 	GetSettings(util.AbsolutePath, *config.Config, logging.Logger) (*AllSettings, error)
+	LatestBundleID(types.ContentID, logging.Logger) (types.BundleID, error)
+	DownloadBundle(types.ContentID, types.BundleID, logging.Logger) ([]byte, error)
 }
--- a/internal/clients/connect/mock_client.go
+++ b/internal/clients/connect/mock_client.go
@@ -127,3 +127,16 @@ func (m *MockClient) GetSettings(base util.AbsolutePath, cfg *config.Config, log
 	}
 	return args.Get(0).(*AllSettings), args.Error(1)
 }
+
+func (m *MockClient) LatestBundleID(contentID types.ContentID, log logging.Logger) (types.BundleID, error) {
+	args := m.Called(contentID, log)
+	return args.Get(0).(types.BundleID), args.Error(1)
+}
+
+func (m *MockClient) DownloadBundle(contentID types.ContentID, bundleID types.BundleID, log logging.Logger) ([]byte, error) {
+	args := m.Called(contentID, bundleID, log)
+	if args.Get(0) == nil {
+		return nil, args.Error(1)
+	}
+	return args.Get(0).([]byte), args.Error(1)
+}
--- a/internal/services/api/post_open_connect_content.go
+++ b/internal/services/api/post_open_connect_content.go
@@ -57,12 +57,7 @@ func PostOpenConnectContentHandlerFunc(lister accounts.AccountList, log logging.
 			InternalError(w, req, log, err)
 			return
 		}
-		connectClient, ok := client.(*connect.ConnectClient)
-		if !ok {
-			InternalError(w, req, log, errors.New("unexpected connect client type"))
-			return
-		}
-		bundleID, err := connectClient.LatestBundleID(types.ContentID(body.ContentGUID), log)
+		bundleID, err := client.LatestBundleID(types.ContentID(body.ContentGUID), log)
 		if err != nil {
 			if handleConnectAPIError(w, log, err, normalizedURL, body.ContentGUID) {
 				return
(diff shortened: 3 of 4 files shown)


extensions/vscode/src/extension.ts
```diff
@@ -147,6 +154,9 @@ async function initializeExtension(context: ExtensionContext) {
     commands.registerCommand(Commands.ShowPublishingLog, () => {
       commands.executeCommand(Commands.Logs.Focus);
     }),
+    commands.registerCommand(Commands.OpenConnectContent, () =>
+      promptOpenConnectContent(),
+    ),
```
amol-: This isn't super necessary, but I guess it's a nice to have and it made possible to implement the test (as it's not easy to trigger url handler otherwise)

test/e2e/support/sequences.js
```diff
@@ -89,17 +94,19 @@ Cypress.Commands.add(
...
+          .get(".quick-input-widget")
+          .find(
+            `[aria-label="${projectDir}/${entrypointFile}, Open Files"], [aria-label="${entrypointFile}, Open Files"]`,
+          ),
+      10,
+      700,
+    ).then(($el) => {
+      cy.wrap($el).scrollIntoView();
+      cy.wrap($el).click({ force: true });
+    });
```
amol-: For reasons I don't understand, the "$projectDir" check wasn't enough.
Sometimes (you can always replicate it if you run the test twice) the $projectDir has a value, but the file is not inside $projectDir, it's in the root.

In my case, sometimes the file was in "static/index.html" and sometimes just "index.html".

I noticed that if I run the test twice, the first time is "static/index.html" while the second time is just "index.html".

Changing the find to look for both cases was the only robust way I could find to make tests pass in a reliable way.

test/e2e/support/commands.js
```diff
@@ -319,6 +319,63 @@ Cypress.Commands.add("loadTomlFile", (filePath) => {
     });
 });
 
+// runCommandPaletteCommand
+// Purpose: Invoke a command by label through the VS Code command palette.
+Cypress.Commands.add("runCommandPaletteCommand", (commandLabel) => {
```
amol-: Initially I attempted with with `{ctrl}{shift}p`,
it wasn't reliable and randomly failed, so I asked Claude to write a code block that did excplicitly click on the UI to open the command and use "> X" syntax to execute it.

test/e2e/support/commands.js
```diff
@@ -319,6 +319,63 @@ Cypress.Commands.add("loadTomlFile", (filePath) => {
...
+  });
+});
+
+// quickInputType
+// Purpose: Fill a quick input prompt and submit the value.
+Cypress.Commands.add("quickInputType", (promptText, value) => {
+  cy.get(".quick-input-message").should("contain.text", promptText);
+  cy.get(".quick-input-widget input").clear().type(`${value}{enter}`);
+});
+
```
amol-: I couldn't find a ready made utility for this.
It just fills a value in a dialog asking for input in Code

extensions/vscode/src/extension.ts
```diff
@@ -50,6 +56,11 @@ export enum SelectionIsPreContentRecord {
   false = "false",
 }
 
+let publisherStateResolve: ((state: PublisherState) => void) | undefined;
+const publisherStateReady = new Promise<PublisherState>((resolve) => {
+  publisherStateResolve = resolve;
+});
```
amol-: When url handlers are fired, the PublisherState might not yet exist.
So the support for the virtual file system needs to receive a promise instead of the PublisherSate object itself, so that it can wait for it to be ready before trying to open content from connect (as it wouldn't have the authentication keys otherwise)

extensions/vscode/src/extension.ts
```diff
@@ -162,6 +180,13 @@ async function initializeExtension(context: ExtensionContext) {
   );
 
   context.subscriptions.push(new PublisherAuthProvider(state));
+  context.subscriptions.push(
+    workspace.onDidChangeWorkspaceFolders((event) => {
+      event.removed.forEach((folder) => {
+        clearConnectContentBundleForUri(folder.uri);
+      });
+    }),
+  );
```
amol-: If we close the folder, we clear the bundle content from memory.
Or we would end up leaking memory as the user jumps from one bundle to the other

marcosnav (commented):
I haven't finished reviewing but already have a comment that could require quite a few changes.

extensions/vscode/src/api/client.ts
```diff
@@ -48,11 +50,16 @@ class PublishingClientApi {
         return response;
       },
       (error) => {
-        // Decode data returned for 500 errors
-        if (error.response.status === 500) {
+        // Decode data returned for 500 errors when the payload is readable text.
+        if (
+          error.response?.status === 500 &&
+          typeof error.response.data === "string"
```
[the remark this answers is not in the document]
marcosnav: Would you get a `tar` in `response.data` even if the status is 500?

PR #3387, opened 2026-01-14
feat(logs): reveal failure line in logs

### LINKED ISSUE #3236: Expand the publisher log section with an error when deployment fails
When the deployment fails, the log section in which the error occurred should be expanded, with the active visible line being shown as the last line of that section (or the specific error if we can find it).

Related issue (which won't be needed when this is completed):
https://github.com/posit-dev/publisher/issues/3192

dotNomad:
This PR reveals a specific `*/failure` log event type in the logs view when it is visible, or the visibility of it is changed to be visible.

## Intent

Resolves #3236

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Look to the diff comments to see different approaches.

## User Impact

When the logs are visible, or made visible, and there is a failure it is now even easier to find.

## Automated Tests

There are new unit tests utilizing mocked VS Code APIs included.

## Directions for Reviewers

Manufacture a deploy failure to see the new behavior. You can toggle the `autoOpenLogsOnFailure` setting to see how it behaves in the different cases.

## Checklist

### DIFF
--- a/extensions/vscode/src/views/logs.test.ts
+++ b/extensions/vscode/src/views/logs.test.ts
@@ -812,6 +812,12 @@ describe("LogsTreeDataProvider", () => {
     });
 
     test("should not reveal failure line when view is not visible", () => {
+      // The extension setting `autoOpenLogsOnFailure` determines if the logs
+      // view is opened when any failure comes through.
+      // This test tests that when the view is not visible, the reveal is not
+      // called avoiding opening the logs view if the user has closed it or
+      // the `autoOpenLogsOnFailure` setting is false.
+
       const { stream, emit } = createMockEventStream();
       const context = createMockContext();
 


extensions/vscode/src/views/logs.ts
```diff
@@ -485,6 +485,7 @@ export class LogsTreeDataProvider implements TreeDataProvider<LogsTreeItem> {
             stage.events.push(msg);
           }
           this.refresh();
+          this.revealFailure();
```
dotNomad: When we get a failure that failure is now revealed if the TreeView is visible.

extensions/vscode/src/views/logs.ts
```diff
@@ -564,13 +565,11 @@ export class LogsTreeDataProvider implements TreeDataProvider<LogsTreeItem> {
...
-   * the first LogsTreeStageItem with `LogStageStatus.failed` or
-   * the last LogsTreeLogItem of that stage if the stage has events
-   *
-   * Returns `undefined` if no LogsTreeStageItems have `LogStageStatus.failed`
+   * Returns the failure log item (last event) from the first failed or canceled stage.
+   * If the failed stage has no events, returns the stage item itself.
+   * Returns `undefined` if no stages have failed or been canceled.
    */
-  private getFailedStageItem(): LogsTreeItem | undefined {
+  private getFailureItem(): LogsTreeItem | undefined {
```
dotNomad: This does the same thing, but made it a bit clearer that it gets a failure item or a stage if there isn't one (in the case of cancellation)

extensions/vscode/src/views/logs.ts
```diff
@@ -701,7 +704,7 @@ export class LogsTreeDataProvider implements TreeDataProvider<LogsTreeItem> {
       treeView,
       treeView.onDidChangeVisibility((e) => {
         if (e.visible) {
-          this.revealFailingState(treeView);
+          this.revealFailure();
```
dotNomad: The `onDidChangeVisibility` is still used so the failure is revealed if the TreeView is closed when the failure occurs. This causes the logs view to show the failure when it is changed to be `visible`

dotNomad: I'll add a comment in the test to clarify what I did in https://github.com/posit-dev/publisher/pull/3387#discussion_r2692350932

Issue #3168, opened 2025-10-02
bug: The publisher output stream names are confusing


jonkeane:
For output streams, qe have two named ones: `Posit Publisher Authentication` and `Posit Publisher Deployment` Practically speaking, the Authentication one has very few lines in it ever (I haven't yet traced what uses it) and the s-called Deployment one has everything in it. 

<img width="238" height="51" alt="Image" src="https://github.com/user-attachments/assets/76381b1b-e3a4-49ae-a65f-20bbded5d2d2" />

What about if we condense these two into one and call it simply `Posit Publisher`? That matches what other extensions do generally (where their output has a single stream and it is ~the same name as the extension).

PR #3390, opened 2026-01-15
Consolidate all of our logging into one channel

jonkeane:
## Intent
Consolidate our logs into one stream, reduce confusion.

Resolves: #3168

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Consolidate our three (3!) log output channels into one place.

## User Impact

No more guessing "is this the place?"

## Automated Tests

yes

## Directions for Reviewers

See fewer logs?

## Checklist

- [ ] I have updated the _root_ [CHANGELOG.md](../CHANGELOG.md) to cover notable changes. (TODO)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -44,4 +44,4 @@ If applicable: find this in Connect Cloud at https://connect.posit.cloud/my-acco
 
 ## Publisher output and logs
 
-<!-- Publisher output helps us to identify the problem, find it in the "Output" window → "Posit Publisher Deployment" option (see https://github.com/posit-dev/publisher/blob/main/docs/troubleshooting.md#publisher-output). Paste Publisher output here: -->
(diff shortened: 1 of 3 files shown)


extensions/vscode/src/servers.ts
```diff
@@ -52,14 +49,28 @@ export class Server implements Disposable {
...
-          // Write stderr to output channel
-          this.outputChannel.append(data.toString());
+          stderrBuffer += data.toString();
+          const lines = stderrBuffer.split("\n");
+          // Keep the last incomplete line in the buffer
+          stderrBuffer = lines.pop() ?? "";
+          // Log each complete line
+          for (const line of lines) {
+            logAgentOutput(line);
+          }
```
dotNomad: Seems like a good candidate for a `logging` function. Totally fine to leave it here.

dotNomad (commented):
Fantastic! So glad to see these combined.

Tested out locally, and works fantastic.

Just two things I think we need prior to merging:

- Let's add a CHANGELOG item
- and change the line in our troubleshooting doc from "selecting the "Posit Publisher Deployment" option." to "selecting the "Posit Publisher" option since that is now its name.

dotNomad (approved):
Fantastic 🎉 :shipit:

PR #3395, opened 2026-01-17
feat: Pre-flight validate the entrypoint file exists

### LINKED ISSUE #2964: Pre-flight checks, validate the entrypoint in the configuration does exist on the system
Noticed at https://github.com/posit-dev/publisher/pull/2936#issuecomment-3230076913

It seems we are not validating that the entrypoint in the deployment configuration does exist, which we could do as part of the pre-flight checks to prevent failed deployments sooner.

dotNomad:
Adds a pre-flight check to validate that the entrypoint file, specified in the configuration, exists prior to deploying to Connect.

## Intent

Resolves #2964

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Added `checkEntrypoint` validation alongside where we do `checkRequirementsFile`.

It does skip validation on any module references, which happens with Shiny Express. The entrypoint looks like `shiny.express.app:app_2e_py` in that case so entrypoints containing `:` are skipped.

I skipped them rather than attempting to see if that file exists or refactoring the order in which we swap out the `EntrypointFile`. To swap in the module reference for Shiny Express, the file has to be read - so we know it exists already.

## User Impact

Users will now receive an immediate, clear error message when attempting to deploy with a missing entrypoint file, rather than having the deployment fail on the server.

## Automated Tests

Unit tests were added for the new functionality 🎉 

## Directions for Reviewers

1. Create a config with a non-existent entrypoint file
2. Attempt to deploy
3. Verify an early failure occurs with a clear error message about the missing entrypoint

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Deployments now fail early with a clear error if the entrypoint file does not exist. (#2964)
 - Log messages in the Publisher Logs panel are no longer duplicated when multiple deployments are made. (#3069)
 - Fix creating deployments of R Markdown documents that include parameters on Connect Cloud. (#3388)
 - The Publisher outputs (detailed logs) have been consolidated into one channel called "Posit Publisher". (#3168)


jonkeane:
I'm a soft -1 on making this change be a hard-stop error for two reasons (1) we don't have an actionable way for someone to resolve this (2) if we get any of the detection wrong, we prevent people from deploying. While these deployments aren't in a good state, they do actually function (e.g. if you have a static website that you send up, say that `index.html` is the entrypoint, but only send up a few other html files, you can get to them at the standalone URL 

As for (2): 
In the original PR, one suggestion was:

> Maybe for static content in general, not just Quarto rendered content, we could ping after deployment and eval that it is a 2XX and not 4XX? The warn the user? 🤔 (thinking out loud here, ideas are welcome)

But even without doing a check for a 2XX, we could make this be a warning (or even an error) at the end of the deploy rather than forcing a positive stop in the deploy process. That way someone has a way out of this and aren't stuck.

as for (1): 
We don't surface the entrypoint (or being able to change it) anywhere in the UI. If all that needs to happen is add the entrypoint to the files list, that's easy enough. But if it's something else (like they need to specify a different entrypoint) for most folks going into the TOML and editing by hand is not what they are expecting + we don't give a good on ramp to that. IMO if we're going to make someone correct this before they move forward, we need to have all the necessary bits available in the UI first. And, even then there is still some of the concern of (2) that stands.

dotNomad:
Thinking through your feedback @jonkeane 

### On point 1

Great points. Do you think its beneficial to use all the same tooling here to warn (rather than fail) that the `entrypoint` in the configuration doesn't exist on the system?

### On point 2

> Maybe for static content in general, not just Quarto rendered content, we could ping after deployment and eval that it is a 2XX and not 4XX? The warn the user? 🤔 (thinking out loud here, ideas are welcome)

We left `404` and `405` open as non-errors since, for some content, that is a valid response for the initial page served by Connect. (An API that only non-GET handlers for example).

We should be able to use the Content configuration to determine its type and have additional logic if its static since that should always be a 2xx response (if valid).

Thanks for re-upping that comment

https://github.com/posit-dev/publisher/blob/785e085fe574b46e4ac9f7f9cd68edc191c0950c/internal/clients/connect/client_connect.go#L478-L499

PR #3404, opened 2026-01-21
fix: renv init failure in CI

karawoo:
## Intent

In previous successful CI runs, `install.packages()` installed renv from CRAN. Recently it's started installing from P3M (not sure why the change, it's not due to anything in the code afaict). `renv::init()` then started failing because the lockfile it generates includes a URL for CRAN, but not for RSPM (because there is no RSPM repo configured in `options("repos")`). 

Our code had special handling for packages marked as `Repository: CRAN`, but not for RSPM. When it encountered an RSPM package, it tried to look it up dynamically, which failed because of the lack of RSPM URL in the lockfile and resulted in an empty source - triggering an error that the package couldn't be reinstalled.

I believe we could fix by setting `options(repos = c(RSPM = "https://packagemanager.posit.co/cran/latest"))` before `renv::init()` in the test, however I think a better fix is to treat CRAN/RSPM/P3M/PPM as equivalent, so that's what I've done here.

## Checklist

### DIFF
--- a/internal/inspect/dependencies/renv/manifest_packages.go
+++ b/internal/inspect/dependencies/renv/manifest_packages.go
@@ -64,6 +64,18 @@ func findAvailableVersion(pkgName PackageName, availablePackages []AvailablePack
 	return ""
 }
 
+// isCRANLike returns true if the repository name is CRAN or a Posit Package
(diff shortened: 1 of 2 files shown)


dotNomad (approved):
Fantastic! Thanks for this

PR #3424, opened 2026-01-28
feat: Prefer quarto rendering of Jupyter Notebooks

### LINKED ISSUE #3423: feat: Allow Jupyter Notebooks to be rendered and published with Quarto instead of nbconvert
## Description
Currently, Posit publisher does not consistently show the option to use quarto render when publishing a notebook and it promotes nbconvert higher.  This has limitations in styling and functionality. It also takes the user out of the Posit ecosystem and does not leverage Quarto's capabilities.

This might be the code that needs to be changed https://github.com/posit-dev/publisher/blob/c8faeeb7806be351c94743792761f0f702bdc871/extensions/vscode/src/api/types/configurations.ts#L93

Example publishing of a notebook

Image

Example deployed notebook with nbconvert: https://connect.posit.it/connect/#/apps/dec401b8-712f-4d50-add6-9220c2c77728/5135

Example deployed notebook with quarto: https://connect.posit.it/connect/#/apps/c98b76ef-1daf-487d-8873-341e3a1d5578/tags/5137 
notice enhanced features and styling like copy cell, etc. 

UPDATED ASK: 

1. Ensure that both options are always shown when trying to publish a .ipynb file (i.e. selected entry file is .ipynb)
Image

2. In order to promote Quarto's richer publishing capabilties, put Quarto render as the first one and update text to: 
Quarto (render notebooks with Quarto's enriched publishing capabilities)

3. [NICE TO HAVE] Can we add a hyperlink in 2 out to Quarto 
https://quarto.org/ 

4. Rename jupyter-notebook to 
nbconvert (render with Jupyter nbconvert) 

5.  [NICE TO HAVE] Can we hyperlink 3 to nbconvert 
https://nbconvert.readthedocs.io/en/latest/

jonkeane:
## Intent

This does two things: 

* Makes quarto have a higher priority when it's available
* Offers quarto rendering (on Connect) even when folks don't have quarto installed locally

Resolves #3423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added automatic discovery of static assets for R Markdown projects. (#3233)
 - Added `positPublisher.enableConnectCloud` setting to disable publishing to Posit Connect Cloud. When set to `false`, Connect Cloud is hidden from the platform picker when creating new credentials, and existing Connect Cloud credentials are filtered from the deployment credential list. Existing Connect Cloud deployments remain accessible. (#3411)
 - Python notebooks now default to Quarto rendering on Connect, additionally, quarto is not required to be installed locally when sending notebooks to render on Connect with quarto. (#3423)
+- Quarto is now the preferred renderer for Jupyter notebooks as well as Rmarkdown and of course Quarto docs. (#3423)
 
 ### Fixed
 


internal/inspect/detectors/all.go
```diff
@@ -60,6 +60,25 @@ var preferredNames = []string{
...
 
+// contentTypePriority defines explicit ordering for content types.
+// Lower values are preferred (sorted first). Types not in this map
+// fall back to alphabetical ordering.
+var contentTypePriority = map[contenttypes.ContentType]int{
+	// Prefer Quarto over jupyter-notebook for .ipynb files
+	contenttypes.ContentTypeQuarto:          1,
+	contenttypes.ContentTypeJupyterNotebook: 2,
+	// Prefer Quarto over RMarkdown for .Rmd files
+	contenttypes.ContentTypeRMarkdown: 3,
```
m--: Out of curiosity, can you talk about the RMarkdown part of this? I don't see it mentioned on the issue this PR closes, and I'm trying to orient myself a bit. Thanks!

internal/inspect/detectors/all.go
```diff
@@ -60,6 +60,25 @@ var preferredNames = []string{
...
+}
+
+func getContentTypePriority(t contenttypes.ContentType) int {
+	if priority, ok := contentTypePriority[t]; ok {
+		return priority
+	}
+	// Types not in the map get a high priority number, then sort alphabetically
```
m--: Minor, but since we're not doing the sorting here, something like this might help to clarify? Also, `100`'s arbitrary, right? Is it worth mentioning that?

```suggestion
	// Types with the same priority number are sorted alphabetically.
	// Set an arbitrarily high priority number for any types not in the map so they sort at the end.
```

marcosnav (approved):
Looks good! Approving since I don't see any blocking feedback. Only one comment related to feature consistency with Rmds

internal/inspect/detectors/all.go
m--: So unrelated to the "Jupyter Notebooks" part of this, but still a useful thing to have. Thanks for the clarification 🎉 are there other things like this we'd want to prefer?

PR #3445, opened 2026-02-04
Fix empty responses from agent for credentials fetching

marcosnav:
## Intent

Improves error handling for credential operations and removes unused performance profiling code from the VSCode extension.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

- **Handle missing responses from Publisher agent** (`8c74f966e`)
  - Add defensive null checks when querying credentials from the Publisher API
  - Properly handle cases where the agent doesn't return a response during:
    - Credential list queries (`multiStepInputs/common.ts:223-229`)
    - Server URL validation (`multiStepInputs/newConnectCredential.ts:251-256`)
    - API key validation (`multiStepInputs/newConnectCredential.ts:454-459`)
  - Show warning messages to users when the agent is unreachable
  - Fix optional chaining in axios error interceptor (`api/client.ts:49`)

### Refactoring
- **Remove performance profiling** (`99001a215`)
  - Remove request timestamp tracking and duration logging from axios interceptors
  - Clean up unused `logDuration()` method
  - Simplify response interceptor logic

## User Impact

These changes make the extension more resilient to network errors or cases where the Publisher agent subprocess fails to respond, preventing silent failures during credential operations. In some cases, performance profiling dropping requests silently.

## Directions for Reviewers

- [ ] Test credential creation when Publisher agent is running normally
- [ ] Test credential creation when Publisher agent is stopped or unreachable
- [ ] Verify appropriate error messages are shown to users
- [ ] Test existing credential list functionality
- [ ] Verify Server URL validation still works correctly
- [ ] Verify API key validation still works correctly

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed credential operations failing silently when the Publisher agent is unreachable or doesn't respond. The extension now shows warning messages when the agent is unavailable during credential listing, server URL validation, or API key validation. (#3396)
+
 ## [1.32.0]
 
 ### Changed
--- a/extensions/vscode/src/api/client.ts
+++ b/extensions/vscode/src/api/client.ts
@@ -1,6 +1,6 @@
 // Copyright (C) 2025 by Posit Software, PBC.
 
-import axios, { AxiosResponse } from "axios";
+import axios from "axios";
 
 import { Credentials } from "./resources/Credentials";
 import { ContentRecords } from "./resources/ContentRecords";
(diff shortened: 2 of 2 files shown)


dotNomad:
@marcosnav are you still investigating here or is this ready for a review?

extensions/vscode/src/multiStepInputs/common.ts
```diff
@@ -220,14 +220,20 @@ export const getExistingCredentials = async (viewId: string) => {
     await showProgress("Initializing::newCredential", viewId, async () => {
       const api = await useApi();
       const response = await api.credentials.list();
-      credentials = response.data;
+      if (response) {
+        credentials = response.data;
```
dotNomad: If there is no `response` won't this fall into the `catch` or am I not understanding that right?

My understanding is that we'd have an `error` but no `error.response`

https://axios-http.com/docs/handling_errors

marcosnav: I suspect `performance.now` is maybe not consistent in different machines or distros.

dotNomad: Huh very interesting. Thanks for the clarification

Issue #3440, opened 2026-02-02
Unable to store credential or publish to Connect Cloud


alex-chisholm:
## Problem statement

After successfully authorizing Connect Cloud in the browser and seeing a success message in the Posit Publisher menu, the credential is not stored and I cannot publish.

## Steps to reproduce

1. Click + for New Credential
2. Select Posit Connect Cloud from menu
3. Redirected to the browser to authorize
4. Authorize and return to the IDE
5. Select one of my Connect Cloud accounts
6. See message in menu: "Successfully connected to Connect Cloud... Enter a unique nickname...."
7. Enter nickname and press enter
8. Nothing happens. Menu closes and credential not added.

**Expected behavior:** 

Credential is stored and I'm able to deploy.

**Actual behavior:**

Credential is not stored and I cannot deploy.

## System Details

- **Posit Publisher Version**: 1.28.0
- **Platform**: darwin arm64
- **IDE**: Positron 1.106.0

## Publishing information

Shiny App

## Additional Info

**Connect Cloud account name:** Alex-Chisholm

## Publisher output and logs

None

PR #3446, opened 2026-02-04
Fix Connect Cloud credential selection when there are multiple accounts

mslynch:
## Intent

Account selection is broken when there are multiple PCC accounts - when the user selects an account, no credential is created, and instead this is logged:

`[Extension Host] User has dismissed the New Credential flow. Exiting. (at console.<anonymous> (file:///Applications/Positron.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:96148:14))`

The bug was introduced here: https://github.com/posit-dev/publisher/pull/3303/changes#diff-85757b4df2fd959fdb90483f09e44e9c37757ff9cf7995217245e569f6680e29

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Fix the condition checking for account - We needed to check if the account from `.find` is present. The actual happy path was removed and only the fallback case was handled.

## User Impact

Allows users who have multiple accounts to be able to create credentials again.

## Automated Tests

No established pattern for unit tests in this area of the code.

## Directions for Reviewers

Set up access on your Connect Cloud user to have access to multiple accounts to be able to publish to, then try to add a credential for one of them.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Connect Cloud users who have permissions to publish to multiple accounts are able to create credentials again. (#3446)
+
 ## [1.34.0]
 
 ### Changed


rodriin (approved):
Nice catch, I suggested a small code change.

tyatposit (approved):
Should we add a unit test to try to catch this should something like it happen again?

dotNomad:
Marking this as a fix for #3440

karawoo (approved):

PR #3464, opened 2026-02-06
fix: detect rpy2 dependency to include R section in configuration

### LINKED ISSUE #3425: bug: Importing `rpy2` doesn't generate configuration with R section
If you select a Python file (ex. `app.py`) that imports `rpy2` to do something like `importr("ggplot2")` the generated configuration from Publisher doesn't include an `[r]` section even though it is required for the deployment.

There is a great example from @samperman at [`samperman/rpy2`](https://github.com/samperman/rpy2).

Select `app.py` as the entrypoint and notice that the configuration doesn't include `[r]`.

If it is added manually to the configuration it deploys, but Publisher should be better about discovering language usage.

zackverham:
## Summary

When a Python file imports `rpy2` (which allows Python to call R code), the generated configuration should include an `[r]` section since R is required at runtime.

This change:
- Adds `HasRpy2Dependency()` function in `pydeps` package to check if `rpy2` is in `requirements.txt`
- Updates `RequiresR()` in the R inspector to check for rpy2 usage when the content type is Python
- Adds unit tests for the new functionality

Fixes #3425

## Test plan

- [x] Unit tests added for `HasRpy2Dependency()` function
- [x] Unit tests added for `RequiresR()` with rpy2 detection
- [x] All existing tests pass
- [x] Manual test: Create a Python project with `rpy2` in `requirements.txt`, verify that the generated configuration includes an `[r]` section

### Test with samperman/rpy2 example
1. Clone https://github.com/samperman/rpy2
2. Select `app.py` as the entrypoint
3. Verify the configuration now includes `[r]` section


zackverham:
# Validation Notes


Attempting to produce a deployment config for the rpy2 project linked above produces the followwing: 

```
# Configuration file generated by Posit Publisher.
# Please review and modify as needed. See the documentation for more options:
# https://github.com/posit-dev/publisher/blob/main/docs/configuration.md
product_type = 'connect'
'$schema' = 'https://cdn.posit.co/publisher/schemas/posit-publishing-schema-v3.json'
type = 'python-shiny'
entrypoint = 'app.py'
validate = true
files = [
  '/app.py',
  '/requirements.txt',
  '/renv.lock',
  '/.posit/publish/rpy2-3E1J.toml',
  '/.posit/publish/deployments/deployment-IF5Q.toml'
]
title = 'rpy2'

[python]

[r]

```

internal/inspect/dependencies/pydeps/project_dependencies_test.go
karawoo: Could we add a test for the behavior if requirements.txt contains a package like `rpy2-arrow`? I think under the logic of this PR that will be treated as "needs R" which I think is probably correct.

karawoo (approved):

PR #2961, opened 2025-08-28
restart agent when it dies

mslynch:
## Intent

When the agent process exits, the extension code starts a new one.

Resolves https://github.com/posit-dev/publisher/issues/2933.

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Automated Tests

I suppose this could be tested... but it's probably not easy.

## Directions for Reviewers

I was able to manually test this myself by adding a panic in the go code.

## Checklist

- [ ] I have updated [CHANGELOG.md](../CHANGELOG.md) to cover notable changes.

PR #3482, opened 2026-02-09
fix: do not restart server process when shutting down

### LINKED ISSUE #3195: The Publisher extension keeps running even when Positron is not
## Problem statement

I've noticed that `publisher` continues to run even after i close out of Positron

## Steps to reproduce

1. Open projects in positron and publish
2. Exit Positrong

**Expected behavior:**

I would expect `publisher` process to exit

**Actual behavior:**

Many `publisher` are running (I have 15 running right now, 1 per project published?)

## System Details

MacOS Tahoe

## System Information

- **Posit Publisher Version**: 99.0.0
- **Platform**: darwin arm64
- **IDE**: Positron 1.103.0

## Publishing information

Various content types (mostly quarto) all to Posit Connect Claude

christierney:
## Intent

The server process auto-restart (#2961) should not trigger if we are intentionally stopping the process, as at shutdown.

Resolves #3195

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

I used Claude to help diagnose the problem and implement the fix.

I was able to observe the buggy behavior sporadically by starting VS Code or Positron, publishing content, and then closing the editor. Sometimes, though not usually, a `publisher` process would linger after the editor was gone.

### things I fixed

The core of the problem seems to be that we have a restart handler that unconditionally restarts the process when it closes, regardless of why it closed. The targeted fix for this was to introduce a `stopping` flag that prevents restart.

I also added error handling around the status bar message setting, because it seems like the behavior of that API might be unpredictable at shutdown time. It might be better to just remove that code---do we really need to send a message when the editor is closing anyway?

### tests

There were no unit tests of Server. I got Claude to write some, which rely on a lot of mocking. We could refactor Server in the future to make it more testable (e.g. inject a ConnectionChecker that would be more easily mocked.)

I used `Run Extension` to manually test a deployment to Dogfood, and verified that the `publisher` process closed when I closed the editor.

### things that seem off but I didn't change

The `dispose` function seems redundant. We might want to consider either embracing the use of `Disposable` (add our server to `context.subscriptions` and let VS Code dispose of it), or removing this function as it might confuse people.

The call to `isDown` is kind of useless as we don't check the return value but I opted to keep this change small and leave that alone.

I'll attach `PLAN.md` with all the notes from the session, including some potential future improvements.

## User Impact

The user should not notice any changes in behavior, but they should no longer have orphaned `publisher` processes running after they exit the editor.

## Automated Tests

Added `servers.test.ts`.

## Directions for Reviewers

Please do review. 🙏🏻  I relied a lot of Claude here although I have read and understood everything in the PR.

## Checklist

### DIFF
--- a/extensions/vscode/src/servers.ts
+++ b/extensions/vscode/src/servers.ts
@@ -104,14 +104,14 @@ export class Server implements Disposable {
    * @returns {Promise<void>} A Promise that resolves when the server stops.
    */
   async stop(): Promise<void> {
+    // Set stopping flag immediately to prevent auto-restart race conditions
+    this.stopping = true;
+
     // Check if server is down
     if (await this.isDown()) {
       // Do nothing if server is already down
       return;
     }
-
-    // Set stopping flag to prevent auto-restart
-    this.stopping = true;
 
     // Display status message to user
     // Guard against VS Code shutdown - API may be unavailable
--- a/extensions/vscode/src/servers.test.ts
+++ b/extensions/vscode/src/servers.test.ts
@@ -89,12 +89,14 @@ describe("Server", () => {
   });
 
   describe("stop", () => {
-    test("does nothing if server is already down", async () => {
+    test("sets stopping flag but does nothing else if server is already down", async () => {
       // Mock isDown to return true (server is down)
       vi.spyOn(server, "isDown").mockResolvedValue(true);
 
       await server.stop();
 
+      // stopping flag is set immediately, even if server is already down
+      expect((server as unknown as ServerInternal).stopping).toBe(true);
       expect(mockProcess.kill).not.toHaveBeenCalled();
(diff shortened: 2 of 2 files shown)


extensions/vscode/src/servers.ts
```diff
@@ -98,22 +109,35 @@ export class Server implements Disposable {
       // Do nothing if server is already down
       return;
     }
+
+    // Set stopping flag to prevent auto-restart
+    this.stopping = true;
```
dotNomad: We probably want to do this right away instead of putting it below `await this.isDown()` to help prevent timing issues.

PR #3367, opened 2025-12-30
Use `with-connect` to run e2e tests against multiple versions of Connect.

jonkeane:
Follow the same pattern we [used elsewhere](https://github.com/posit-dev/with-connect/issues/27) to test against multiple versions of Connect.

resolves #2451

PR #3492, opened 2026-02-11
fix(e2e): use Python 3.11 instead of 3.11.3 for embedded deployments

zackverham:
## Summary

- Fix Python version mismatch in embedded-deployments.cy.js tests
- Changed `version = '3.11.3'` to `version = '3.11'`

## Problem

The Connect server in the e2e environment has Python 3.11.11 installed, not 3.11.3. When the test specified the exact version `3.11.3`, the deployment would fail during environment restore with a `ModuleNotFound` error because Connect couldn't find a matching Python installation.

## Solution

Use the minor version (`3.11`) instead of the patch version (`3.11.3`), which allows Connect to match with the available Python 3.11.11 installation.

## Test plan

- [ ] Verify embedded-deployments.cy.js tests pass in CI

### DIFF
--- a/test/e2e/tests/embedded-deployments.cy.js
+++ b/test/e2e/tests/embedded-deployments.cy.js
@@ -52,7 +52,7 @@ describe("Embedded Deployments Section", () => {
         },
       )
         .then((tomlFiles) => {
-          return cy.writeTomlFile(tomlFiles.config.path, "version = '3.11.3'");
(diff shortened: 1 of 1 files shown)


karawoo:
bleh well mine is still failing the same 2 tests

dotNomad:
We (mainly @zackverham) got #3367 to pass all of the e2e tests so I suggest we just use that as the fix.

PR #3539, opened 2026-02-17
fix(e2e): fix race condition in open-connect-content test

zackverham:
## Intent

I asked Claude to take a look at why open-connect-content continues to periodically flake out. This is what it came up with. The diagnosis and fix seems reasonable - obviously its hard to prove a race condition like this, but the fix itself is pretty non-invasive and reasonable regardless of a race condition existing. It doesn't change anything about what the tests are asserting, so seems pretty safe to try.

-----

The test was flaky because it started looking for the content GUID in the explorer immediately after the quick input closed, but before VS Code had actually started reloading the workspace.

Fix: Wait for the URL to change after submitting the command, which indicates the workspace reload has actually started, before searching for the GUID in the explorer.

## Type of Change

Cypress test change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

Asked claude to diagnose and fix the flaky test

## User Impact

n/a

## Automated Tests

n/a

## Checklist

### DIFF
--- a/test/e2e/tests/open-connect-content.cy.js
+++ b/test/e2e/tests/open-connect-content.cy.js
@@ -24,32 +24,40 @@ describe("Open Connect Content", () => {
 
     cy.getPublisherTomlFilePaths("static").then((filePaths) => {
       cy.loadTomlFile(filePaths.contentRecord.path).then((contentRecord) => {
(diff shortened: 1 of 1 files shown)


karawoo (approved):
seems reasonable to me

PR #3548, opened 2026-02-18
ci: add windows-arm64 binary to published extensions

### LINKED ISSUE #3547: ci: Add `windows-arm64` binary to our published extensions
We currently publish 5 different targets to the VS Code and Open VSX marketplaces

https://github.com/posit-dev/publisher/blob/e8fd879fc372eb9ae089cf597708731c8f9bdc29/.github/workflows/publish.yaml#L21-L26

We should publish a `windows-arm64` build of Publisher as well.

We will need to build and package that executable as part of our workflows.

tdstein:
## Summary

- Adds `windows/arm64` to `scripts/get-platforms.bash` so the Go binary is compiled and the `.vsix` is packaged for that platform in CI
- Adds `windows-arm64` to the target matrices in `.github/workflows/publish.yaml` for both the VS Code Marketplace and Open VSX publish jobs

Closes #3547

### DIFF
--- a/internal/interpreters/python_test.go
+++ b/internal/interpreters/python_test.go
@@ -5,6 +5,7 @@ package interpreters
 import (
 	"errors"
 	"os"
+	"path/filepath"
 	"runtime"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


karawoo (approved):
lgtm assuming CI passes

PR #3574, opened 2026-02-20
fix: properly detect docs-only changes in CI

zackverham:
The dorny/paths-filter action doesn't support negation patterns as expected. Each pattern is evaluated independently using some() logic, so '!**/*.md' doesn't actually exclude markdown files when '**' also matches.

This fix inverts the approach: detect docs files explicitly and compare the docs file list to the all file list. If they match, only docs changed.

Also adds a test changelog entry to verify detection (delete after testing).

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Updated Go from 1.25 to 1.26. (#3554)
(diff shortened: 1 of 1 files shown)


karawoo:
ahh ok

PR #3585, opened 2026-02-20
security: add tag ancestry verification to release workflows

zackverham:
## Summary

- Add `verify-tag` job to `release.yaml` and `publish.yaml`
- Ensures tagged commits are on the `main` branch before proceeding with build/release

## Why

Without this check, someone with write access could:
1. Create a branch with unauthorized changes
2. Push a tag (e.g., `v99.0.0`) pointing to that branch
3. Trigger a release of code that was never merged to main

## How It Works

The verification step runs:
```bash
git merge-base --is-ancestor ${{ github.sha }} origin/main
```

This checks if the tagged commit is reachable from main's history:
- ✅ Tag on HEAD of main → passes
- ✅ Tag on older commit on main → passes  
- ✅ Tag on commit from merged PR → passes
- ❌ Tag on unmerged branch → fails
- ❌ Tag on arbitrary commit → fails

## Changes

| Workflow | Change |
|----------|--------|
| `release.yaml` | Added `verify-tag` job, `build` now depends on it |
| `publish.yaml` | Added `verify-tag` job, `build` now depends on it |

## Test plan

- [ ] Verify normal releases still work (tag on main)
- [ ] Verify workflow fails if tag points to non-main commit (can test manually)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -12,17 +12,7 @@ jobs:
     name: Verify tag is on main branch
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
-      - name: Verify tag points to commit on main
-        run: |
-          git fetch origin main
-          if ! git merge-base --is-ancestor ${{ github.sha }} origin/main; then
-            echo "::error::Tag must point to a commit on the main branch. This tag points to a commit not reachable from main."
-            exit 1
-          fi
-          echo "✅ Verified: Tag points to commit on main branch"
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yaml
```diff
@@ -7,7 +7,24 @@ on:
 permissions:
   contents: read
 jobs:
+  # Security: Verify the tag points to a commit on main branch
+  verify-tag:
```
dotNomad: I wonder if we should use a composite action in the repo to avoid duplication here

https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action

PR #3601, opened 2026-02-23
fix: resolve nightly test failures

zackverham:
## Summary

Fixes two issues causing nightly test failures since Feb 20.

### 1. E2E: Fix `check-positron-extension.sh` version detection

PR #3577 fixed `install-positron-extension.sh` to use git-based version detection, but forgot to update `check-positron-extension.sh`. 

**Before:** Check script looked for `posit.publisher-99.0.0` (from package.json)
**After:** Check script now correctly looks for `posit.publisher-1.35.0` (from git tag)

The logic now matches the install script:
- Exact semver tags (e.g., `1.35.0`) → use tag version
- Non-release versions (e.g., `1.34.0-7-gabcdef`) → use `99.0.0`

### 2. Unit test: Fix flaky SSE log test race condition

`TestWatchLogsWithSSEServer` was failing intermittently with:
```
FAIL: 3 out of 4 expectation(s) were met.
```

**Root cause:** The test closed the `processed` channel on the 3rd mock call (Debug), but the 4th call (runtime Info) hadn't executed yet when `AssertExpectations` ran.

**Fix:** Moved `close(processed)` to the last expected call.

## Test plan

- [x] Unit test passes 5x in a row locally
- [ ] Nightly workflow passes after merge

### DIFF
--- a/internal/clients/connect_cloud_logs/client_connect_cloud_logs_test.go
+++ b/internal/clients/connect_cloud_logs/client_connect_cloud_logs_test.go
@@ -65,12 +65,12 @@ func (s *ConnectCloudLogsClientSuite) TestWatchLogsWithSSEServer() {
 	mockLogLogger := loggingtest.NewMockLogger()
 	mockLogLogger.On("Info", "Test build info message").Once()
 	mockLogLogger.On("Error", "Test build error message").Once()
-	// For the last expected call, we'll signal the processed channel to indicate that all logs were processed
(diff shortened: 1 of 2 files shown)


zackverham:
> maybe worth extracting the expected_version logic into a shared helper? or not 😄

really, I'd rather this behavior just not be in bash - which, maybe we can do that rewrite to python pretty easily

PR #3646, opened 2026-03-04
Add connect api contract test harness

zackverham:
## Intent

Adds a test harness that validates and enforces the contract publisher has with the connect api. 

The readme provides a thorough description of the new functionality here, but the important high level bits are: 

- Introduce a test harness binary that is a light wrapper around all of the go backend interactions with the connect api. 
- Introduce a mock connect server with fixtures which are programmatically checked against the current swagger spec to ensure we don't drift. 

Runs a suite of tests that exercise the interactions across the connect api. 

We also have a stubbed typescript implementation of the same test harness - so as we begin building the typescript equivalents we can slot them in here and run the same test suite to validate the contract is still being enforced.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [x] Tooling 

## Approach

The test harness code does introduce more lines of go - but once the go code has been swapped out, the test harness can be thrown away. It isn't intended to be a permanent element in the tests. 

There's a lot of lines of code in here, but most of it is in service of setting up mocks that we can rapidly iterate against instead of relying on heavy / flaky cypress tests against a real connect installation.

## User Impact

none

## Automated Tests

adds a suite of automation that is easy to run locally but also is configured to run as part of CI. 

The swagger spec validation runs nightly as well so we will be aware of something drifts right away.

## Directions for Reviewers

Again - I know this is a lot of code. The test harness binary is written to be thrown away so don't worry about if that implementation is clean / maintainable. I mostly want help reading through the actual typescript tests for correctness. 

## Checklist

### DIFF
--- a/test/connect-api-contracts/package.json
+++ b/test/connect-api-contracts/package.json
@@ -10,6 +10,7 @@
     "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
+    "@types/node": "^22.0.0",
     "ajv": "^8.18.0",
     "ajv-formats": "^3.0.1",
     "typescript": "^5.7.0",
--- a/test/connect-api-contracts/package-lock.json
+++ b/test/connect-api-contracts/package-lock.json
@@ -6,6 +6,7 @@
     "": {
       "name": "connect-api-contracts",
       "devDependencies": {
+        "@types/node": "^22.0.0",
         "ajv": "^8.18.0",
         "ajv-formats": "^3.0.1",
         "typescript": "^5.7.0",
@@ -843,6 +844,16 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/@types/node": {
+      "version": "22.19.15",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-22.19.15.tgz",
+      "integrity": "sha512-F0R/h2+dsy5wJAUe3tAU6oqa2qbWY5TpNfL/RGmo1y38hiyO1w3x2jPtt76wmuaJI4DQnOBu21cNXQ2STIUUWg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "undici-types": "~6.21.0"
+      }
+    },
     "node_modules/@vitest/expect": {
       "version": "4.0.18",
       "resolved": "https://registry.npmjs.org/@vitest/expect/-/expect-4.0.18.tgz",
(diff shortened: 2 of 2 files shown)


test/connect-api-contracts/vitest.config.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export default defineConfig({
+  test: {
+    globalSetup: ["src/setup.ts"],
+    testTimeout: 30_000,
+    hookTimeout: 60_000,
+    include: ["src/endpoints/**/*.test.ts"],
+    // All test files share a single mock server, so they must run sequentially
+    // to avoid interfering with each other's captured requests and overrides.
+    fileParallelism: false,
```
dotNomad: Is this the only difference between the default config and `validate`?

zackverham: I think so, yeah - is that bad form?

dotNomad: No worries then 😄 definitely not necessary

justfile
```diff
@@ -234,6 +234,26 @@ test *args=("-short ./..."):
...
+build-connect-harness:
+    go build -o test/connect-api-contracts/harness/harness ./test/connect-api-contracts/harness/
+
+# Run Connect API contract tests (builds harness automatically via setup.ts)
+test-connect-contracts:
+    #!/usr/bin/env bash
+    set -eou pipefail
+    {{ _with_debug }}
+
+    cd test/connect-api-contracts && npx vitest run
```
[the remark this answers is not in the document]
dotNomad: I'm realizing I didn't explain why I asked for this change. We should use the local version of `vitest` we specify rather than the latest via `npx`

test/connect-api-contracts/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "connect-api-contracts",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "test:update": "vitest run --update",
+    "validate-fixtures": "vitest run --config vitest.validate.config.ts"
+  },
```
dotNomad: We should add a `tsc` call in here to make sure our strict type checking via our `tsconfig` is passing

something like
```suggestion
  "scripts": {
    "test": "vitest run",
    "test:watch": "vitest",
    "test:update": "vitest run --update",
    "validate-fixtures": "vitest run --config vitest.validate.config.ts",
    "typecheck": "tsc --noEmit"
  },
```

[`--noEmit`](https://www.typescriptlang.org/tsconfig/#noEmit) means we won't emit the output JS files

PR #3655, opened 2026-03-05
feat: add TypeScript TOML config loader and consolidate types

### LINKED ISSUE #3651: consolidate types
Clean up the configuration file types. There are two issues:

- TS types have not been kept up to date with Go types
- Both include fields that have never been used and are not allowed according to the (non-draft) schema

As we are writing all-new TypeScript code for reading, writing, and validating configurations, there's little value in carrying forward the unused fields that have been in draft for years. We can always add them back if we want them later.

### LINKED ISSUE #3652: load and validate config files
Add TypeScript support for loading config files into the TypeScript types, and validating against the schema.

christierney:
## Intent

Add a TypeScript-native TOML configuration loader that matches Go's config.FromFile behavior, as a building block for migrating away from the Go backend.

Fixes #3651
Fixes #3652

## Type of Change

- - [x] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

- Add smol-toml, ajv, ajv-formats dependencies for TOML parsing and JSON Schema validation
- Create src/toml/ module: loader, key converter, error factories, and JSON schema copy
- Consolidate TS types to match Go and schema (remove dead types/fields, add missing ones)
- Remove dead Go types (Schedule, AccessType, ConnectAccessControl) and unreachable code in content.go
- Fix schema: auth_type in integration_requests (was camelCase), add additionalProperties: false to integration_requests items

## User Impact

Minimal. Schema now accurately reflects what the code produces. The schema is also stricter now, so in theory if somebody had hand-edited their integration requests TOML and introduced errors, they will get better error messaging.

## Automated Tests

New tests for new code. I used the existing Go config loading tests as a guide for coverage.

## Directions for Reviewers

The type cleanup is hopefully fairly easy to follow. Unused stuff that got deleted should be easy to prove as unused since little to no code referenced it. It's a little hard to read the JSON schema manually and see that the TS types now align more closely with it, but the code for parsing and validating TOML helps prove that.

Nothing is calling the TS toml code yet (other than tests) but I tried to design it to be a clean building block for replacing the APIs, and there are some "downstream compatibility" tests to help demo that.

A note on the schema change: this could be considered breaking, since I'm renaming an existing field, but I think it's ok for a couple reasons:
- the extension code actually uses the snake_case field name, not the camelCase the schema referenced. So nobody would have the camelCase version unless they wrote it by hand. This change brings the schema in line with reality.
- the integration requests type was not doing strict field checking prior to this change, so we could argue that the type was not fully locked down until now

The schema now lives in two places (under the Go code and under TS). I think this is acceptable for now as we're in flux. We should clean this up as we get rid of the Go side.

## Checklist

Mentioned the schema change since that is potentially user-visible. The rest is all internal changes.

### DIFF
--- a/extensions/vscode/src/toml/loader.ts
+++ b/extensions/vscode/src/toml/loader.ts
@@ -48,9 +48,9 @@ export async function loadConfigFromFile(
   const content = await fs.readFile(configPath, "utf-8");
 
   // Parse TOML
-  let parsed: Record<string, unknown>;
+  let parsed;
   try {
-    parsed = parseTOML(content) as Record<string, unknown>;
+    parsed = parseTOML(content);
   } catch (err: unknown) {
     if (err instanceof TomlError) {
       const line = err.line ?? 0;


extensions/vscode/src/toml/schema.test.ts
```diff
@@ -0,0 +1,253 @@
+// Copyright (C) 2026 by Posit Software, PBC.
```
christierney: this is based on a similar test on the Go side---it's basically checking that we have defined our schema correctly. (It's probably not comprehensive).

extensions/vscode/src/toml/convertKeys.test.ts
```diff
@@ -0,0 +1,135 @@
...
+            some_custom_key: "value",
+            ANOTHER_KEY: "val2",
+          },
+        },
+      ],
+    });
+  });
+
+  it("passes through already-camelCase keys unchanged", () => {
+    const result = convertKeysToCamelCase({
```
zackverham: does this test actually validate camelCase keys?

christierney: it's validating that keys which are already camelCase (`authType`, `loadFactor`, `defaultImageName`) get passed through unchanged

extensions/vscode/src/toml/convertKeys.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      integration_requests: [{ auth_type: "oauth", guid: "abc-123" }],
+    });
+    expect(result).toEqual({
+      integrationRequests: [{ authType: "oauth", guid: "abc-123" }],
+    });
+  });
+
+  it("preserves environment keys (user-defined)", () => {
+    const result = convertKeysToCamelCase({
```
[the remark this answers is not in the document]
zackverham: I got mixed up on keys vs values, disregard!

dotNomad (approved):
Thanks for that change 🎉

PR #3672, opened 2026-03-09
Migrate /api/interpreters endpoint from Go to TypeScript

zackverham:
## Summary

Move interpreter detection (Python/R version, lockfiles, package managers) from the Go backend API into a new TypeScript module in the VSCode extension. This eliminates the HTTP round-trip to the Go process for logic that can run directly in TypeScript using `child_process`.

- Add `src/interpreters/` module with Python and R detection, version constraint parsing, and project metadata reading (`.python-version`, `pyproject.toml`, `setup.cfg`, `DESCRIPTION`, `renv.lock`)
- Replace `api.interpreters.get()` calls in `state.ts` with `getInterpreterDefaults()`
- Remove `GET /api/interpreters` Go endpoint and route registration
- Remove `Interpreters` API resource class from TypeScript client
- Add 74 unit tests covering all new modules

## How requiresPython / requiresR worked in the Go implementation

In the Go backend, the `Python` and `R` config structs in `internal/config/types.go` each had a `FillDefaults()` method that was called with an interpreter instance. The flow was:

1. **`Python.FillDefaults(pythonInterpreter)`** — if `RequiresPythonVersion` was empty, called `python.GetPythonRequires()`, which searched `.python-version` → `pyproject.toml` → `setup.cfg` in order.
2. **`R.FillDefaults(rInterpreter)`** — if `RequiresRVersion` was empty, called `rLang.GetRRequires()`, which searched `DESCRIPTION` → `renv.lock` in order.

The key difference is that in Go, `FillDefaults` was called on the config struct itself — it mutated the config in place. The interpreters were Go objects already constructed with the project directory and executable path, and `GetPythonRequires()` / `GetRRequires()` were methods on those interpreter objects.

In the TypeScript migration, this is now handled in two places:

- **`detectPythonInterpreter` / `detectRInterpreter`** — call `getPythonRequires()` / `getRRequires()` and include the result in the returned config.
- **`UpdateConfigWithDefaults`** in `configurations.ts` — fills in empty fields on the config, same as `FillDefaults` did in Go.

## Go components intentionally retained

The following Go interpreter code is **not removed** by this PR because it is still used by other backend endpoints:

| Component | Used by |
|-----------|---------|
| `InterpretersFromRequest()` in `api_helpers.go` | `POST /api/deployment` (post_deployment.go), `POST /api/inspect` (post_inspect.go), `POST /api/packages/r/scan` (post_packages_r_scan.go) |
| `internal/interpreters/` package (Python/R interpreter structs, version detection, requires parsing) | `InterpretersFromRequest`, `config.Python.FillDefaults()`, `config.R.FillDefaults()`, bundle manifest generation |
| `config.Python.FillDefaults()` / `config.R.FillDefaults()` in `internal/config/types.go` | Server-side config population during deployment and inspection |

These can be migrated in follow-up work as their consuming endpoints are moved to TypeScript.

## Type of Change

- [x] Refactor

## Approach

Ported the interpreter detection logic from Go to TypeScript, matching the existing behavior:
- Python version via `python -E -c "import sys; ..."` with pyenv shim cache bypass
- R version via `R --version` with combined stdout/stderr parsing
- renv lockfile resolution via `renv::paths$lockfile()` with fallback to default `renv.lock`
- Version constraint adaptation (PEP 440 operators, compatible release `~=`)
- Project metadata parsing for Python requires (`.python-version`, `pyproject.toml`, `setup.cfg`) and R requires (`DESCRIPTION`, `renv.lock`)

## User Impact

No user-facing behavior change. Interpreter detection now runs in-process in the extension instead of making an HTTP call to the Go backend.

## Automated Tests

74 new Vitest unit tests across 6 test files covering:
- Version constraint adaptation (28 tests)
- Python requires parsing (10 tests)
- R requires parsing (10 tests)
- Python interpreter detection (9 tests)
- R interpreter detection (11 tests)
- `getInterpreterDefaults()` public API (6 tests)

32 integration tests in `integration.test.ts` that run against:
- A real temporary filesystem (created with `mkdtemp`, cleaned up after)
- Real Python and R executables when available on PATH

Tests are organized in layers from low-level to high-level:
1. `fsUtils` — filesystem helpers
2. `getPythonRequires` / `getRRequires` — version constraint extraction from project files
3. `detectPythonInterpreter` / `detectRInterpreter` — full detection with real executables
4. End-to-end tests — detection + project files combined
5. `getInterpreterDefaults` — top-level orchestrator for both interpreters

Tests that require a specific interpreter use `test.skipIf(!available)` so the suite runs in any environment. In CI, `interpreter-integration.yaml` installs known Python and R versions across a 12-configuration matrix (Python 3.9–3.13, R 4.1–4.4, combined, cross-platform).

### E2E coverage

No new E2E tests are needed. The existing deployment E2E tests (`deployments.cy.js`, `embedded-deployments.cy.js`) implicitly validate this migration — they open real Python and R projects, wait for interpreter detection via `waitForInterpreterReady()`, and deploy to Connect. If interpreter detection regressed (wrong version, missing lockfile, bad package manager), those deployments would fail.

This migration changes internal plumbing (HTTP call to Go → direct `child_process` execution), not the inputs or outputs. The sidebar displays the same data and configurations receive the same defaults, so the existing E2E suite covers the end-to-end behavior without modification.

## Checklist

### DIFF
--- a/extensions/vscode/src/interpreters/integration.test.ts
+++ b/extensions/vscode/src/interpreters/integration.test.ts
@@ -1,7 +1,32 @@
 // Copyright (C) 2026 by Posit Software, PBC.
 //
-// Integration tests that run against a real filesystem and real interpreters.
-// Interpreter-specific tests are skipped when the executable is not available.
+// Integration tests for the interpreter detection pipeline.
+//
+// Unlike the unit tests (e.g. pythonRequires.test.ts, rInterpreter.test.ts)
+// which mock the filesystem and child processes, these tests exercise the
+// real code paths against:
+//   - A real temporary filesystem (created with mkdtemp, cleaned up after)
+//   - Real Python and R executables when available on the PATH
+//
+// Tests that require a specific interpreter use `test.skipIf(!available)` so
+// the suite can run in any environment without failures. In CI, the
+// interpreter-integration.yaml workflow installs known Python and R versions
+// across a matrix of configurations to ensure full coverage.
+//
+// The tests are organized in layers from low-level to high-level:
+//   1. fsUtils          – filesystem helpers (readFileText, fileExistsAt)
+//   2. getPythonRequires – version constraint extraction from project files
+//   3. getRRequires      – version constraint extraction from R project files
+//   4. detectPythonInterpreter – full Python detection (exec + project files)
+//   5. detectRInterpreter      – full R detection (exec + project files)
+//   6. End-to-end tests  – detection + project files combined
+//   7. getInterpreterDefaults  – top-level orchestrator for both interpreters
+//
+// Run these tests with:
+//   npx vitest run src/interpreters/integration.test.ts
+//
+// Or as part of the dedicated npm script:
+//   npm run test-integration-interpreters
 
 import { execFile } from "child_process";
 import { mkdtemp, rm, writeFile } from "node:fs/promises";
@@ -21,6 +46,9 @@ import { detectRInterpreter } from "./rInterpreter";
 
 const execFileAsync = promisify(execFile);
 
+// Shared temp directory for simple tests that don't need isolation.
+// Tests that verify priority ordering or need a clean project directory
+// create their own temp dirs to avoid file leakage between tests.
 let tmpDir: string;
 
 beforeAll(async () => {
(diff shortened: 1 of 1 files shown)


extensions/vscode/src/interpreters/rRequires.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      setFile(
+        "/project",
+        "DESCRIPTION",
+        "Depends: package1, package2, package3",
+      );
+      const result = await getRRequires("/project");
+      expect(result).toBe("");
+    });
+
+    test("does not match tinyR", async () => {
```
zackverham: is this correct? Do we not want to match tinyR?

PR #3673, opened 2026-03-09
feat: implement TypeScript Connect API client

zackverham:
## Intent

Implement a TypeScript Connect API client package (`@posit-dev/connect-api`) and validate it against contract tests alongside the existing Go backend.

## Type of Change

- [x] New Feature
- [x] Refactoring

## Approach

Created a standalone `packages/connect-api/` package with a `ConnectAPI` class that wraps the Posit Connect REST API. All 15 API methods are implemented:

- **Auth & User**: `testAuthentication`, `getCurrentUser`
- **Content CRUD**: `contentDetails`, `createDeployment`, `updateDeployment`
- **Environment**: `getEnvVars`, `setEnvVars`
- **Bundles**: `uploadBundle`, `latestBundleId`, `downloadBundle`
- **Deployment**: `deployBundle`, `waitForTask`, `validateDeployment`
- **Server Info**: `getIntegrations`, `getSettings`

Key design decisions:

- **axios** for HTTP — uses `axios.create()` with a pre-configured `baseURL` and default `Authorization` header. Each method calls `this.client.request<T>()` directly with response generics for strong typing.
- **Default axios error handling** — non-2xx responses throw `AxiosError` automatically. No custom error classes. Two methods use per-request `validateStatus` overrides: `testAuthentication` (inspects error response bodies for messages) and `validateDeployment` (accepts non-5xx status codes).
- **Raw API responses** — every method returns the exact JSON the Connect API sends back (full DTOs), rather than constructing new objects or extracting single fields. The contract test adapter handles any field extraction needed.
- **Branded ID types** (`ContentID`, `BundleID`, `TaskID`) for type safety on parameters

## User Impact

None — this is preparatory code.

## Automated Tests

- 43 unit tests in `packages/connect-api/` (vitest) covering all methods, error paths, and URL construction
- All 68 contract tests pass with both backends (`npx vitest run` and `API_BACKEND=typescript npx vitest run`)
- CI runs contract tests against both Go and TypeScript backends

### DIFF
--- a/test/connect-api-contracts/src/clients/ts-direct-client.ts
+++ b/test/connect-api-contracts/src/clients/ts-direct-client.ts
@@ -20,6 +20,12 @@ import type { CapturedRequest } from "../mock-connect-server";
  * Thin adapter that wraps the production ConnectAPI for contract testing.
  * Handles mock server request capture and maps return values to the
  * contract result shapes ({ contentId }, { bundleId }, etc.).
+ *
+ * NOTE: The dispatch() method manually reshapes responses (e.g. extracting
+ * `data.guid` as `contentId`, `data.id` as `bundleId`) to produce output
+ * that matches the Go client's contract results for apples-to-apples
+ * comparison. Once the Go implementation is migrated to use the TypeScript
+ * client, this adapter can be reorganized to remove the reshaping layer.
  */
 export class TypeScriptDirectClient implements ConnectContractClient {
   private readonly connectClient: ConnectAPI;
@@ -45,7 +51,7 @@ export class TypeScriptDirectClient implements ConnectContractClient {
       result = await this.dispatch(method, params ?? {});
     } catch (err) {
       status = "error";
-      // For testAuthentication, the error carries a structured result
+      // Reshape error into { user, error } to match Go contract shape
       if (method === Method.TestAuthentication && err instanceof Error) {
         const msg = err.message;
         result = { user: null, error: { msg } };
@@ -77,10 +83,12 @@ export class TypeScriptDirectClient implements ConnectContractClient {
         return c.getCurrentUser();
 
       case Method.ContentDetails:
+        // Unwrap AxiosResponse to match Go contract shape
         return (await c.contentDetails(ContentID(params.contentId as string)))
           .data;
 
       case Method.CreateDeployment: {
+        // Map guid → contentId to match Go contract shape
         const { data } = await c.createDeployment(
           (params.body as Record<string, unknown>) ?? {},
         );
@@ -95,6 +103,7 @@ export class TypeScriptDirectClient implements ConnectContractClient {
         return undefined;
 
       case Method.GetEnvVars:
+        // Unwrap AxiosResponse to match Go contract shape
         return (await c.getEnvVars(ContentID(params.contentId as string))).data;
 
       case Method.SetEnvVars:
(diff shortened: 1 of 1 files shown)


packages/connect-client/src/client.ts
```diff
@@ -0,0 +1,356 @@
...
+  constructor(options: ConnectClientOptions) {
+    this.url = options.url;
+    this.apiKey = options.apiKey;
+  }
+
+  // ---------------------------------------------------------------------------
+  // HTTP helpers
+  // ---------------------------------------------------------------------------
+
+  private async request(
```
dotNomad: I would prefer to keep using `axios` despite some of the problems we've had so we don't have to write our own error handling logic, or change how we do it in a lot of spots. That would simplify what we have here quite a bit.

packages/connect-client/src/client.ts
```diff
@@ -0,0 +1,356 @@
...
+  /** Retrieves the current authenticated user without validation checks. */
+  async getCurrentUser(): Promise<User> {
+    const dto = await this.requestJson<UserDTO>("GET", "/__api__/v1/user");
+    return {
+      id: dto.guid,
+      username: dto.username,
+      first_name: dto.first_name,
+      last_name: dto.last_name,
+      email: dto.email,
+    };
```
dotNomad: This is a preference in design, but in my opinion it has a lot of benefits:

I would prefer to return the exact thing the API call responds with, including all data, and not creating new objects.

This will keep this code _extremely_ single use and the only thing we will need to adjust overtime are the types. We won't have to adjust the functions since they just pass the response right through. We can always extract the data from the response.

It is much more flexible.

You can read more in our README.md for the Go API client we wrote: https://github.com/posit-dev/publisher/blob/183abd7d744fb8f6d8bcffc323717547521eca57/extensions/vscode/src/api/README.md

packages/connect-api/src/client.ts
```diff
@@ -0,0 +1,355 @@
...
+export class ConnectAPI {
+  private readonly client: AxiosInstance;
+
+  constructor(options: ConnectAPIOptions) {
+    this.client = axios.create({
+      baseURL: options.url,
+      headers: {
+        Authorization: `Key ${options.apiKey}`,
+      },
+      validateStatus: () => true,
```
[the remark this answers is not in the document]
zackverham: My understanding is that this would let the caller interpret the HTTP code - which is how I'd expect to handle it in go, but if its better to throw I'm happy to do that.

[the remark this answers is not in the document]
dotNomad: It's the normal `axios` behavior to throw errors for non-200 error codes. If I was using an `axios` client I would definitely be surprised by that behavior.

packages/connect-api/src/client.test.ts
```diff
@@ -0,0 +1,902 @@
...
+    expect(call.method).toBe("GET");
+  });
+});
+
+// ---------------------------------------------------------------------------
+// contentDetails
+// ---------------------------------------------------------------------------
+
+describe("contentDetails", () => {
+  const contentId = "content-guid-abc" as ContentID;
```
dotNomad: I think we can make a factory function to assist with reducing the number of type coercions we use in here. Very minor nit though.

dotNomad (commented):
I'm leaving comments to go off and do another review, but I'll do some work in the background on this and push up changes based on my comments for you to take a look at tomorrow morning @zackverham

PR #3677, opened 2026-03-09
Fix flaky open-connect-content e2e test

zackverham:
## Summary

Fixes the flaky `open-connect-content` e2e test that was failing across multiple Connect versions (preview, release, 2023.03.0, 2025.03.0).

### Root cause

The `ConnectContentFileSystemProvider` was **blocking VS Code's workspace validation** during the workspace switch. When `updateWorkspaceFolders` is called, code-server validates the new folder by calling `stat()` and `readDirectory()` on the root URI. Our implementation awaited the full bundle fetch pipeline (Go backend startup → credential refresh → Connect API call → tar extraction) before responding. This caused code-server to silently reject the workspace change because the validation timed out.

Additionally, in code-server, `updateWorkspaceFolders` creates a **multi-root workspace** where the content folder appears at `aria-level="2"` instead of `aria-level="1"`, so the test's CSS selector was missing the GUID entirely.

### Extension fixes (`connect_content_fs.ts`)

- **Non-blocking `stat()` for root URIs**: Returns an immediate directory entry without waiting for the bundle fetch. The bundle is kicked off in the background.
- **Non-blocking `readDirectory()` for root URIs**: Returns whatever is cached (or `[]` if nothing yet), then fires a `FileChangeType.Changed` event when the bundle fetch completes to trigger VS Code to refresh the explorer tree.
- **Non-blocking error handling**: `showErrorMessage` and the credential dialog are fire-and-forget (`void`) instead of `await`ed, so they don't stall the filesystem provider.

### Test fixes (`open-connect-content.cy.js`)

- Removed `aria-level='1'` restriction from the GUID selector — matches the folder at any tree depth to handle both single-folder and multi-root workspace modes.
- Added notification toast dismissal during polling.
- Increased `waitUntil` timeout to 120s.
- Added explorer DOM diagnostics to the error message for easier debugging of future failures.

## Test plan

- [x] All 4 e2e matrix variants pass (preview, release, 2023.03.0, 2025.03.0)
- [x] Unit tests pass (18/18) including new error-handling coverage
- [x] Lint and formatting clean

### DIFF
--- a/extensions/vscode/src/connect_content_fs.test.ts
+++ b/extensions/vscode/src/connect_content_fs.test.ts
@@ -317,4 +317,26 @@ describe("ConnectContentFileSystemProvider", () => {
       expect(mockOpenConnectContent).toHaveBeenCalledTimes(1);
     });
   });
+
+  describe("error handling", () => {
+    test("stat resolves with empty directory when bundle fetch fails", async () => {
+      mockOpenConnectContent.mockRejectedValue(new Error("network error"));
+
+      const stat = await provider.stat(
+        makeUri(testAuthority, `/${testContentGuid}`),
+      );
+
+      expect(stat.type).toBe(DIRECTORY);
+    });
+
+    test("readDirectory returns empty list when bundle fetch fails", async () => {
+      mockOpenConnectContent.mockRejectedValue(new Error("network error"));
+
+      const entries = await provider.readDirectory(
+        makeUri(testAuthority, `/${testContentGuid}`),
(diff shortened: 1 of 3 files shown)


dotNomad (changes requested):
Caught a bug with this. Detailed in an inline comment.

zackverham:
@christierney @dotNomad I haven't been noticing this test flaking out as much as it used to - I'm considering closing this as a red herring.

PR #3678, opened 2026-03-10
Remove unused util.Map function

zackverham:
## Summary
- Deletes `internal/util/slice.go` which contains a single unused generic `Map` function
- The function has zero callers in the codebase (the test that previously used it was removed)
- Go's native range loops and standard library cover this use case

## Test plan
- [x] `go build ./...` passes with the file removed
- No test changes needed — there were no tests for this function


dotNomad (approved):
✨

PR #3804, opened 2026-03-18
fix: use execFile instead of runTerminalCommand for quarto availability check

### LINKED ISSUE #3801: isQuartoBinAvailable() should not open a visible terminal
Tracked in #3713.

## Problem

`isQuartoBinAvailable()` in `quartoProjectHelper.ts` uses `runTerminalCommand()` to check if Quarto is installed by running `quarto --version`. This creates a visible VSCode terminal, and if Quarto is not installed (or the command fails), the terminal exits with code 1, causing VSCode to show a generic error:

> The terminal process "/bin/zsh '-l'" terminated with exit code: 1.

This is confusing for users since it doesn't indicate what went wrong.

## Suggested Fix

Replace the `runTerminalCommand("quarto --version")` call with `child_process.exec` (or similar) so the check runs silently in the background without opening a visible terminal.

zackverham:
## Intent

`isQuartoBinAvailable()` was using `runTerminalCommand()` which opens a visible VSCode terminal. When quarto is not installed, this causes a confusing error message about the terminal process exiting with code 1. Replace with `child_process.execFile` to run the check silently in the background.

Closes #3801

## Type of Change

- [x] Bug Fix

## Approach

Replaced the `runTerminalCommand("quarto --version")` call in `isQuartoBinAvailable()` with `child_process.execFile("quarto", ["--version"], callback)`. This runs the quarto check as a background child process instead of opening a visible terminal tab, eliminating the confusing "exited with code 1" error when quarto is not installed.

Updated unit tests and smoke tests to mock `execFile` instead of intercepting the `runTerminalCommand` call for the version check.

## User Impact

Users without Quarto installed will no longer see a terminal flash open with an error message when the extension checks for quarto availability. The check now happens silently in the background.

## Automated Tests

- Updated `quartoProjectHelper.test.ts` to mock `child_process.execFile` for the quarto bin check
- Updated `quartoProjectHelper.smoke.test.ts` to mock `execFile` while preserving real filesystem usage for other smoke tests

## Directions for Reviewers

1. Verify no terminal opens when quarto is not installed
2. Verify quarto projects still render correctly when quarto IS installed
3. See smoke test plan in PR comments

## Checklist


zackverham:
## Smoke Test Plan

### Test Cases

#### 1. Quarto NOT installed — no spurious terminal error (primary fix)
1. Ensure `quarto` is not on PATH
2. Open VSCode with the extension installed
3. Open a project that contains a `.qmd` file
4. Trigger a deployment via the Publisher deploy flow

**Expected:** No terminal opens. No "process exited with code 1" error message appears. The extension gracefully treats quarto as unavailable.

#### 2. Quarto installed — project renders correctly
1. Ensure `quarto` is on PATH
2. Open a Quarto project (e.g., one with `_quarto.yml` and `.qmd` files)
3. Trigger a deployment that invokes Quarto rendering (via the Publisher deploy flow)

**Expected:** Quarto is detected as available. The render command executes in the terminal as before. Deployment succeeds.

extensions/vscode/src/utils/quartoProjectHelper.ts
```diff
@@ -55,13 +56,12 @@ export class QuartoProjectHelper {
...
-    } catch {
-      return Promise.resolve(false);
-    }
+  isQuartoBinAvailable(): Promise<boolean> {
+    return new Promise((resolve) => {
+      execFile("quarto", ["--version"], (error) => {
+        resolve(!error);
+      });
+    });
   }
```
dotNomad: We can adjust this to use `async` `await`. I find it a bit easier to read, but its totally optional.


```suggestion
  async isQuartoBinAvailable(): Promise<boolean> {
    try {
      await execFileAsync("quarto", ["--version"]);
      return true;
    } catch {
      return false;
    }
  }
```

It would also require us to use the `promisify` `util` with an additional import up top

```
import { execFile } from "child_process";
import { promisify } from "util";
import { fileExistsAt } from "../interpreters/fsUtils";
import { runTerminalCommand } from "./window";

const execFileAsync = promisify(execFile);
```

dotNomad (approved):
A suggestion on syntax, but its totally optional - I just find it a bit easier to read.

zackverham:
Tested by removing quarto and trying to render - I now just see an informative error, not the shell command.

PR #3836, opened 2026-03-20
fix: add TLS certificate verification support to ConnectAPI client

### LINKED ISSUE #3835: ConnectAPI client does not support TLS certificate verification settings
The `ConnectAPI` class from `@posit-dev/connect-api` creates an axios instance without any TLS configuration. The `ConnectAPIOptions` type only accepts `{url, apiKey}`.

The extension has a `verifyCertificates()` setting that controls whether TLS certificates are verified. The Go backend respects this, but the TypeScript `ConnectAPI` client does not — there's no way to pass an `httpsAgent` with `rejectUnauthorized: false`.

This affects all existing uses of `new ConnectAPI()` in the extension (integration requests, etc.), not just the new TS publish path. Users connecting to Connect servers with self-signed certificates will get TLS errors from the TS code paths.

Fix: Add an optional `httpsAgent` or `insecure` option to `ConnectAPIOptions` and pass it through to `axios.create()`.

Related: #3793

m--:
## Summary

- Add `rejectUnauthorized` option to `ConnectAPIOptions` in `@posit-dev/connect-api`
- Pass extension's `verifyCertificates` setting to all `ConnectAPI` instantiations
- Users with self-signed certificates can now disable TLS verification in TypeScript code paths

## Test plan

- [x] Unit tests added for `rejectUnauthorized` option (3 new tests)
- [ ] Manual test with self-signed certificate Connect server

Fixes #3835

### DIFF
--- a/extensions/vscode/src/connect_content_fs.ts
+++ b/extensions/vscode/src/connect_content_fs.ts
@@ -22,7 +22,6 @@ import { Commands } from "src/constants";
 import { PublisherState } from "./state";
 import { ConnectAPI, ContentID, BundleID } from "@posit-dev/connect-api";
 import { Credential } from "./api/types/credentials";
-import { extensionSettings } from "./extension";
 
 type ConnectContentEntry = {
   type: FileType;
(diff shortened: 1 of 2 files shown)


extensions/vscode/src/connect_content_fs.ts
```diff
@@ -220,9 +220,14 @@ export class ConnectContentFileSystemProvider implements FileSystemProvider {
       normalizedServerUrl = await this.ensureCredentialsForServer(serverUrl);
       const credential =
         await this.findCredentialForServer(normalizedServerUrl);
+      const verifyCerts =
+        workspace
+          .getConfiguration("positPublisher")
+          .get<boolean>("verifyCertificates") ?? true;
```
m--: Originally we had a different approach here, but ran into import problems and so Claude went this way. I'm not sure if this is what we should be doing for this kind of thing, but if it is, any thoughts about giving Claude more specific instructions re: imports in the extension?

christierney (approved):
lgtm

Issue #3833, opened 2026-03-19
TS publish path: add cancellation support via AbortController


christierney:
The Go publish path supports cancellation via the `cancelDeployment()` API endpoint. The new TypeScript publish orchestrator (`connectPublish()`) does not yet support cancellation.

To add this:
- Thread an `AbortController` signal through `connectPublish()` and into each API call
- Wire it to the VSCode progress notification's cancellation token
- Handle cleanup (partial deployment records) on abort

Deferred from the initial wiring PR to keep scope manageable.

Related: #3793

PR #3839, opened 2026-03-20
Wire TS publish orchestrator into VSCode extension

### LINKED ISSUE #3794: Add event emission / progress reporting for TypeScript publish flow
## Context

Part of #3788 (migrate publish flow). Depends on #3793.

## What

The Go publish flow emits SSE events at each phase (start, success, failure, log) so the extension UI can show progress. The TypeScript publish orchestration needs equivalent progress reporting.

## Options

1. **Emit SSE events from TS** — match the Go event format so existing webview consumers work unchanged
2. **Use VS Code progress API** — simpler implementation, but different UX and would require webview changes
3. **Keep SSE pattern** — maintain consistency with existing webview event consumers

## Where to look

- Go event emission: \`internal/events/\` (event types and operations)
- Go publish events: grep for \`PublishOp\`, \`PublishCreateBundleOp\`, etc. in \`internal/publish/\`
- TS event consumption: \`src/events.ts\` (SSE client)
- Webview event handling: \`webviews/homeView/\`

### LINKED ISSUE #3834: TS publish path: stream detailed deployment logs like Go SSE path
The Go publish path streams fine-grained deployment events via SSE — individual file names during bundling, pip/renv output during environment restore, and server-side task output. These appear in the Posit Publishing Log view and in the progress notification.

The TypeScript publish orchestrator currently only reports step-level progress (e.g., "Creating bundle...", "Uploading..."). It does not stream:
- File lists during bundle creation
- R package resolution output
- Server-side task output lines (though `waitForTask` now supports an `onOutput` callback that can be used for this)

Step-level progress is sufficient for the initial release, but parity with the Go path would improve the user experience.

Related: #3793

### LINKED ISSUE #3838: wire rejectUnauthorized through TS code
When #3835 is closed, we need to update the users of connect-api to pass the option along correctly. #3836 already does some of this but currently in-flight work might still need updates.

### LINKED ISSUE #3925: TS publish: preflight capability/settings validation (requires connect-api getSettings)
The TS publish orchestrator (`connectPublish.ts`) is missing several preflight validation checks that the Go path performs in `CheckCapabilities()` (`internal/clients/connect/capabilities.go:62-80`). These checks require calling `GET /__api__/server_settings` which is not yet available in the TS connect-api client.

## Missing checks (require `getSettings()` in connect-api)

1. **Content type licensing** — verify the server is licensed for the content's app mode (API, Shiny, etc.) via `applications.enabled`, `applications.shiny_enabled`, etc.
2. **RACU (Run-As-Current-User)** — check `applications.run_as_current_user` licensing and configuration
3. **Kubernetes licensing and configuration** — check `applications.kubernetes_enabled`, image selection settings
4. **Runtime settings not applied to static content** — warn if Python/R runtime settings are set for static content types
5. **Python/R/Quarto version availability** — verify the configured runtime versions are available on the server

## Prerequisite

Add a `getSettings()` method to `packages/connect-api/src/client.ts` that fetches `GET /__api__/server_settings` and returns the server settings object. Define a `ServerSettingsDTO` type covering the fields needed for the checks above.

## Checks implemented without connect-api changes

The following checks have been implemented directly in the orchestrator (no server settings needed):
- Requirements file existence validation
- Description length <= 4096 characters

## Context

- Go source of truth: `internal/clients/connect/capabilities.go` and `internal/clients/connect/settings.go`
- TS orchestrator: `extensions/vscode/src/publish/connectPublish.ts` (preflight step)
- Gap analysis: `TS_PUBLISH_GAP_ANALYSIS.md` Gap 7

christierney:
## Summary

Wires the TypeScript publish orchestrator (from #3828) into the VSCode extension so that plain Connect deployments use the native TS path instead of the Go backend. Includes extensive parity work to match Go's behavior across event streaming, preflight validation, error classification, and deployment records.

Closes #3794. Closes #3834. Closes #3925. Fixes #3838.

### Routing

- `canUseTsPublishPath()` gates on server type and config — Connect Cloud, Snowflake, and `packagesFromLibrary` fall back to Go

### Progress & Log Streaming

- `tsDeployProgress.ts` wraps `window.withProgress()` with step labels matching the Go UX
- Injects synthetic `EventStreamMessage` into the `EventStream` via `injectMessage()`, so the logs tree view and raw log viewer work identically for TS and Go paths
- Server task output lines from `waitForTask` emitted individually as `publish/restoreEnv/log` events
- `restoreEnv` → `runContent` stage transition detected via launch/static patterns (mirrors Go's `eventOpFromLogLine`)
- R/Python package installation progress detected via regex (mirrors Go's `packageEventFromLogLine`)
- R package resolution log messages: lockfile loading, auto-scan, dependency detection

### Preflight Validation

- `checkServerSettings()` mirrors Go's `CheckCapabilities`: API licensing, RACU, runtime version/availability, Kubernetes image selection, content type licensing
- Requirements file inclusion check matches Go (fails even with empty files list)
- Description length, locked content, app mode mismatch checks
- 404/403-specific messages for missing/forbidden content (mirrors Go's `preflightAgentError`)

### Error Handling

- Error codes match Go exactly: `authFailedErr`, `deploymentNotFoundErr`, `appModeNotModifiableErr`, `deployFailed`, `requirementsFileReadingError`
- `errCode` propagated from step failures to top-level `publish/failure` event
- `deployedAt` set on every deployment record write (initial, post-upload, error) matching Go
- Non-JSON auth response guard in `testAuthentication()` for auth-proxy scenarios
- `getSettings()` scheduler path guarded with known app mode allowlist

### Lifecycle

- `publish/start|success|failure` stream events are sole authority — HomeView's existing handlers pick them up with no duplicate callbacks
- Success notification is fire-and-forget so progress bar closes immediately

### Known gaps (issues filed)

- Cancellation: #3833

## Test plan

- [x] Deploy Python app to Connect (first deploy) — verify all tree stages, progress steps, success notification with "View" button
- [x] Redeploy same app — verify updateContent path, preserved createdAt, correct tree stages
- [x] Deploy Quarto doc — verify no restoreEnv, runContent stage appears
- [x] Deploy R Shiny app with renv.lock — verify R package log messages, renv in deployment record
- [x] Deploy with env vars (regular + secrets) — verify not logged but set
- [x] Deploy with bad API key — verify auth error in tree, deployment record error
- [x] Deploy to deleted content — verify "Content cannot be found" preflight error
- [x] Deploy with broken dependency — verify waitForTask failure, "View Connect Logs" link
- [ ] Verify Connect Cloud deploy still uses Go path
- [x] Run `just test` from `extensions/vscode/` — all unit tests pass

### DIFF
--- a/extensions/vscode/src/views/tsDeployProgress.test.ts
+++ b/extensions/vscode/src/views/tsDeployProgress.test.ts
@@ -497,6 +497,11 @@ describe("runTsDeployWithProgress", () => {
       });
       onProgress({
         step: "createNewDeployment",
+        status: "log",
+        message: "Creating new deployment",
+      });
+      onProgress({
+        step: "createNewDeployment",
         status: "success",
         data: { contentId: "abc-123", saveName: "my-app" },
       });
@@ -512,6 +517,12 @@ describe("runTsDeployWithProgress", () => {
     );
     expect(startMsg).toBeDefined();
     expect(startMsg!.data.saveName).toBe("my-app");
+
+    const logMsg = stream.injected.find(
+      (m) => m.type === "publish/createNewDeployment/log",
+    );
+    expect(logMsg).toBeDefined();
+    expect(logMsg!.data.message).toBe("Creating new deployment");
 
     const successMsg = stream.injected.find(
       (m) => m.type === "publish/createNewDeployment/success",


christierney:
**R package metadata key names in deployment records**

The renv package sections in deployment TOML files (`[renv.packages.*]`) use slightly different key names between the Go and TS paths:

Go path:
```toml
[renv.packages.R6]
package = 'R6'
version = '2.6.1'
source = 'Repository'
bugreports = 'https://github.com/r-lib/R6/issues'
roxygennote = '7.3.2'
needscompilation = 'no'
```

TS path:
```toml
[renv.packages.R6]
package = "R6"
version = "2.6.1"
source = "Repository"
bug_reports = "https://github.com/r-lib/R6/issues"
roxygen_note = "7.3.2"
needs_compilation = "no"
```

Go uses explicit TOML struct tags that lowercase the DESCRIPTION field names without separators. TS applies a generic PascalCase-to-snake_case conversion on the renv.lock keys.

This is cosmetic — these keys are write-only record-keeping in the deployment TOML. They are never read back by the extension or sent to Connect (the manifest uses PascalCase JSON). Matching Go exactly would require special-casing these DESCRIPTION field names, which isn't worth the complexity for unused fields.

extensions/vscode/src/views/canUseTsPublishPath.test.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ProductType } from "src/api/types/contentRecords";
+import type { ConfigurationDetails } from "src/api/types/configurations";
+
+function makeConfig(
+  overrides?: Partial<ConfigurationDetails>,
+): ConfigurationDetails {
+  return {
+    $schema: "",
+    productType: ProductType.CONNECT,
+    type: "python-dash" as ConfigurationDetails["type"],
```
dotNomad: We can use the `ContentType` `enum` directly to avoid the `as` here.

```suggestion
    type: ContentType.PYTHON_DASH,
```

extensions/vscode/src/views/homeView.ts
```diff
@@ -307,6 +311,114 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
     configurationName: string,
     projectDir: string,
     secrets?: Record<string, string>,
+  ) {
+    const credential = this.state.findCredential(credentialName);
```
dotNomad: I wonder if we should access the Credential via the `CredentialService` directly rather than relying on the `PublisherState.credentials` in the future. It would remove a bit of syncing and message listening.

extensions/vscode/src/views/homeView.ts
```diff
@@ -307,6 +311,114 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+        connectPublish({
+          api: connectApi,
+          projectDir: absProjectDir,
+          saveName: deploymentName,
+          config: config.configuration,
+          configName: config.configurationName,
+          serverUrl: credential.url,
+          serverType: credential.serverType,
+          existingContentId,
+          existingCreatedAt,
```
dotNomad: Is there a reason that we don't let `runTsDeployWithProgress` handle grabbing the Content Record or pass it a full one?

Nitpicky, but curious.

extensions/vscode/src/views/tsDeployProgress.ts
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/**
+ * Inject a synthetic EventStreamMessage into the event stream so that
+ * the Publishing Log tree view and raw log viewer pick it up.
+ */
+function injectStageEvent(
+  stream: EventStream,
+  step: PublishStep,
+  suffix: "start" | "success" | "failure" | "log",
```
dotNomad: Nice little TypeScript trick. We can avoid creating a new type here and use `PublishEvent["status"]` which is where these suffix statuses are defined.

We could of course extract them to another `type` too.

```suggestion
  suffix: PublishEvent["status"],
```

extensions/vscode/src/views/tsDeployProgress.ts
```diff
@@ -0,0 +1,350 @@
...
+                  waitForTaskStage = "publish/runContent";
+                }
+              }
+
+              // Detect package installations for progress label updates.
+              const pkgEvent = packageEventFromLogLine(msg);
+              if (pkgEvent) {
+                stream.injectMessage(
+                  makeMessage(
+                    `${waitForTaskStage}/status` as EventSubscriptionTarget,
```
dotNomad: TypeScript is saying this can be `"publish/runContent/status"` which isn't a `EventSubscriptionTarget`. The type assertion here is hiding a bug.

extensions/vscode/src/views/tsDeployProgress.ts
```diff
@@ -0,0 +1,350 @@
...
+                    pkgEvent,
+                  ),
+                );
+              }
+
+              const type = `${waitForTaskStage}/log` as EventSubscriptionTarget;
+              stream.injectMessage(
+                makeMessage(type, { message: msg, level: "INFO" }),
+              );
```
dotNomad: To get rid of the type assertion we can do


```suggestion
              stream.injectMessage(
                makeMessage(`${waitForTaskStage}/log`, {
                  message: msg,
                  level: "INFO",
                }),
              );
```

packages/connect-api/src/client.ts
```diff
@@ -41,6 +41,31 @@ export class ConnectAPIError extends Error {
...
+  "python-streamlit",
+  "python-gradio",
+  "python-panel",
+  "quarto-shiny",
+  "quarto-static",
+  "api",
+  "shiny",
+  "rmd-shiny",
+  "rmd-static",
+]);
```
dotNomad: These are the same as the app modes we use in `appMode.ts`. I wonder if we should:
- move the set there
- have a generic type so we aren't duplicating quite as much between the two.

christierney: Moving the set into appMode.ts feels a little weird to me because then the connect-api package would be depending on the extension. I could see using a shared type that lives in connect-api since the extension already imports from here. What do you think?

PR #3857, opened 2026-03-20
feat: add token auth support to TypeScript ConnectAPI client

zackverham:
## Summary

- Ports Go token-based authentication (RSA key-pair signing from PR #2769) to the TypeScript `ConnectAPI` client in `packages/connect-api/`
- The client now supports two auth modes: API key (`Authorization: Key <apiKey>`) and token auth (per-request `X-Auth-Token`, `X-Auth-Signature`, `X-Content-Checksum`, `Date` headers)
- Updates all 4 extension call sites to pass `token` and `privateKey` from credentials to `ConnectAPI`

## Changes

**New files:**
- `packages/connect-api/src/auth.ts` — Token signing module: MD5 checksum, canonical request building, RSA-SHA1 signing, and `signRequest()` entry point
- `packages/connect-api/src/auth.test.ts` — 15 unit tests with round-trip signature verification

**Modified files:**
- `packages/connect-api/src/types.ts` — `apiKey` is now optional; added `token` and `privateKey` fields to `ConnectAPIOptions`
- `packages/connect-api/src/client.ts` — Constructor validates auth options and adds an axios request interceptor for token auth
- `packages/connect-api/src/client.test.ts` — Mock axios updated to support interceptors; added token auth and constructor validation tests
- `extensions/vscode/src/views/homeView.ts` — 3 call sites pass `token`/`privateKey`; guard updated to `credential?.apiKey || credential?.token`
- `extensions/vscode/src/connect_content_fs.ts` — 1 call site passes `token`/`privateKey`

## Token auth protocol

Per-request headers computed by signing a canonical request (`METHOD\nPATH\nDATE\nCHECKSUM`) with RSA-SHA1 using a base64-encoded DER PKCS#1 private key. Matches the Go implementation in `internal/api_client/auth/token.go`.

## Test plan

- [x] `cd packages/connect-api && npx vitest run` — 72 tests pass (15 new auth tests + updated client tests)
- [x] `cd extensions/vscode && just lint` — no lint errors
- [x] `cd extensions/vscode && npm run test-unit` — 865 extension unit tests pass


extensions/vscode/src/views/homeView.ts
christierney: is it ok that we don't check for the presence of url/apiKey/token/privateKey here like we do above? I know that difference is pre-existing, so maybe it's fine.

PR #3866, opened 2026-03-21
fix: adjust TS toml output

### LINKED ISSUE #3859: deployment TOML formatting differences
Comparing a deployment record created by Go with one created by TOML, there are some differences. Maybe worth cleaning up as the Go format seems more readable. Also I am using diff to compare deployments created by each path and the diffs are really annoying to read right now.

- comments in [configuration] — TypeScript includes it, Go omits it. Probably does not make sense to copy these into deployment record.
- Single vs double quotes — Go uses single quotes, TypeScript uses double. Pick one for consistency (single quotes are more idiomatic in TOML). This probably doesn't matter but makes diffing the files very annoying during testing.
- Array formatting — Go uses multi-line arrays; TypeScript uses inline. Multi-line is more readable and diff-friendly. Consider matching Go's style.
- dismissed_at missing — TypeScript doesn't write this field. If consumers expect it, this is a bug. (TS is omitting empty fields, probably ok)

christierney:
## Intent

Improve the readability/editability of the TOML produced by the TS code.

- put each element of top-level arrays (e.g. `files`) on a new line
- omit fields that should be omitted when printing a config inside a deployment (e.g. comments)

Fixes #3859

Not included:
- single quotes: not supported by any of the libraries I looked at
- missing fields: the TS writer strips empty fields before writing; not a problem

## Type of Change

- - [x] Bug Fix 

## Approach

Looked for another TOML library that would be a suitable replacement for smol-toml and allow the formatting options we want. Didn't find any good candidates that wouldn't introduce other problems (e.g. loss of detailed information for TOML parse errors.)

Instead, added a manual step after stringifying that reformats top-level arrays to introduce newlines. Skips more complicated nested structures to be safe, although our code doesn't produce any of those at the moment.

Copied existing logic to delete select Config fields into the deployment writer.

## User Impact

TOML files will be easier to manually inspect and edit (e.g. changing file include/exclude patterns.)

## Automated Tests

Added tests to cover new behavior.

## Checklist

### DIFF
--- a/extensions/vscode/src/toml/tomlHelpers.ts
+++ b/extensions/vscode/src/toml/tomlHelpers.ts
@@ -77,7 +77,7 @@ export function expandInlineArrays(toml: string, indent = "    "): string {
       continue;
     }
 
-    const prefix = match[1];
+    const prefix = match[1] ?? "";
     const arrayBody = line.slice(prefix.length);
     const items = parseInlineArray(arrayBody);
 


dotNomad (approved):
Looks great! I cannot wait to get this in.

PR #3882, opened 2026-03-24
feat: surface content error logs when deployment validation fails

### LINKED ISSUE #3879: display logs in publisher on content validation failure
see https://github.com/posit-dev/positron/issues/12679

If a deployment succeeds but then content validation fails, we don't display the error logs---user has to click to go view them in Connect. They are available via the Connect API and should be displayed without leaving publisher.

christierney:
## Summary
- When content validation fails during deployment, fetches the latest failed job's error log from Connect via `GET /v1/content/{guid}/jobs/{key}/log`
- Displays the tail 10 lines as synthetic log events in the Logs tree view under the "Validate Deployment Record" stage
- Clicking the header line opens the full job log in a new editor tab
- Adds `getJobs()` and `getJobLog()` methods to the `@posit-dev/connect-api` TypeScript client
- Tracks `contentId` and `serverUrl` during publish events and resolves credentials to make direct Connect API calls from the extension

Fixes #3879 

## Test plan
- [ ] Deploy content that fails validation and verify tail log lines appear in the tree view
- [ ] Click the "Showing last N of M" header to open the full log
- [ ] Verify no job log fetch occurs for non-validation failures
- [ ] Run `npx vitest run src/views/logs.test.ts` (24 tests pass)
- [ ] Run `npm test` in `packages/connect-api` (58 tests pass)

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -212,6 +212,12 @@
         "category": "Posit Publisher"
       },
       {
+        "command": "posit.publisher.logs.viewJobLog",
+        "title": "View Full Job Error Log",
+        "icon": "$(output)",
+        "category": "Posit Publisher"
+      },
+      {
         "command": "posit.publisher.homeView.navigateToDeployment.Server",
         "title": "Browse Deployment Server",
         "icon": "$(server)",
@@ -347,6 +353,10 @@
         },
         {
           "command": "posit.publisher.logs.visit",
+          "when": "false"
+        },
+        {
+          "command": "posit.publisher.logs.viewJobLog",
           "when": "false"
         },
         {
--- a/extensions/vscode/src/constants.ts
+++ b/extensions/vscode/src/constants.ts
@@ -44,6 +44,7 @@ const baseContexts = {
 
 const logsCommands = {
   Visit: "posit.publisher.logs.visit",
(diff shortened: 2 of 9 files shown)


christierney:
this is just a prototype, not ready to merge

some early feedback:

- **Medium** — `extensions/vscode/src/views/logs.ts:378-385`
  The job selection logic is too broad: `jobs.filter((j) => j.status !== 0)` treats every non-zero status as a failed job and then picks the newest one across the entire content item. If Connect uses non-terminal status codes (queued/running/canceled), or if the content already has older failed jobs, this can surface the wrong log. Suggested fix: filter to terminal failed jobs only and scope selection to the current publish (for example by publish start time, task/job ID, or a short retry until the current failed job appears).

- **Medium** — `extensions/vscode/src/views/logs.ts:419-420`
  `console.error("Failed to fetch job error log:", error)` may dump the full Axios error object, which commonly includes request config and the `Authorization: Key ...` header. That risks leaking API credentials into logs/devtools. Suggested fix: log a sanitized message/status only, or redact headers before logging.

- **Low** — `packages/connect-api/src/client.test.ts:1000-1037`
  `getJobLog()` claims to support the real `{ entries: [...] }` response shape, but the tests only cover a raw array. Add a unit/contract test for the wrapped payload so the production response format is actually validated.

PR #3893, opened 2026-03-26
feat: show directory checkbox state based on children inclusion

### LINKED ISSUE #2219: Use `indeterminate` state in Project Files with new API attributes

### LINKED ISSUE #3472: feat: show folders as checked if all children are checked
## Description
#2219 covers what to do about indeterminate state, but this issue is to cover the behavior when all items in a folder are included in the deployment. Currently, we don't show the containing folder as checked, but if all the children are included, we should.

You can see this if you create a project with a subfolder where the entry references files in the subfolder, and those references are auto-detected. For example:

Repro:
1. Create an `index.html` file with a reference to a `style.css` file in a `css` sub-folder (and no other files in `css/`)
2. Use the `index.html` as the entry
3. Look at the project files: the `index.html` and `css/style.css` files will be checked, but `css/` won't be.

dotNomad:
## Intent

Add an indeterminate state for Project File checkboxes that shows when some of the children in the directory are included in the deployment. The checkbox will also now show as checked if all children are included in the deployment.

<details>
  <summary>Preview</summary>
  
https://github.com/user-attachments/assets/ec26ba79-f4b6-44d3-b0f5-12a888a42da6

</details> 

Resolves #3472
Resolves #2219

## Type of Change

- [x] New Feature

## Approach

The backend (`fileTree.ts`) already computes `allIncluded`/`allExcluded` on every directory node. `CodeCheckbox.vue` already had CSS for the `:indeterminate` pseudo-class. This PR wires the two together:

Simplified the event from `changed: [boolean]` to `toggle: []` since the boolean payload was unused by the only consumer.

## User Impact

It is now clearer what files are included and which aren't when configuring a deployment.

## Automated Tests

Added Vitest component tests for all three modified components (29 new tests):

- `CodeCheckbox.test.ts` — DOM state for each `state` prop value, disabled behavior, toggle event
- `TreeItemCheckbox.test.ts` — state passthrough, event routing for all three states
- `ProjectFile.test.ts` — checkState computation for 7 scenarios (included/excluded/no-reason files, all-included/all-excluded/mixed/empty dirs), listStyle, tooltip, store event calls

Also converted `src/test/mocks/index.js` → `.ts` with proper types, and extended the shared mocks with `useFileStore` and additional home store fields.

## Directions for Reviewers

1. Run the extension (F5)
2. Create a project with subfolders containing mixed included/excluded files
3. Verify directory checkboxes reflect their children's state
4. Verify clicking an indeterminate directory includes all children
5. Hover over an indeterminate directory to verify the tooltip

## Checklist

### DIFF
--- a/extensions/vscode/webviews/homeView/src/components/views/projectFiles/ProjectFile.vue
+++ b/extensions/vscode/webviews/homeView/src/components/views/projectFiles/ProjectFile.vue
@@ -71,11 +71,20 @@ const isIncluded = computed((): boolean => {
 });
 
 const checkState = computed((): CheckState => {
+  // Directories with children use tri-state based on child inclusion
   if (props.file.isDir && props.file.fileCount > 0) {
-    if (props.file.allIncluded) return "checked";
-    if (!props.file.allExcluded) return "indeterminate";
-    return "unchecked";
+    if (props.file.allIncluded) {
+      // Every child is included
+      return "checked";
+    }
+    if (props.file.allExcluded) {
+      // Every child is excluded
+      return "unchecked";
+    }
+    // Some children included, some excluded
+    return "indeterminate";
   }
+  // Individual files are either included or not
   return isIncluded.value ? "checked" : "unchecked";
 });
 


extensions/vscode/webviews/homeView/src/components/CodeCheckbox.vue
```diff
@@ -2,7 +2,9 @@
   <div class="vscode-checkbox" :class="{ disabled: disabled }">
     <label class="checkbox-control">
       <input
-        :checked="checked"
+        :checked="state === 'checked'"
+        :indeterminate="state === 'indeterminate'"
+        :aria-checked="state === 'indeterminate' ? 'mixed' : undefined"
```
m--: is `undefined` right here if the state is not `'indeterminate'`? Should it not be `true` or `false`?

extensions/vscode/webviews/homeView/src/components/views/projectFiles/ProjectFile.vue
```diff
@@ -69,6 +70,15 @@ const isIncluded = computed((): boolean => {
   return Boolean(props.file.reason?.exclude === false);
 });
 
+const checkState = computed((): CheckState => {
+  if (props.file.isDir && props.file.fileCount > 0) {
+    if (props.file.allIncluded) return "checked";
+    if (!props.file.allExcluded) return "indeterminate";
```
m--: Agreed with the auto-review on this being kind of tricky to parse — is there a reason to organize these conditionals this way? If yes, a comment would be helpful I think.

Issue #3638, opened 2026-03-03
TS inspection service migration


christierney:
Migrate the `/api/configurations/inspect` behavior from the Go API to TypeScript.

## Goal
TypeScript inspection reproduces the currently implemented behavior of the Go `/inspect` endpoint.

## Scope boundary
Only behavior that is currently implemented by the Go `/inspect` endpoint is in scope for this parent issue.

## Done when
- The currently supported `/inspect` endpoint behavior is implemented in TypeScript
- `/inspect` requests no longer depend on the Go endpoint

## Tracking
Use the existing sub-issues already attached to this parent; no tracking model changes are required.

jonkeane:
> ... and R/Python processes.

I know I've said this [in a few places](https://github.com/posit-dev/publisher/issues/3443#issuecomment-3937108252), but leaning on Positron for the process running will give folks environments they are used to (+ and we will be able to delete a bunch of code that attempts to marshal an R or Python process into what it's like in Positron).

"But VSCode!" I hear you saying. The vast majority of our users are on Positron, so if we can provide an excellent experience on Positron that is slightly-less-excellent on VSCode that's the right trade off to make.

dotNomad:
Closing this up as all sub-issues have been finished 🎉

PR #3921, opened 2026-03-31
feat: switch inspection call sites to TypeScript service

zackverham:
## Summary
- Updates 3 call sites to use the new TypeScript `inspectProject()` instead of the Go backend's `POST /inspect` endpoint (#3638)
- Removes the Go backend dependency for content type detection
- **Depends on #3920** (the inspection service module PR)

## Changes
- `entrypointTracker.ts` — resolves relative dir to absolute path, calls `inspectProject()` directly
- `multiStepInputs/newDeployment.ts` — replaces `api.configurations.inspect()` with `inspectProject()`
- `multiStepInputs/selectNewOrExistingConfig.ts` — replaces `api.configurations.inspect()` with `inspectProject()`

## Test plan
- [x] `just lint` — 0 errors, 0 warnings
- [x] `npm run test-unit` — all tests pass
- [x] Manual: open a Flask project → verify content type detected as "python-flask"
- [x] Manual: open a Shiny for Python project → verify detected as "python-shiny"
- [x] Manual: open an R Shiny project with app.R → verify detected as "r-shiny"
- [x] Manual: verify new deployment wizard shows correct content types
- [x] Manual: verify entrypoint detection works in the status bar

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newDeployment.ts
+++ b/extensions/vscode/src/multiStepInputs/newDeployment.ts
@@ -202,6 +202,7 @@ export async function newDeployment(
         pythonPath: python?.pythonPath,
         rPath: r?.rPath,
         entrypoint: relEntryPointFile,
+        relativeDir: relEntryPointDir,
       });
 
       inspectionResults.forEach((result) => {
--- a/extensions/vscode/src/inspect/index.test.ts
+++ b/extensions/vscode/src/inspect/index.test.ts
@@ -89,4 +89,33 @@ describe("inspectProject", () => {
       expect(result.projectDir).toBe(".");
     }
   });
+
+  test("uses relativeDir as projectDir in results when provided", async () => {
+    mockReaddir.mockResolvedValue(["app.py"]);
+    mockStat.mockResolvedValue({ isFile: () => true });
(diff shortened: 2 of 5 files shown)


extensions/vscode/src/multiStepInputs/newDeployment.ts
```diff
@@ -195,16 +193,18 @@ export async function newDeployment(
...
-        {
-          entrypoint: relEntryPointFile,
-        },
-      );
+      const root = workspaces.path();
+      const absoluteDir = root
+        ? path.resolve(root, relEntryPointDir)
+        : relEntryPointDir;
+
+      inspectionResults = await inspectProject({
```
christierney: TS style question: why no `const` here?

dotNomad: Good question. It looks like `inspectionResults` were raised at one point so multiple steps could access them, but now they are only used in a single spot. I'll clean that up a bit.

dotNomad:
It looks like this is missing a bit that lets users select the two different ways to publish a Quarto project if they have the Quarto CLI installed for inspection and rendering.

CleanShot 2026-04-07 at 12 35 07

PR #3945, opened 2026-04-07
feat: add cancellation support to TS publish path

### LINKED ISSUE #3833: TS publish path: add cancellation support via AbortController

christierney:
Closes #3833

## Summary

Adds cancellation support to the TypeScript publish path so users can dismiss in-progress deployments from the progress notification, matching the Go path's existing UX.

This is an improvement over the Go path: the Go `CancelDeployment()` only marks the deployment record as dismissed while the goroutine continues running all publish steps. The TS path actually stops making API calls when cancelled, and aborts in-flight HTTP requests (including the long `waitForTask` polling loop).

## How it works

1. User clicks cancel on the VS Code progress notification
2. `AbortController.abort()` fires, which does two things simultaneously:
   - Injects `publish/failure` with `canceled: "true"` into the EventStream (logs tree shows "Deployment dismissed")
   - Sends `PUBLISH_CANCEL` to the webview (sidebar resets)
3. Inside `connectPublish`, the signal is checked at the next step boundary (`throwIfCanceled`), or the current axios request aborts mid-flight
4. The deployment record is written with `dismissed_at` (no `deployment_error`)
5. `CanceledError` propagates up and is silently handled (no error notification shown)
6. Any server-side task already running continues — Connect doesn't support task cancellation (same as Go path)

## Suggested reading order

**Start with the types and public API surface** — read the new `CanceledError` class and `signal` additions to `ConnectPublishOptions` in `connectPublish.ts` (lines 51–60, 117–119).

**Then the ConnectAPI plumbing** — `packages/connect-api/src/client.ts` is mechanical: every method got an optional `signal` parameter forwarded to axios. The interesting one is `waitForTask` which checks `signal.throwIfAborted()` at the top of each poll iteration.

**Then the orchestrator** — the `throwIfCanceled()` helper and catch-block guards in `connectPublish.ts` are the core logic. Note the two-phase catch: `CanceledError` (from between-step checks) and `signal?.aborted` (from in-flight axios aborts that throw axios's own error type).

**Finally the UI wiring** — `tsDeployProgress.ts` bridges VS Code's `CancellationToken` to the `AbortController` and handles event injection. `homeView.ts` is a one-line change to pass the signal through.

## Test plan

- [ ] 6 new tests in `connectPublish.test.ts` — pre-aborted signal, between-step abort, dismissedAt written, no deploymentError, no failure events, in-flight abort normalization
- [ ] 4 new tests in `tsDeployProgress.test.ts` — cancellable flag, canceled event injection, onCancel callback, no duplicate failure
- [ ] 2 new tests in `client.test.ts` — testAuthentication rethrows cancel errors, still wraps non-cancel errors
- [ ] 84 existing signal-forwarding tests in `client.test.ts`
- [ ] All 1359 extension vitest tests pass
- [ ] All 105 connect-api tests pass
- [x] Manual: deploy via TS path, click cancel during waitForTask, verify dismissed_at in TOML and "Deployment dismissed" in logs tree

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed deployment records being written with an empty `created_at` timestamp after publishing. (#3718)
 - Fixed configuration schema to use `auth_type` (snake_case) for integration requests, matching the format the extension actually produces. Added strict property validation to integration request items. (#3651)
+- Fixed a bug with cancellation that could cause redeployment to create a new content item instead of updating the existing one. (#3305)
 
 ## [1.36.0]
 


extensions/vscode/src/publish/connectPublish.ts
```diff
@@ -585,6 +652,26 @@ export async function connectPublish(
       bundleId: bundleDTO.id,
     };
   } catch (err) {
+    // Cancellation is not an error — dismissedAt was already written by throwIfCanceled
+    if (err instanceof CanceledError) {
+      throw err;
```
dotNomad: Ah I missed a beat here. At first I thought we shouldn't throw but I see that the `connectPublish` is inside of a `try` `catch`, but a bit obscured in `tsDeployProgress.ts`. We pass the `deploy` function which itself is a in a `try` `catch` 👍

PR #3953, opened 2026-04-08
feat: migrate Python package scan from Go to TypeScript

### LINKED ISSUE #3917: Migrate Python package scan endpoint from Go to TypeScript
Parent: #3636

The "Scan" button in the sidebar triggers `onScanForPythonPackageRequirements()` in `homeView.ts`, which calls the Go backend's `POST /api/packages/python/scan` endpoint to generate a `requirements.txt` file.

The TypeScript `pythonPackages.ts` module (`scanPythonPackages`) already performs the same work — running the Python interpreter, calling `pip list`, and matching imports to installed packages. The migration is to call the TS scanner directly, write the file from the extension, and remove the Go endpoint.

## Sub-tasks

- [ ] Replace `api.packages.createPythonRequirementsFile()` call in `homeView.ts` with direct call to TS `scanPythonPackages`
- [ ] Write the requirements file from the extension (the Go endpoint currently handles file writing)
- [ ] Return incomplete import warnings to the UI (same UX as current Go path)
- [ ] Remove `Packages.ts` API resource and Go `POST /api/packages/python/scan` endpoint

### LINKED ISSUE #3963: Python/R Packages panel shows 'file does not exist' after scanning when package file was absent at startup
## Description

When a Python or R project is opened without a requirements.txt or renv.lock file, and the user scans to create one, the Packages panel sometimes continues to show "file does not exist" even though the scan succeeded and the file was created. Clicking the Refresh button does not help.

## Root Cause

`detectPythonInterpreter` checks whether requirements.txt exists on disk and sets `packageFile` to `""` when it doesn't. This empty string propagates through `UpdateConfigWithDefaults` into the cached configuration.

`refreshPythonPackages()` had two variables: `packageFile` (raw config value, potentially `""`) and `resolvedPackageFile` (with `"requirements.txt"` fallback). The file read used the resolved value and succeeded, but the webview message sent the raw `""`, which is falsy. The webview's `isMissingRequirements` computed property evaluated to `true`, showing the "file does not exist" message.

The same pattern existed in `refreshRPackages()` with `renv.lock`.

## Fix

Resolve the packageFile fallback before both reading the file and reporting to the webview, so the webview always receives the resolved filename when the file exists. Also explicitly refresh packages and files list after scan completes rather than relying solely on debounced file watchers.

Fixed in `chris/ts-python-scan` branch (commit aa4cf4b24).

christierney:
## Summary

Migrates the Python package scanning endpoint (`POST /api/packages/python/scan`) from the Go backend to a TypeScript implementation that runs directly in the VSCode extension. This eliminates an HTTP round-trip and replaces Go's regex-based import extraction with Python's own `ast.parse` for more accurate results.

Fixes #3917
Fixes #3963

## Overview

**Approach:** An embedded Python script (following the same pattern as `rPackages.ts` for R scanning) is written to a temp file and executed via `execFile`. The script:

1. Walks the project directory for `.py`, `.ipynb`, `.qmd` files (skipping `.git`, `__pycache__`, `node_modules`, `.venv`, etc.)
2. Extracts imports using `ast.parse` (`.py`), JSON parsing (`.ipynb`), and line-by-line code fence detection (`.qmd`)
3. Filters stdlib imports via `sys.stdlib_module_names` (Python 3.10+) with a hardcoded fallback for Python 3.9
4. Filters local imports (matching local `.py` files or package directories)
5. Maps remaining imports to installed packages via `importlib.metadata.distributions()` + `top_level.txt`
6. Deduplicates by package name and outputs JSON: `{"requirements": [...], "incomplete": [...]}`

**Key improvements over the Go implementation:**
- `ast.parse` is more accurate than regex for import extraction (handles multi-line imports, comments, etc.)
- `importlib.metadata` is the canonical way to map imports to packages (reads `top_level.txt` for accurate mapping)
- No HTTP overhead — runs directly in the extension process

**Files changed:**
- **New:** `scanPythonDependencies.ts` — core scanner module with embedded Python script
- **New:** `scanPythonDependencies.test.ts` — 30 unit tests (mocked `execFile`/`fs`)
- **Modified:** `integration.test.ts` — 8 new integration tests with real Python
- **Modified:** `homeView.ts` — rewired to call TS scanner directly
- **Deleted:** `Packages.ts` (API wrapper), `post_packages_python_scan.go` + tests, `ScanPythonPackagesResponse` type
- **Modified:** `api_service.go` — removed route registration
- **Modified:** `client.ts` — removed `Packages` resource

## Bug fixes found during testing

Manual testing of the scanner uncovered several bugs, some pre-existing:

- **Per-project package file detection** (`state.ts`, `configurations.ts`): `refreshConfigurations` was calling `getInterpreterDefaults` once with the workspace root for all configs. Subdirectory projects couldn't find their `requirements.txt`/`renv.lock`. Fixed to compute defaults per-config using each project's directory. Removed `UpdateAllConfigsWithDefaults` to prevent reintroducing the single-defaults-for-all pattern.
- **Notebook cell boundary joining** (scanner): Jupyter notebooks omit trailing `\n` on the last line of each cell's source array. The scanner concatenated cells without separators, producing invalid syntax like `import numpyimport pandas` which `ast.parse` silently dropped. Fixed by appending `\n` after each cell.
- **Package name mapping for modern packages** (scanner): Packages like scikit-learn (1.8.0+) ship without `top_level.txt`. Added a `dist.files`/RECORD fallback to map import names (e.g., `sklearn` → `scikit-learn`) before falling back to package-name normalization.
- **Pre-existing: Packages panel showing "file does not exist" after scan** (`homeView.ts`, #3963): When interpreter defaults were computed before `requirements.txt` existed, `packageFile` was set to `""`. `refreshPythonPackages` read the file correctly via a fallback but sent `""` to the webview, which treated it as missing. Fixed by resolving the fallback before both reading and reporting. Same fix applied to R packages.

## For reviewers

- The embedded Python script lives as a template literal in `scanPythonDependencies.ts`. It uses `FENCE = chr(96) * 3` to avoid literal backticks in the JS string.
- Path validation (`validateSaveName`) uses `path.relative`-based containment checking — platform-aware and handles edge cases (root dirs, `..foo` literals).
- Validation runs **before** the Python scan to fail fast on bad inputs.
- The `homeView.ts` change resolves the relative `projectDir` from the webview to an absolute path (matching the R scan pattern).
- Integration tests use `hasRequirement()` helper for environment independence — assertions work whether packages are installed or not.

## Manual test checklist

- [x] Open a Python project in VSCode with the extension loaded
- [x] Click "Scan" / refresh on the Python requirements in the sidebar
- [x] Verify `requirements.txt` is created with correct header (`# auto-generated by Posit Publisher`) and package list
- [x] Verify stdlib imports (`os`, `sys`, `json`) are **not** included
- [x] Verify local imports (project `.py` files/packages) are **not** included
- [x] Verify third-party packages show `name==version` format for installed packages
- [x] Verify uninstalled imports appear as bare names and trigger the "incomplete" warning notification
- [x] Test with a project containing `.ipynb` notebooks — verify imports are extracted from code cells
- [x] Test with a project containing `.qmd` files — verify imports are extracted from `{python}` code blocks only (not from markdown text)
- [x] Test with no Python interpreter detected — verify fallback to `python3` and appropriate error if unavailable
- [x] Verify the Go backend is no longer called (check network tab or logs for absence of `POST /api/packages/python/scan`)


christierney:
Thanks for the thorough review!

**Re: minor suggestions**

1. **Extract Python script to separate file** — Agree this would be nice for DX, but it matches the existing pattern in `rPackages.ts` (embedded R script) and the build pipeline doesn't have a "bundle text files" step. Leaving as-is for consistency; worth revisiting if we add more embedded scripts.

2. **Error message specificity** — Good catch. Fixed in the next commit — now combines both `error.message` and `stderr` so neither is lost.

3. **Timeout configuration** — 60s is generous (the Go implementation had no timeout at all). Leaving hardcoded to avoid settings surface area for an unlikely scenario.

**Re: questions**

1. **`os.walk` and symlinks** — `os.walk` defaults to `followlinks=False`, so symlinks are already not followed. No change needed.

2. **Notebook `source` as single string** — The `\n` append at line 213 runs after both the `list` and `str` branches, so it handles both cases.

karawoo (approved):
the `ast.parse` approach is really nice! tried locally and everything worked as expected.

Issue #3714, opened 2026-03-12
Migrate credentials to TypeScript


dotNomad:
We already starting duplicating credentials to VS Code's `SecretStorage` API in #3635

Since we are using `SecretStorage` we do not need to handle the keyring and file credential differences, that handles all of that for us.

There are a few utilities we still need to move over from the Go side that are used in the Credential flows like:
- `ServerTypeFromURL()`
- `NormalizeServerURL()`
- `ConflictCheck()`
- `CreateCredentialDetails` validation
- all of our error types

We need to use the VS Code logging:

https://github.com/posit-dev/publisher/blob/d742bc21dcf4e4bc2497be54cb0fc2d8f804bbb1/extensions/vscode/src/logging.ts#L7-L10

Issue #3466, opened 2026-02-06
feat: support publishing to snowflake SPCS with token


christierney:
The snowflake support added in #2632 no longer works since Connect switched from trusting Snowflake headers for user identification, to using OIDC. To publish to Connect running in SPCS we now require dual authentication, both Snowflake and Connect.

This issue is for adding dual auth with Connect tokens.

See https://connect.posit.it/snowflake/demo.html for the current address of the dogfood SPCS server, which we can use for development and testing.

This will only work when Publisher is not also running in SPCS (see #3226).

Issue #3465, opened 2026-02-06
feat: fix dual auth for Snowflake SPCS (Connect API key + Snowflake token, and Connect token + Snowflake token)


christierney:
## Background

When publishing to Connect running in Snowflake SPCS, requests route through the Snowflake ingress proxy at `*.snowflakecomputing.app`. The proxy requires a Snowflake auth header. Connect itself also requires auth — and since Connect no longer trusts Snowflake headers for user identity (switched to OIDC), every request needs **two** auth headers.

## Current state (broken)

The `SnowflakeAuth` variant in `ConnectAPIOptions` (`packages/connect-api/src/types.ts`) sends only:

| Header | Value |
|---|---|
| `Authorization` | `Snowflake Token="<snowflake-token>"` |

No Connect credential is attached. Requests pass the ingress proxy but fail at Connect with 401. The credential creation wizard (`newConnectCredential.ts`) only collects a Snowflake connection name for Snowflake URLs — no Connect auth is gathered. `CredentialsService.create()` actively rejects Snowflake credentials that have an API key or token, making it impossible to store dual-auth credentials.

## Scope

Two dual-auth combinations need to work:

### 1. Snowflake token + Connect API key

| Header | Value |
|---|---|
| `Authorization` | `Snowflake Token="<snowflake-token>"` |
| `X-RSC-Authorization` | `Key <connect-api-key>` |

### 2. Snowflake token + Connect token (RSA keypair signing)

The token auth signing headers from `signRequest()` (`packages/connect-api/src/auth.ts`) do **not** use `Authorization` — they use `X-Auth-Token`, `X-Auth-Signature`, `X-Content-Checksum`, and `Date`. These can coexist with the Snowflake `Authorization` header, so:

| Header | Value |
|---|---|
| `Authorization` | `Snowflake Token="<snowflake-token>"` |
| `X-Auth-Token` | `<connect-token-id>` |
| `X-Auth-Signature` | `<per-request RSA-SHA1 signature>` |
| `X-Content-Checksum` | `<MD5 of request body>` |
| `Date` | `<request date>` |

### 1. `packages/connect-api/src/types.ts`

Replace the current `SnowflakeAuth` interface (which has no Connect auth) with two new discriminated union variants:

```typescript
interface SnowflakeApiKeyAuth extends ConnectAPIBaseOptions {
  snowflakeToken: string;
  apiKey: string;          // Connect API key → X-RSC-Authorization: Key <apiKey>
  token?: never;
[...]

PR #3958, opened 2026-04-09
Migrate Snowflake connection discovery to TypeScript

### LINKED ISSUE #3904: Migrate Snowflake connection discovery to TypeScript
Parent: #3714

Migrate the Snowflake connection discovery feature from Go (`internal/api_client/auth/snowflake/`) to TypeScript, replacing the `GET /api/snowflake-connections` Go API endpoint.

The Go implementation (~650 LOC) handles:
- Parsing Snowflake `connections.toml` config files
- Discovering Snowflake credentials from environment variables
- JWT token generation using Snowflake's OAuth token provider
- Connection validation against target Connect server
- URL normalization for various Snowflake URL formats

## Sub-tasks

- [ ] Parse Snowflake `connections.toml` in TypeScript
- [ ] Implement JWT token generation (Snowflake-specific)
- [ ] Connection validation against Connect server
- [ ] Wire into extension, replace Go API call in `common.ts`
- [ ] Remove Go `GET /api/snowflake-connections` endpoint

dotNomad:
## Summary

Resolves #3904 (parent: #3714)

- Migrates Snowflake connection discovery from the Go backend (`GET /api/snowflake-connections`) to TypeScript, running directly in the VSCode extension
- New `src/snowflake/` module handles config file parsing (connections.toml/config.toml), JWT and OAuth token generation, and connection validation via `@posit-dev/connect-api`
- Refactors `ConnectAPIOptions` to a discriminated union type enforcing auth method mutual exclusivity at compile time
- Removes the Go HTTP handler, its tests, and the extension's API resource wrapper

## Test Plan

- [x] 21 new Snowflake-specific tests (connections, token providers, discovery orchestrator)
- [x] Full extension test suite passes (1430 tests)
- [x] Go build and tests pass after endpoint removal
- [ ] Manual testing: create a credential against a Snowflake-proxied Connect server
- [x] Verify `deadcode` confirms no unintended Go code removal


packages/connect-api/src/types.ts
```diff
@@ -20,19 +20,47 @@ export const GUID = (id: string) => id as GUID;
 // Client options
 // ---------------------------------------------------------------------------
 
-export interface ConnectAPIOptions {
+interface ConnectAPIBaseOptions {
```
dotNomad, line 23: Rather than just having the keys here I used a discriminated union so the options had to be correct with the different Auth options. This way we cannot (without TypeScript complaining) pass an invalid option setup.

extensions/vscode/package.json
```diff
@@ -695,6 +696,7 @@
     "filenamify": "^7.0.1",
     "get-port": "^7.2.0",
     "js-yaml": "^4.1.1",
+    "jsonwebtoken": "^9.0.3",
```
dotNomad, line 700: Added a new dependency to handle JWTs in an ergonomic way, and to avoid writing our own `crypto` implementation.

extensions/vscode/src/credentials/service.ts
```diff
@@ -39,6 +39,36 @@ export interface CreateCredentialInput {
   privateKey?: string;
 }
 
+/**
+ * Build a ConnectAPIOptions discriminated union from a stored Credential.
+ * Picks the correct auth variant (ApiKeyAuth, TokenAuth, or NoAuth) based on
+ * which fields are present on the credential (empty string = absent).
+ */
+export function connectAPIOptionsFromCredential(
```
dotNomad, line 47: Created this to help create our `ConnectAPIOptions` based on the credential passed in. Especially with the new, more restricted typing.

christierney: nice!

christierney: what about the snowflake variant?

extensions/vscode/src/snowflake/discovery.ts
```diff
@@ -0,0 +1,56 @@
...
+  const urlCandidates = getListOfPossibleURLs(serverUrl);
+  const connections = listConnections();
+  const results: SnowflakeConnection[] = [];
+
+  for (const [name, config] of Object.entries(connections)) {
+    let tokenProvider;
+    try {
+      tokenProvider = createTokenProvider(config);
+    } catch {
+      // Skip connections with invalid config (e.g., missing key file)
```
christierney: nit: maybe helpful to log this? or maybe too noisy

extensions/vscode/src/snowflake/connections.ts
```diff
@@ -0,0 +1,237 @@
...
+];
+
+/**
+ * Returns potential config directories for Snowflake connections.
+ * Matches the Go implementation's search order.
+ */
+function findConfigDirs(): string[] {
+  const snowflakeHome = process.env.SNOWFLAKE_HOME;
+  if (snowflakeHome) {
+    // SNOWFLAKE_HOME is exclusive — when set it replaces all default paths.
```
christierney, line 34: I think this is a slight behavior change. (Go code continues to build up the list of potential other config dirs). This might be more correct, arguably.

christierney (approved):
Looks great! I think this is the right approach for now. When we do more follow-up work for snowflake we may want to look into what https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver provides.

extensions/vscode/src/credentials/service.ts
dotNomad: Right now `ConnectAPI` uses a `snowflakeToken` during `discovery.ts` before any credential is created. 

Once we move snowflake publishing from the Go side we can expand the helper to accept a Snowflake credential. I can do it now if you'd prefer though.

christierney: that's fine, I left that comment before I had read through everything else and forgot to circle back

extensions/vscode/src/snowflake/connections.ts
dotNomad: It is a behavior change good catch.

Going off of the snowflake docs: https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-using-the-connections-toml-file

> If a `~/.snowflake` directory exists on your machine, the Python Connector uses the `~/.snowflake/connections.toml` file. You can override the default `~/.snowflake` directory by setting the location in the `SNOWFLAKE_HOME` environment variable.
Otherwise, the Python Connector uses the `connections.toml` file in the one of the following locations, based on your operating system:
Linux: `~/.config/snowflake/connections.toml, but you can update it with XDG vars`
Windows: `%USERPROFILE%\AppData\Local\snowflake\connections.toml`
Mac: `~/Library/Application Support/snowflake/connections.toml`

Based on that I do think this is more correct. `SNOWFLAKE_HOME` is supposed to be the first, overriding default. If it doesn't exist there are fallbacks.

extensions/vscode/src/snowflake/discovery.ts
dotNomad: Good call. I added logging in https://github.com/posit-dev/publisher/pull/3958/commits/1ea79b8dfddd8892c5caec255f123746b86ce9ca

extensions/vscode/src/snowflake/connections.ts
christierney: Yeah, I confirmed this behavior with the snow cli too (it ignores other directories if SNOWFLAKE_HOME is set, even if SNOWFLAKE_HOME is invalid / contains no configs)

dotNomad:
There are some follow-ups here to get Snowflake publishing working as expected (#3465 and #3466).

The work here is matching the Go implementation as it is. We can adjust the TS implementation after merging in those other issues.

PR #3883, opened 2026-03-25
feat: generate requirements.txt from pylock.toml, uv.lock, or pyproject.toml

nealrichardson:
## Summary

- Auto-generates `requirements.txt` from Python lockfiles and project metadata when no `requirements.txt` exists on disk, enabling deployment without a manually maintained requirements file
- Supports three dependency sources with precedence: `pylock.toml` (PEP 751) > `uv.lock` > `pyproject.toml`
- Generated requirements are injected ephemerally into the deployment bundle without writing to disk
- Supports `optional_dependency_groups` config option for including extras from `pyproject.toml`
- Implemented in both Go backend (for deployments) and TypeScript extension (for UI package previews), with full test coverage
- Fixes a pre-existing bug where the `canGenerate` check in `checkRequirementsFile` incorrectly bypassed the file-inclusion validation
- Fixes error handling in `addDependenciesToTarget` to only fall back to generation on file-not-found errors, not all errors

Closes #2824

## Test plan

- [x] Go unit tests pass (`go test ./internal/inspect/dependencies/pydeps/...`)
- [x] TypeScript unit tests pass (28 tests in `pythonDependencySources.test.ts`)
- [ ] Manual test: deploy a Python project with only `pylock.toml` present (no `requirements.txt`)
- [ ] Manual test: deploy a Python project with only `uv.lock` present
- [ ] Manual test: deploy a Python project with only `pyproject.toml` present
- [ ] Manual test: verify precedence order when multiple sources exist
- [ ] Manual test: verify `optional_dependency_groups` includes extras from `pyproject.toml`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3966, opened 2026-04-10
feat: generate Python requirements from pylock.toml, uv.lock, or pyproject.toml

### LINKED ISSUE #2824: Generate requirements.txt from pyproject.toml, uv.lock, whatever else
It's fine (in the sense of a dog having coffee surrounded by flames) that Connect requires a requirements.txt file for specifying dependencies and doesn't support the other various Python standards. But just like how Publisher automatically generates a manifest.json based on its TOML, it could generate a requirements.txt from pyproject.toml, uv.lock, and whatever other files we wanted, so that it doesn't just error.

christierney:
## Summary

Adapts the TypeScript portions of #3883 to the current main branch, dropping all Go changes since that code is being migrated away. Key decisions and differences from the original PR:

- **TS-only**: No Go parser code. The lockfile parsers (`pythonDependencySources.ts`) and all wiring live entirely in the VSCode extension TypeScript.
- **"Generate once, use twice"**: Requirements are generated during preflight and stashed as `string[]`. The same array feeds both a synthetic `requirements.txt` in the bundle archive and the deployment record — no redundant lockfile parsing.
- **Synthetic file injection**: Added `syntheticFiles` support to the bundler so generated requirements can be included in the tar archive without writing a temporary file to disk. Synthetic files participate in `onProgress` callbacks and manifest hashing.
- **Empty-is-authoritative semantics**: An empty `packages` array in a lockfile (e.g., pylock.toml with `packages = []`) returns `[]`, stopping the priority chain. A missing section returns `null`, allowing fallback to the next source. Priority: pylock.toml > uv.lock > pyproject.toml.
- **Gated fallback**: Lockfile generation only triggers for the default `requirements.txt`. If the user configured a custom `package_file`, a missing file throws immediately rather than silently substituting generated dependencies.
- **Resolved bundler conflicts**: The original PR's `syntheticFiles` parameter and main's `onProgress` callback both touched the same function signatures in `archive.ts` / `bundler.ts` — merged both additions.

Closes #2824

## Test plan

- [ ] 29 unit tests for lockfile parsers (`pythonDependencySources.test.ts`) — all formats, priority ordering, edge cases, empty-vs-missing semantics
- [ ] 12 unit tests for `pythonPackages.ts` — fallback logic, gated on default file, preference for on-disk file
- [ ] 98 unit tests for `connectPublish.ts` — including 6 new tests for generated requirements (preflight, bundler synthetic files, deployment record, no-fallback cases)
- [ ] 2 bundler tests for synthetic file injection and progress events

### Manual tests
- [x] pylock.toml is used as requirements source
- [x] uv.lock is used as requirements source
- [x] pyproject.toml is used as requirements source
- [x] pylock.toml is used when all three options are present
- [x] existing requirements.txt is preferred when present
- [x] project with no dependency source fails preflight with clear error
- [x] project with custom package file (overriding "requirements.txt") fails preflight when that file is missing (does not fall back to scanning other lock files)
- [x] project with custom package file uses that file when it is present
- [x] pylock.toml with empty packages is respected (does not fall through to uv.lock even though package list is empty)
- [x] publish log progress still works with synthetic requirements.txt
- [x] generated bundle includes valid md5 for synthetic requirements.txt

### DIFF
--- a/extensions/vscode/src/interpreters/pythonDependencySources.ts
+++ b/extensions/vscode/src/interpreters/pythonDependencySources.ts
@@ -105,7 +105,7 @@ export async function readPyLockDependencies(
     result.push(`${pkg.name}==${pkg.version}`);
   }
 
-  return result.length > 0 ? result : null;
+  return result;
 }
 
 /** Shape of a [[package]] entry in uv.lock */
@@ -157,7 +157,7 @@ export async function readUvLockDependencies(
     result.push(`${pkg.name}==${pkg.version}`);
   }
 
-  return result.length > 0 ? result : null;
+  return result;
 }
 
 /**
--- a/extensions/vscode/src/interpreters/pythonDependencySources.ts
+++ b/extensions/vscode/src/interpreters/pythonDependencySources.ts
@@ -1,0 +1,197 @@
+// Copyright (C) 2026 by Posit Software, PBC.
+
+import path from "node:path";
+import { parse as parseTOML } from "smol-toml";
+
+import { readFileText } from "./fsUtils";
+
+/**
+ * Read Python dependencies from pyproject.toml's [project].dependencies.
+ *
+ * Optionally includes packages from [project.optional-dependencies] for the
+ * specified groups.
+ *
+ * Returns PEP 508 dependency specifier strings (e.g. "requests>=2.20"),
+ * or null if the file doesn't exist or has no [project].dependencies.
+ */
+export async function readPyProjectDependencies(
+  projectDir: string,
+  optionalGroups?: string[],
+): Promise<string[] | null> {
+  const content = await readFileText(path.join(projectDir, "pyproject.toml"));
+  if (content === null) {
+    return null;
+  }
+
+  let parsed: Record<string, unknown>;
+  try {
+    parsed = parseTOML(content) as Record<string, unknown>;
+  } catch {
+    return null;
+  }
+
+  const project = parsed.project;
+  if (!project || typeof project !== "object") {
+    return null;
+  }
+
+  const projectTable = project as Record<string, unknown>;
+  const deps = projectTable.dependencies;
+  if (!Array.isArray(deps)) {
+    return null;
+  }
+
+  const result: string[] = deps.filter(
+    (d): d is string => typeof d === "string",
+  );
+
+  if (optionalGroups && optionalGroups.length > 0) {
+    const optDeps = projectTable["optional-dependencies"];
+    if (optDeps && typeof optDeps === "object") {
+      const optTable = optDeps as Record<string, unknown>;
+      for (const group of optionalGroups) {
+        const groupDeps = optTable[group];
+        if (Array.isArray(groupDeps)) {
+          for (const d of groupDeps) {
+            if (typeof d === "string") {
+              result.push(d);
+            }
+          }
+        }
+      }
+    }
+  }
+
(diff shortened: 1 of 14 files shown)


extensions/vscode/src/interpreters/pythonDependencySources.ts
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Optionally includes packages from [project.optional-dependencies] for the
+ * specified groups.
+ *
+ * Returns PEP 508 dependency specifier strings (e.g. "requests>=2.20"),
+ * or null if the file doesn't exist or has no [project].dependencies.
+ */
+export async function readPyProjectDependencies(
+  projectDir: string,
+  optionalGroups?: string[],
```
christierney, line 19: as of now this param is unused. The proposal is to let users manually edit their config TOML to specify optional dependency groups, which would require a schema change. Filed #3967 to discuss later.

extensions/vscode/src/interpreters/pythonDependencySources.ts
```diff
@@ -0,0 +1,197 @@
...
+): Promise<string[] | null> {
+  const content = await readFileText(path.join(projectDir, "pyproject.toml"));
+  if (content === null) {
+    return null;
+  }
+
+  let parsed: Record<string, unknown>;
+  try {
+    parsed = parseTOML(content) as Record<string, unknown>;
```
zackverham, line 28: this looks like a lot of the type coercion stuff that jordan was flagging as a potential problem - I don't know how much we can get away from it when we're parsing a toml file, so it seems reasonable here to me. But just want to point it out.

dotNomad: Yeah we cannot get rid of it when we parse TOML files unfortunately.

Could be handy to have a utility for reading the various files, validating they are the shape we expect, and then returning the object with strict typing.

extensions/vscode/src/interpreters/pythonDependencySources.ts
```diff
@@ -0,0 +1,197 @@
...
+  const deps = projectTable.dependencies;
+  if (!Array.isArray(deps)) {
+    return null;
+  }
+
+  const result: string[] = deps.filter(
+    (d): d is string => typeof d === "string",
+  );
+
+  if (optionalGroups && optionalGroups.length > 0) {
```
zackverham, line 48: so because that param is unused, this code chunk is also unused? Is there any value in including it right now?

christierney: Fair question, I lean yes because I expect we will do the follow-up PR pretty quickly (I just wanted a chance to think through it and run it by the team first) and the code already existed (on Neal's PR that I was porting) so it felt worth hanging onto rather than recreate from scratch. If we opt to wontfix #3967 for some reason then we would want to delete it

extensions/vscode/src/interpreters/pythonDependencySources.ts
zackverham: could be a good follow up issue, i agree

extensions/vscode/src/interpreters/pythonDependencySources.ts
zackverham: sweet, makes sense to me

zackverham (approved):

Issue #3815, opened 2026-03-18
Port legacy R packages_from_library mapper to TypeScript


christierney:
Parent: #3788

## Context

The Go backend has a legacy `defaultPackageMapper` that is used when the user sets `r.packages_from_library = true` in their deployment config. Unlike the default lockfile-only mapper (already ported in #3810), this mapper reads each package's `DESCRIPTION` file from the locally installed R library and queries R for available packages and Bioconductor repos. It produces richer metadata (fields like `Packaged`, `Date/Publication`, `Built`).

The decision has been made to port this to TypeScript. Until this is done, the TS publish path routes `packages_from_library` configs to the Go backend.

## What needs to be ported

The Go implementation (`internal/publish/publish.go:104-109`, `internal/bundles/` package mapper):

- Read the `renv.lock` for the package list
- For each package, read its `DESCRIPTION` file from the local R library (`renv_library/`)
- Query R for available packages and Bioconductor repos
- Produce manifest package entries with the richer metadata fields

## Sub-tasks

- [ ] Implement `DESCRIPTION` file reader for R packages in the local library
- [ ] Implement R subprocess call to query available packages and Bioconductor repos
- [ ] Produce manifest package entries matching the Go output format
- [ ] Wire into TS publish orchestrator when `packages_from_library = true`
- [ ] Remove `packages_from_library` routing to Go path in `canUseTsPublishPath()`
- [ ] Add tests with golden file comparison against Go output

## Related

- Lockfile-only mapper (already ported): #3810
- TS publish orchestrator: #3828

PR #3978, opened 2026-04-13
Add DCF parser and test fixtures for R library mapper

zackverham:
## Summary
- Ports the DCF (Debian Control File) parser from the `kara-legacy-r-mapper` POC branch — used to read R DESCRIPTION files
- Adds test fixtures (renv.lock, DESCRIPTION files, expected-library.json golden files) copied from Go testdata, for use by the library mapper in follow-up PRs

Part 1 of 3 for #3815.

## Test plan
- [x] `parseDcf` unit tests pass (9 tests including golden file comparison against Go output)
- [x] `npm run test-unit` — all 1503 tests pass
- [x] `just lint` — clean


extensions/vscode/src/publish/dcfParser.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function trimRight(s: string): string {
+  let i = s.length;
+  while (i > 0 && WHITESPACE.includes(s.charAt(i - 1))) i--;
+  return s.slice(0, i);
+}
+
+function trim(s: string): string {
+  return trimRight(trimLeft(s));
+}
```
dotNomad, line 21: Isn't this just `String.prototype.trim()`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

extensions/vscode/src/publish/dcfParser.ts
```diff
@@ -0,0 +1,93 @@
...
+function trimLeft(s: string): string {
+  let i = 0;
+  while (i < s.length && WHITESPACE.includes(s.charAt(i))) i++;
+  return s.slice(i);
+}
+
+function trimRight(s: string): string {
+  let i = s.length;
+  while (i > 0 && WHITESPACE.includes(s.charAt(i - 1))) i--;
+  return s.slice(0, i);
```
dotNomad, line 16: I'm guessing we don't use `trimStart` and `trimEnd` because we need to know what was trimmed?

dotNomad (commented):
Little strange to see the custom String functions. Are those necessary?

extensions/vscode/src/publish/dcfParser.ts
zackverham: I think this is necessary because we're trimming on a particular whitespace marker (" \t") - it looks like the String.prototype.trim() functions just trim all whitespace. But whitespace seems to have a specific representation in DCF files.

dotNomad: Ahh that makes more sense! Thank you

dotNomad (approved):
Very nice port 🌟

PR #3979, opened 2026-04-13
Implement R library mapper in TypeScript

zackverham:
## Summary
- Ports Go's `defaultPackageMapper` to TypeScript (`rLibraryMapper.ts`)
- Includes R subprocess helpers (`getLibPaths`, `listAvailablePackages`, `getBioconductorRepos`) and package resolution logic (`toManifestPackage`, `readPackageDescription`)
- Uses `PackageLister` dependency injection for testability (no real R needed in tests)
- Golden file tests validate output matches Go's expected.json for CRAN and Bioconductor scenarios

Part 2 of 3 for #3815. Stacked on #3978.

## Test plan
- [x] Library mapper unit tests pass (20 tests: isCRANLike, toManifestPackage, readPackageDescription, golden files, error cases, integration with mocked lister)
- [x] `npm run test-unit` — all 1503 tests pass
- [x] `just lint` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3980, opened 2026-04-13
Wire library mapper into publish path and remove packagesFromLibrary gate

### LINKED ISSUE #3815: Port legacy R packages_from_library mapper to TypeScript

zackverham:
## Summary
- Wires `libraryToManifestPackages()` into `buildManifest()` — when `config.r.packagesFromLibrary` is true, uses the library mapper instead of the lockfile-only mapper
- Removes the `packagesFromLibrary` check from `canUseTsPublishPath()`, so these deployments now use the TS publish path instead of routing through Go
- Requires R interpreter when `packages_from_library` is enabled (throws if not found)

Part 3 of 3 for #3815. Stacked on #3979.

Closes #3815.

## Test plan
- [x] `canUseTsPublishPath` test updated — `packagesFromLibrary: true` now returns `true`
- [x] `npm run test-unit` — all 1503 tests pass
- [x] `just lint` — clean
- [ ] Manual test with a real R project that has `packages_from_library = true` — deployment should use TS path

### DIFF
--- a/extensions/vscode/src/publish/integration.test.ts
+++ b/extensions/vscode/src/publish/integration.test.ts
@@ -1,0 +1,407 @@
+// Copyright (C) 2026 by Posit Software, PBC.
+//
+// Integration tests for the R library mapper.
+//
+// Unlike the unit tests (rLibraryMapper.test.ts) which mock the PackageLister
+// interface, these tests exercise the real code paths against:
+//   - Real R subprocesses (getLibPaths, listAvailablePackages)
+//   - Real renv library directories (readPackageDescription)
+//   - Real renv::init() to populate a project's library
+//
+// Tests that require R or renv use `test.skipIf(!available)` so the suite can
+// run in any environment without failures. In CI, the
+// interpreter-integration.yaml workflow installs known R versions with renv
+// across a matrix of configurations to ensure full coverage.
+//
+// Run these tests with:
+//   npx vitest run src/publish/integration.test.ts
+//
+// Or as part of the dedicated npm script:
+//   npm run test-integration-publish
+
+import { execFile } from "node:child_process";
+import { mkdtemp, rm, writeFile, readFile, mkdir } from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { promisify } from "node:util";
+import { describe, expect, test } from "vitest";
+
+import {
+  getLibPaths,
+  listAvailablePackages,
+  readPackageDescription,
+  libraryToManifestPackages,
+  type PackageLister,
+} from "./rLibraryMapper";
+
+const execFileAsync = promisify(execFile);
+
+/** Create a temp directory, pass it to `fn`, then clean up. */
+async function withTempDir<T>(fn: (dir: string) => Promise<T>): Promise<T> {
+  const dir = await mkdtemp(path.join(os.tmpdir(), "publisher-libmap-"));
+  try {
+    return await fn(dir);
+  } finally {
+    await rm(dir, { recursive: true, force: true });
+  }
+}
+
+/**
+ * Check if an executable is available on PATH by attempting to run it.
+ * Used to determine whether interpreter-dependent tests should be skipped.
+ */
+async function isExecutableAvailable(name: string): Promise<boolean> {
+  try {
+    await execFileAsync(name, ["--version"]);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Check if the renv R package is installed by attempting to load it.
+ * Used to determine whether renv-dependent tests should be skipped.
+ */
+async function isRenvAvailable(): Promise<boolean> {
+  try {
+    await execFileAsync("Rscript", ["-e", "library(renv)"]);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+// ---------------------------------------------------------------------------
+// getLibPaths — real R subprocess
+// ---------------------------------------------------------------------------
+
+describe("getLibPaths (real R subprocess)", { timeout: 30_000 }, async () => {
+  const rAvailable = await isExecutableAvailable("R");
+
+  test.skipIf(!rAvailable)("returns non-empty array of absolute paths", () =>
+    withTempDir(async (dir) => {
+      const libPaths = await getLibPaths("R", dir);
+      expect(libPaths.length).toBeGreaterThan(0);
+      for (const p of libPaths) {
+        // Absolute path: starts with / on Unix or drive letter on Windows
+        // R on Windows may use forward slashes (C:/...) or backslashes (C:\...)
+        expect(p).toMatch(/^(\/|[A-Z]:[/\\])/i);
+      }
+    }),
+  );
+});
+
+// ---------------------------------------------------------------------------
+// listAvailablePackages — queries CRAN via a real R subprocess
+// ---------------------------------------------------------------------------
+
+describe(
+  "listAvailablePackages (real R + CRAN)",
+  { timeout: 120_000 },
+  async () => {
+    const rAvailable = await isExecutableAvailable("R");
+
+    test.skipIf(!rAvailable)(
+      "returns packages from CRAN including jsonlite",
+      () =>
+        withTempDir(async (dir) => {
+          const repos = [{ Name: "CRAN", URL: "https://cran.rstudio.com" }];
+          const packages = await listAvailablePackages("R", dir, repos);
+
+          expect(packages.length).toBeGreaterThan(0);
+          for (const pkg of packages) {
+            expect(pkg.name).toBeTruthy();
+            expect(pkg.version).toBeTruthy();
+            expect(pkg.repository).toBeTruthy();
+          }
+
+          // jsonlite is a well-known, stable CRAN package
+          const jsonlite = packages.find((p) => p.name === "jsonlite");
+          expect(jsonlite).toBeDefined();
+          expect(jsonlite!.version).toMatch(/^\d+\.\d+/);
+        }),
+    );
+  },
+);
+
+// ---------------------------------------------------------------------------
+// readPackageDescription — reads from a real renv library
+// ---------------------------------------------------------------------------
+
+describe(
+  "readPackageDescription (real renv library)",
+  { timeout: 120_000 },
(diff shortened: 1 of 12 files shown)


zackverham:
## Smoke Test Plan

This covers the full stacked PR set: #3978 (DCF parser), #3979 (R library mapper), #3980 (wire into publish path + remove gate).

### Prerequisites

- A Posit Connect server you can deploy to
- R installed locally (>= 4.3)
- An R project with `renv` initialized and library populated (`renv::restore()`)
- Build the extension from this branch (`just` from `extensions/vscode/`)

Test content is available in `smoke-test-content/` at the repo root (on this branch). Run `renv::restore()` in each project before testing.

---

### 1. Happy Path — `packages_from_library = true` (new library mapper) [SUCCESS]

**Setup:** Use `smoke-test-content/r-shiny-library/` or any renv-based R project. Ensure the `.posit/publish/*.toml` config has:

```toml
[r]
packages_from_library = true
package_manager = "renv"
```

**Steps:**
1. Open the project in VSCode with the dev extension
2. Add a Connect credential if not already present
3. Initiate a deployment via the Publisher sidebar
4. **Verify:** The deployment uses the **TS publish path** (not Go) — you should NOT see SSE event streaming from the Go backend
5. **Verify:** Progress log shows `"Loading packages from local R library"` message during manifest creation
6. **Verify:** The deployment completes successfully on Connect
7. **Verify:** The deployed content runs correctly (Shiny app loads, packages are available)

### 2. Happy Path — Standard lockfile mapper (regression) [SUCCESS]

**Setup:** Use `smoke-test-content/r-shiny-lockfile/` or any renv-based R project. Ensure the config does NOT have `packages_from_library = true` (or has it set to `false`).

**Steps:**
1. Open the project in VSCode with the dev extension
2. Initiate a deployment
3. **Verify:** The deployment uses the TS publish path
4. **Verify:** Progress log shows `"Scanning R lockfile for packages"` (not the library mapper message)
5. **Verify:** The deployment completes successfully on Connect

### 3. Error — Missing lockfile with `packages_from_library = true` [SUCCESS]

**Setup:** Use an R project with `packages_from_library = true` in config, but **delete or rename `renv.lock`**.

**Steps:**
1. Attempt to deploy
2. **Verify:** Error message: `packages_from_library requires an renv.lock file (expected at "renv.lock"). Run renv::snapshot() to create one.`

### 4. Error — Version mismatch between lockfile and library [SUCCESS]

**Setup:** Use a project with `packages_from_library = true`. Manually edit `renv.lock` to change a package's version to differ from what's installed in `renv/library/`.

**Steps:**
1. Attempt to deploy
2. **Verify:** Error message includes: `versions in lockfile '...' and library '...' are out of sync. Use renv::restore() or renv::snapshot() to synchronize`


### 5. Error - Lockfile doesn't contain a dependency that is installed in `renv/library` [SUCCESS WITH FUTURE IMPROVEMENT]

Similar to test 5 - but remove the dependency entirely.

Behavior is that the deployment succeeds, but the content fails to execute because the dependency is missing. 

This matches the behavior in the go implementation: https://github.com/posit-dev/publisher/blob/95b2e29cb/internal/inspect/dependencies/renv/manifest_packages.go#L261

Tracking issue for this here: https://github.com/posit-dev/publisher/issues/4006

extensions/vscode/src/publish/rLibraryMapper.ts
```diff
@@ -51,7 +51,8 @@ function runRExpression(
       {
         cwd,
         timeout: R_SUBPROCESS_TIMEOUT,
-        env: { ...process.env, RENV_CONFIG_AUTOLOADER_ENABLED: "FALSE" },
```
karawoo, line 54: was it intentional to remove `RENV_CONFIG_AUTOLOADER_ENABLED`?

extensions/vscode/src/views/canUseTsPublishPath.ts
```diff
@@ -5,18 +5,18 @@ import type { ConfigurationDetails } from "src/api/types/configurations";
...
+ * instead of the Go backend. Returns false for server types that are
+ * only implemented in Go (Connect Cloud).
+ *
+ * Note: packagesFromLibrary was previously gated here but is now
+ * handled by the TypeScript library mapper (rLibraryMapper.ts).
  */
 export function canUseTsPublishPath(
   serverType: ServerType,
-  config: ConfigurationDetails,
+  _config: ConfigurationDetails,
```
karawoo: can this be removed altogether if it's now unused?

karawoo (commented):
For the smoke tests, we might consider adding one where R is available and the lockfile is present but one or more packages is missing entirely from the library. That ought to be a similar error to test 5 (version mismatch).

extensions/vscode/src/views/canUseTsPublishPath.ts
zackverham: yes! fixed in the commit I just added.

karawoo: too fast for me!

extensions/vscode/src/publish/rLibraryMapper.ts
zackverham: yeah, its described in the comment above - it looks we need the autoloader to correctly get the libPaths to resolve appropriately.

extensions/vscode/src/publish/rLibraryMapper.ts
```diff
@@ -38,20 +38,41 @@ export function escapeForRString(s: string): string {
...
+ * environment-level R config, but user .Rprofile side effects (e.g.,
+ * setRepositories(), options(), network calls) will execute.
  */
 function runRExpression(
   rPath: string,
   expression: string,
   cwd: string,
 ): Promise<string> {
   return new Promise((resolve, reject) => {
+    // Use individual flags instead of -s (--slave) because -s implies
```
zackverham:
```suggestion
    // Use individual flags instead of -s because -s implies
```

extensions/vscode/src/publish/rLibraryMapper.ts
karawoo: cool yeah, I left this review before that comment was there I think

extensions/vscode/src/publish/rLibraryMapper.ts
```diff
@@ -167,15 +165,8 @@ export async function getBioconductorRepos(
...
-  const code = [
-    `if (requireNamespace("BiocManager", quietly = TRUE)`,
-    `    || requireNamespace("BiocInstaller", quietly = TRUE)) {`,
-    `  repos <- getFromNamespace("renv_bioconductor_repos", "renv")("${escapedDir}")`,
-    `  ; repos <- repos[setdiff(names(repos), "CRAN")]`,
-    `  ; cat(repos, labels = names(repos), fill = 1)`,
-    `  ; invisible()`,
-    `}`,
-  ].join("\n");
+  // Single-line to avoid Windows R -e parsing issues.
```
dotNomad, line 168: This sounds like a handy `CLAUDE.md` addition similar to the "Cross-Platform Path Handling" section

extensions/vscode/src/publish/integration.test.ts
```diff
@@ -0,0 +1,407 @@
...
+
+/** Create a temp directory, pass it to `fn`, then clean up. */
+async function withTempDir<T>(fn: (dir: string) => Promise<T>): Promise<T> {
+  const dir = await mkdtemp(path.join(os.tmpdir(), "publisher-libmap-"));
+  try {
+    return await fn(dir);
+  } finally {
+    await rm(dir, { recursive: true, force: true });
+  }
+}
```
dotNomad: Very nice use of generics

dotNomad: Ah I'm seeing this is the same as the utilities in `inspect/integration.test.ts`. I wonder if we can combine these into test utils since we have some duplication.

dotNomad (commented):
Fantastic 🎉 

Just had a couple notes for potential improvements, but nothing blocking.

extensions/vscode/src/publish/integration.test.ts
zackverham: @claude can you implement this suggestion?

extensions/vscode/src/publish/rLibraryMapper.ts
zackverham: @claude can you implement this suggestion?

zackverham:
@karawoo 

> For the smoke tests, we might consider adding one where R is available and the lockfile is present but one or more packages is missing entirely from the library. That ought to be a similar error to test 5 (version mismatch).

This is a good idea! I ended up just adding this type of test as one of the automated integration tests, so we have that coverage moving forward.

dotNomad (approved):
Fantastic improvements. Realizing now I didn't originally Approve, but I did mean to 😓

Issue #3916, opened 2026-03-30
Port Connect Cloud publish orchestrator sub-tasks


christierney:
Parent: #3795

Detailed sub-tasks for porting the Connect Cloud publish flow to TypeScript. The `packages/connect-cloud-api/` client already exists with OAuth support (#3820 merged).

## Sub-tasks

- [ ] Create Connect Cloud publish orchestrator (mirror `connectPublish.ts` pattern with Cloud-specific API calls)
- [ ] Integrate with `connect-cloud-api` OAuth token management for authenticated requests
- [ ] Wire routing in `canUseTsPublishPath()` to include Connect Cloud
- [ ] Add progress/log streaming for Cloud deployments (reuse `injectMessage()` pattern)

## Key differences from Connect path

The Connect Cloud API uses different endpoints and a different deployment model:
- `createContent` / `updateContent` via Cloud API (not Connect REST API)
- OAuth bearer tokens instead of API keys
- `publishContent` triggers deployment (different from Connect's bundle upload → deploy flow)
- Revision-based status polling instead of task-based

christierney:
closed by #3987

Issue #3795, opened 2026-03-17
Port Connect Cloud publisher to TypeScript


christierney:
## Context

Part of #3788 (migrate publish flow). Lower priority — do after Connect publishing works (#3793).

## What

Connect Cloud has a separate publish flow with different steps:
- Initiate publish (different API)
- Await completion
- Stream logs

This is a distinct implementation path from the Connect publisher.

## Where to look

- Go implementation: \`internal/publish/connect_cloud/\`
- TS types: \`ConnectCloudConfig\` in \`src/api/types/configurations.ts\`

PR #3986, opened 2026-04-13
Add Cloud SSE log streaming client

christierney:
## Summary

Part of #3916 (SSE log streaming portion)

Adds a Server-Sent Events client for streaming server-side logs during Connect Cloud deployments. This is **PR 1 of 2** for the Connect Cloud TypeScript publish migration (#3795).

- Adds `watchCloudLogs()` function to `@posit-dev/connect-cloud-api` that opens an SSE connection to the Cloud logs endpoint, parses batched log entries, and streams them to a callback
- Adds log streaming types (`LogLine`, `LogsEventData`, `WatchLogsOptions`) and environment base URLs (`cloudLogsBaseUrls`) to `types.ts`
- Matches Go behavior: 60-second lookback, Bearer token auth via custom fetch, silent skip of malformed events, resolve (not reject) on server-initiated stream closure
- Uses `eventsource@4.1.0` npm package for high-level SSE client with custom fetch support
- 20 unit tests covering URL construction, auth headers, log parsing, abort signal, stream closure, connection errors, and malformed JSON

## Test plan

- [x] `npx vitest run` in `packages/connect-cloud-api/` — all 77 tests pass (20 new)
- [x] Type-check passes (`npx tsc --noEmit`)
- [x] CI passes

### DIFF
--- a/packages/connect-cloud-api/src/logs.ts
+++ b/packages/connect-cloud-api/src/logs.ts
@@ -92,8 +92,11 @@ export async function watchCloudLogs({
     // code's WatchLogs resolves without error on stream close. Only reject
     // for genuine connection failures (readyState !== CLOSED).
     es.addEventListener("error", (err) => {
+      // Capture readyState before close() — close() sets it to CLOSED,
+      // which would mask genuine connection failures.
+      const wasCleanClose = es.readyState === EventSource.CLOSED;
       es.close();
-      if (es.readyState === EventSource.CLOSED) {
+      if (wasCleanClose) {
         resolve();
       } else {
         reject(
--- a/packages/connect-cloud-api/src/logs.ts
+++ b/packages/connect-cloud-api/src/logs.ts
@@ -29,9 +29,13 @@ import {
  * });
  * ```
  */
-export async function watchCloudLogs(options: WatchLogsOptions): Promise<void> {
-  const { environment, logChannel, authToken, onLog, signal } = options;
-
+export async function watchCloudLogs({
+  environment,
+  logChannel,
+  authToken,
+  onLog,
+  signal,
+}: WatchLogsOptions): Promise<void> {
   // Construct SSE URL with 60-second lookback (matching Go's logLookback)
   const baseUrl = cloudLogsBaseUrls[environment];
   const nowNanos = Date.now() * 1_000_000;
--- a/packages/connect-cloud-api/src/logs.ts
+++ b/packages/connect-cloud-api/src/logs.ts
@@ -36,12 +36,17 @@ export async function watchCloudLogs({
   onLog,
   signal,
 }: WatchLogsOptions): Promise<void> {
+  // Short-circuit if already aborted — avoid opening a connection at all.
+  if (signal?.aborted) {
+    return;
+  }
+
   // Construct SSE URL with 60-second lookback (matching Go's logLookback)
   const baseUrl = cloudLogsBaseUrls[environment];
   const nowNanos = Date.now() * 1_000_000;
   const lookbackNanos = 60 * 1_000_000_000; // 60 seconds in nanoseconds
   const sortKeyGt = nowNanos - lookbackNanos;
-  const url = `${baseUrl}/v1/logs/${logChannel}/stream?sort_key__gt=${sortKeyGt}`;
+  const url = `${baseUrl}/v1/logs/${encodeURIComponent(logChannel)}/stream?sort_key__gt=${sortKeyGt}`;
 
   return new Promise<void>((resolve, reject) => {
     // Create EventSource with custom fetch to inject Authorization header
--- a/packages/connect-cloud-api/src/logs.ts
+++ b/packages/connect-cloud-api/src/logs.ts
@@ -71,36 +71,28 @@ export async function watchCloudLogs(options: WatchLogsOptions): Promise<void> {
             message: entry.message,
           });
         }
-      } catch (err) {
+      } catch {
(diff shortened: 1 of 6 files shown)


packages/connect-cloud-api/src/types.ts
```diff
@@ -250,3 +250,46 @@ export interface AuthorizationResponse {
...
+/**
+ * Shape of each SSE event's JSON `data` field from the Cloud logs endpoint.
+ * Each event contains an array of log messages.
+ */
+export interface LogsEventData {
+  data: Array<{
+    timestamp: number;
+    sort_key: number;
+    message: string;
+    type: string; // "build" | "runtime"
```
dotNomad: If this is truly just `"build"` and `"runtime"` I think we should extract a type.

packages/connect-cloud-api/src/types.ts
```diff
@@ -250,3 +250,46 @@ export interface AuthorizationResponse {
...
+// ---------------------------------------------------------------------------
+// Log streaming types (from connect_cloud_logs/types.go)
+// ---------------------------------------------------------------------------
+
+export type LogLevel = "debug" | "info" | "error";
+
+export interface LogLine {
+  level: LogLevel;
+  message: string;
+}
```
dotNomad: Is `LogLine` part of the `LogsEventData.data`? If so we should collapse those types.

packages/connect-cloud-api/src/types.ts
```diff
@@ -250,3 +250,46 @@ export interface AuthorizationResponse {
...
+ */
+export interface LogsEventData {
+  data: Array<{
+    timestamp: number;
+    sort_key: number;
+    message: string;
+    type: string; // "build" | "runtime"
+    level: LogLevel;
+  }>;
+}
```
dotNomad: Based on the answer to my other questions here is a suggestion.

```suggestion
export type LogEntryType = "build" | "runtime";

export interface LogEntry {
  timestamp: number;
  sort_key: number;
  message: string;
  type: LogEntryType;
  level: LogLevel;
}

export type LogLine = Pick<LogEntry, "level" | "message">;

/**
 * Shape of each SSE event's JSON `data` field from the Cloud logs endpoint.
 * Each event contains an array of log messages.
 */
export interface LogsEventData {
  data: LogEntry[];
}
```

packages/connect-cloud-api/src/logs.ts
```diff
@@ -0,0 +1,98 @@
...
+ *   authToken: "cloud-token",
+ *   onLog: (line) => console.log(`[${line.level}] ${line.message}`),
+ *   signal: controller.signal,
+ * });
+ * ```
+ */
+export async function watchCloudLogs(options: WatchLogsOptions): Promise<void> {
+  const { environment, logChannel, authToken, onLog, signal } = options;
+
+  // Construct SSE URL with 60-second lookback (matching Go's logLookback)
```
dotNomad, line 44: Fun fact you can destruct in function parameters

```suggestion
export async function watchCloudLogs({
  environment,
  logChannel,
  authToken,
  onLog,
  signal,
}: WatchLogsOptions): Promise<void> {
  // Construct SSE URL with 60-second lookback (matching Go's logLookback)
```

packages/connect-cloud-api/src/logs.ts
```diff
@@ -0,0 +1,98 @@
...
+  const baseUrl = cloudLogsBaseUrls[environment];
+  const nowNanos = Date.now() * 1_000_000;
+  const lookbackNanos = 60 * 1_000_000_000; // 60 seconds in nanoseconds
+  const sortKeyGt = nowNanos - lookbackNanos;
+  const url = `${baseUrl}/v1/logs/${logChannel}/stream?sort_key__gt=${sortKeyGt}`;
+
+  return new Promise<void>((resolve, reject) => {
```
dotNomad, line 51: Am I reading this right that this will return a promise that resolves when the `EventSource` is `CLOSED`. We will have to be careful not to `await` this.

packages/connect-cloud-api/src/logs.ts
```diff
@@ -0,0 +1,98 @@
...
+      signal.addEventListener("abort", () => {
+        es.close();
+        resolve();
+      });
+    }
+
+    // Handle incoming SSE messages
+    es.addEventListener("message", (event) => {
+      try {
+        const data = JSON.parse(event.data) as LogsEventData;
```
dotNomad, line 75: Great use of a type assertion here 👍 

I call these out so much I feel the need to call out the good ones 😓

packages/connect-cloud-api/src/logs.ts
dotNomad: Totally fine with that, but once I see the usage I may have an alternative suggestion

dotNomad (approved):
Looks great! Excited to see it in use

packages/connect-cloud-api/src/logs.ts
christierney: Yeah, this follows the same pattern as the Go version, where we stream logs in a goroutine and simultaneously start polling another endpoint. It is not awaited in the current usage (next PR) but I am definitely open to alternatives.

packages/connect-cloud-api/src/logs.ts
christierney: I always appreciate the advice!

dotNomad (approved):
Great catch on the encoding of the URI component

PR #3987, opened 2026-04-13
Migrate Connect Cloud publishing to TypeScript

### LINKED ISSUE #3795: Port Connect Cloud publisher to TypeScript

christierney:
## Summary

- Extract shared publish utilities (`publishShared.ts`) from `connectPublish.ts` for reuse across Connect and Cloud orchestrators
- Add Cloud content request building (`cloudContentRequest.ts`) with content type mapping, environment-specific URLs, 5-level access control, and request construction
- Implement Cloud publish orchestrator (`connectCloudPublish.ts`) matching Go's flow: createManifest → createBundle → createContent/updateContent → initiatePublish → uploadBundle → watchLogs + awaitCompletion
- Wire Cloud publish path in HomeView with `ConnectCloudAPI` creation and automatic OAuth token refresh
- Extend `tsDeployProgress.ts` to handle Cloud-specific steps (`watchLogs`, `awaitCompletion`) with explicit start/success lifecycle events
- Remove the `canUseTsPublishPath` gate entirely — all server types now route through TypeScript
- Fix lossy `ViewPublicEditTeam` reverse mapping: derive `orgAccess=editor` (not `viewer`) to prevent silent permission downgrade on redeploy
- Surface HTTP response body in publish error messages for better diagnostics

Closes #3795, #3916

## Test plan

- [x] 1688 unit tests pass across orchestrator, content request, and progress handler test files
- [x] TypeScript strict-mode compilation clean
- [x] ESLint clean
- [ ] Manual test: Cloud first deploy from VS Code extension
- [ ] Manual test: Cloud redeploy (verify access control preserved)
- [ ] Manual test: Cancel mid-deploy
- [ ] Manual test: Token refresh during long deploy [not really sure how to test this one especially since I have an SSO account]


extensions/vscode/src/publish/cloudContentRequest.ts
```diff
@@ -0,0 +1,483 @@
...
+// "editor".
+//
+// See also: Go implementation in internal/publish/connect_cloud/content_request_base.go
+// ---------------------------------------------------------------------------
+
+/**
+ * Derives the orgAccess dimension from an existing ContentAccess value.
+ * Used during redeploy partial merges to preserve the dimension the user
+ * didn't explicitly set.
+ */
```
christierney, line 160: reviewers: you may want to compare the logic in the next few functions to https://github.com/posit-dev/publisher/blob/8a4b595d4943798551a1094510bc0757617be299/internal/publish/connect_cloud/content_request_base.go#L21

extensions/vscode/src/views/homeView.ts
```diff
@@ -416,7 +396,75 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+        refreshToken: credential.refreshToken,
+        environment: CONNECT_CLOUD_ENVIRONMENT,
+        onTokenRefresh: async (tokens) => {
+          credential.accessToken = tokens.access_token;
+          credential.refreshToken = tokens.refresh_token;
+          await storeCredential(this.context.secrets, credential);
+        },
+      });
+
+      runTsDeployWithProgress({
```
zackverham, line 415: not for now, but we probably can fast-follow and just call this `runDeployWithProgress`

extensions/vscode/src/views/tsDeployProgress.ts
```diff
@@ -1,16 +1,24 @@
...
-import { CanceledError } from "src/publish/connectPublish";
+import type { PublishResult, PublishStep } from "src/publish/publishShared";
+import { CanceledError } from "src/publish/publishShared";
+import type { CloudPublishStep } from "src/publish/connectCloudPublish";
 import type { EventStream } from "src/events";
 import type { EventStreamMessage, EventSubscriptionTarget } from "src/api";
 import type { ErrorCode } from "src/utils/errorTypes";
 
+// Union of all possible publish step types (standard Connect + Cloud)
+type AnyPublishStep = PublishStep | CloudPublishStep;
```
zackverham: types!

extensions/vscode/src/views/tsDeployProgress.test.ts
```diff
@@ -755,4 +755,192 @@ describe("runTsDeployWithProgress", () => {
...
+
+      const statusMsgs = stream.injected.filter(
+        (m) => m.type === "publish/restoreEnv/status",
+      );
+      expect(statusMsgs).toHaveLength(1);
+      expect(statusMsgs[0]!.data.name).toBe("numpy");
+      expect(statusMsgs[0]!.data.version).toBe("1.24.3");
+      expect(statusMsgs[0]!.data.runtime).toBe("python");
+    });
+  });
```
zackverham, line 945: not a big deal, but you * could * unit test `isServerLogStep`

extensions/vscode/src/publish/cloudContentRequest.ts
```diff
@@ -0,0 +1,483 @@
...
+export function getCloudUIURL(env: CloudEnvironment): string {
+  switch (env) {
+    case CloudEnvironment.Development:
+      return "https://dev.connect.posit.cloud";
+    case CloudEnvironment.Staging:
+      return "https://staging.connect.posit.cloud";
+    case CloudEnvironment.Production:
+      return "https://connect.posit.cloud";
+    default: {
+      const _exhaustive: never = env;
```
zackverham: what is this pattern doing? For my own understanding.

zackverham: @claude can you respond on PR comments? Can you explain what this pattern is doing?

extensions/vscode/src/publish/cloudContentRequest.ts
```diff
@@ -0,0 +1,483 @@
...
+} from "@posit-dev/connect-cloud-api";
+import { ContentType } from "../api/types/configurations";
+import type {
+  ConfigurationDetails,
+  ConnectCloudAccessControl,
+} from "../api/types/configurations";
+import type { ConnectRuntime } from "../api/types/connect";
+import { logger } from "../logging";
+
+// ---------------------------------------------------------------------------
```
zackverham, line 24: minor observation: the header comment here is maybe unnecessary, we can tell what the type is without it

christierney: claude seems to love these

extensions/vscode/src/views/tsDeployProgress.test.ts
christierney: done!

extensions/vscode/src/publish/connectCloudPublish.ts
```diff
@@ -0,0 +1,715 @@
+// Copyright (C) 2026 by Posit Software, PBC.
+
+import * as path from "path";
```
zackverham, line 3: maybe a question for @dotNomad  - is there any standard around import ordering in typescript? 

I only think about this because @christierney is very diligent about catching when I flub up the import ordering in go code ;)

extensions/vscode/src/publish/connectCloudPublish.ts
```diff
@@ -0,0 +1,715 @@
...
+    // Step 5: Upload bundle
+    lastStep = "uploadBundle";
+    onProgress({ step: "uploadBundle", status: "start" });
+    onProgress({
+      step: "uploadBundle",
+      status: "log",
+      message: "Uploading files",
+    });
+
+    if (!content.next_revision) {
```
zackverham: I'm assuming this is ordered the way it is to match the go implementation - but we could throw this error earlier, right? As soon as we have the `content` object.

extensions/vscode/src/publish/publishShared.ts
```diff
@@ -0,0 +1,470 @@
+// Copyright (C) 2026 by Posit Software, PBC.
```
zackverham, line 1: should we add unit test coverage for the functions in this file?

extensions/vscode/src/publish/connectCloudPublish.ts
christierney: This does match Go but I think you have a good point. I'll move it earlier.

extensions/vscode/src/publish/publishShared.ts
```diff
@@ -0,0 +1,470 @@
...
+  bundleId?: string;
+  bundleUrl?: string;
+  files?: string[];
+  requirements?: string[];
+  renv?: Record<string, unknown>;
+  deployedAt?: string;
+  dismissedAt?: string;
+  deploymentError?: { code: string; message: string; operation: string };
+  connectCloud?: { accountName: string };
+};
```
dotNomad, line 112: For the future, not this PR. It would be helpful to distinguish between a pre-deployment record and a deployment record like we did previously so we know when, say, `dashboardURL` is available or not.

extensions/vscode/src/publish/cloudContentRequest.ts
```diff
@@ -0,0 +1,483 @@
...
+// Credential Info Type
+// ---------------------------------------------------------------------------
+
+export interface CloudCredentialInfo {
+  accountId: string;
+  accountName: string;
+  environment: CloudEnvironment;
+}
```
dotNomad, line 32: This looks to be very close to our `Credential` type, or least the sections that matter for Cloud. Can we use `Pick<>` or something else to keep these related?

extensions/vscode/src/publish/publishShared.ts
christierney: They're all essentially helpers that get tested by way of the higher-level publishing tests. (And some of them are tested directly in roundtrip.test.ts, I just didn't move them.)

extensions/vscode/src/publish/cloudContentRequest.ts
dotNomad: These aren't necessary with strict `enum` typing really. I mean they are _fine_, but a bit confusing. With `CloudEnvironment` being an `enum`:
- if we add a key with a `default` case then `getCloudUIURL` will fail TypeScript compilation because it won't always return a `string`
- if we remove a key it will fail TypeScript compilation with `Property 'X' does not exist on type 'typeof CloudEnvironment'`

So I think we can remove these.

extensions/vscode/src/publish/cloudContentRequest.ts
```diff
@@ -0,0 +1,483 @@
...
+    case ContentAccess.ViewPrivateEditPrivate:
+    case ContentAccess.ViewPublicEditPrivate:
+      return "disabled";
+    case ContentAccess.ViewTeamEditPrivate:
+      return "viewer";
+    // Diverges from Go (content_request_base.go:67-68), which groups
+    // ViewPublicEditTeam with ViewTeamEditPrivate under "viewer". That's
+    // lossy: on redeploy with only publicAccess set, org=editor silently
+    // downgrades to org=viewer. We return "editor" to preserve the
+    // original setting.
```
dotNomad, line 174: Thanks for the diverges comment. I was a bit confused here.

extensions/vscode/src/publish/cloudContentRequest.ts
christierney: I think `Pick` doesn't work here because `environment` is an enum (whereas it's just a `string` on `Credential`)?

extensions/vscode/src/publish/connectCloudPublish.ts
dotNomad: Not really. There is `eslint-plugin-import` that has a `import/order` rule to sort them: https://github.com/import-js/eslint-plugin-import

We don't use that or enforce anything.

Typically, and what you'll see from the default [`import/order](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md)` rule is that there is a grouping of built in, external, and the internal group. So putting the built in `path` import at the top makes sense.

extensions/vscode/src/publish/cloudContentRequest.ts
christierney: I poked at this a bit but I think I'm inclined to leave it and pursue type consolidation / cleanup separately if that works for you.

dotNomad: Totally fine with that 👍

extensions/vscode/src/publish/connectCloudPublish.ts
```diff
@@ -0,0 +1,718 @@
...
+
+export type ConnectCloudPublishOptions = {
+  api: ConnectCloudAPI;
+  projectDir: string;
+  saveName: string;
+  config: ConfigurationDetails;
+  configName: string;
+  serverType: ServerType;
+  credential: CloudCredentialInfo;
+  existingContentId?: string;
```
dotNomad: We have to use a type assertion below (`contentId as ContentID`). Should `existingContentId` be `ContentID`?

extensions/vscode/src/publish/connectCloudPublish.ts
```diff
@@ -0,0 +1,718 @@
...
+
+  /** Check if canceled; if so, write dismissedAt and throw CanceledError. */
+  async function throwIfCanceled(): Promise<void> {
+    if (signal?.aborted) {
+      record.dismissedAt = new Date().toISOString();
+      record.deploymentError = undefined;
+      try {
+        await writePublishRecord(deploymentPath, record);
+      } catch {
+        // Don't mask cancellation
```
dotNomad, line 165: Aren't we masking cancellation here by not re-throwing?

I guess we `throw new CanceledError()` below, but not sure what is going on here.

extensions/vscode/src/publish/connectCloudPublish.test.ts
```diff
@@ -0,0 +1,914 @@
...
+      },
+      fileCount: 1,
+      totalSize: 100,
+    });
+  }),
+}));
+
+vi.mock("../logging", () => ({
+  logger: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}));
```
dotNomad: Don't have to do this with https://github.com/posit-dev/publisher/blob/660f2be4582dbcaa540e09a9896fe6c6c91a72cd/extensions/vscode/src/__mocks__/logging.ts.

If we can make more `__mocks__` like this we can if we are remaking these a lot.

dotNomad: Could be a good thing to put in the `CLAUDE.md` too, to look for `__mocks__` when using `vi.mock`

extensions/vscode/src/publish/connectCloudPublish.test.ts
```diff
@@ -0,0 +1,914 @@
...
+    vi.useFakeTimers();
+    vi.clearAllMocks();
+    // Reset watchCloudLogs to default — clearAllMocks doesn't reset
+    // implementations, so a mockImplementation override could leak.
+    vi.mocked(watchCloudLogs).mockResolvedValue(undefined);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
```
dotNomad: Since we use fake timers for all of them, why not an `afterAll()`?

extensions/vscode/src/publish/connectCloudPublish.test.ts
```diff
@@ -0,0 +1,914 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe("connectCloudPublish", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
```
dotNomad, line 214: Could put this in a `beforeAll` as well

dotNomad (approved):
Nothing I caught needs to be fixed in this PR just a couple things that could be tidied up.

I did have one question about error handling which would be good to get insight into.

extensions/vscode/src/views/homeView.ts
dotNomad, line 344: Can we get rid of the `initiateGoDeployment` block?

extensions/vscode/src/publish/connectCloudPublish.ts
christierney: To cancel we:

1. try to write the publish record to record the cancellation time
2. throw new CanceledError()

If 1 throws an exception that we don't catch & swallow, the publish will fail with a file write error instead of showing as cancelled. Basically we don't really care if this writePublishRecord fails for some reason as it is not critical.

extensions/vscode/src/publish/connectCloudPublish.ts
christierney: good idea

extensions/vscode/src/views/homeView.ts
christierney: I'm going to do a second PR right after this lands to remove all the Go stuff. I felt like this one was getting too big.

PR #4010, opened 2026-04-16
Fix dev version check blanking Source for CRAN-like packages

zackverham:
## Summary
- Removes `isDevVersion` check entirely for CRAN-like packages in `rLibraryMapper.ts`
- Also removes the now-unused `findAvailableVersion` and `packageVersion` helper functions

## Root cause
The `isDevVersion` check blanked out `Source` and `Repository` for CRAN-like packages whose installed version was newer than what the queried mirror returned. This caused `libraryToManifestPackages` to throw:

> cannot re-install packages installed from source; all packages must be installed from a reproducible location such as a repository. Package renv, Version 1.2.2

Diagnosed via #4009: in CI, `renv::init()` installs renv 1.2.2 (current CRAN release), but `listAvailablePackages` queries PPM (`https://packagemanager.posit.co/cran/latest`) which only has 1.2.1. The `isDevVersion` check treats 1.2.2 > 1.2.1 as a "dev version" and blanks the Source, which then fails validation.

## Why full removal is correct
The `isDevVersion` check only fires inside the `Source: "Repository"` + `isCRANLike(repository)` code path. Real dev versions (GitHub, GitLab, local installs) have `Source: "GitHub"` / `Source: "Local"` etc. and hit different switch branches. r-universe packages aren't CRAN-like. So this check can **only** produce false positives from mirror sync lag — never catch an actual dev version.

This was independently confirmed by the Go codebase: the same bug hit Windows ARM runners (where binary repos had newer versions than `available.packages(type="source")` returned), and the fix on branch `fix-windows-arm-renv-dev-version` was identical — remove `isDevVersion` entirely. That Go fix hasn't merged yet, but the TypeScript port copied the code before it landed.

## Test plan
- [x] Updated unit tests: dev version scenarios now expect successful resolution with CRAN source
- [x] Updated `libraryToManifestPackages` test: dev version now succeeds instead of throwing
- [ ] CI interpreter-integration tests should pass for all R configurations

### DIFF
--- a/extensions/vscode/src/publish/rLibraryMapper.test.ts
+++ b/extensions/vscode/src/publish/rLibraryMapper.test.ts
@@ -498,30 +498,6 @@ describe("library mapper error cases", () => {
     expect(description["Version"]).not.toBe(pkg.Version);
   });
 
-  test("CRAN package newer than mirror still resolves (toManifestPackage)", () => {
-    const pkg: RenvPackage = {
-      Package: "mypkg",
-      Version: "1.2.3",
-      Source: "Repository",
-      Repository: "CRAN",
-    };
-    const available: AvailablePackage[] = [
-      {
-        name: "mypkg",
-        version: "1.0.0", // installed is newer
-        repository: "https://cran.rstudio.com",
-      },
-    ];
-    const result = toManifestPackage(
-      pkg,
-      [{ Name: "CRAN", URL: "https://cran.rstudio.com" }],
-      available,
-      [],
-    );
-    expect(result.Source).toBe("CRAN");
-    expect(result.Repository).toBe("https://cran.rstudio.com");
-  });
-
   test("missing DESCRIPTION throws with renv::restore() guidance", async () => {
     await expect(readPackageDescription("mypkg", [])).rejects.toThrow(
       "consider running renv::restore()",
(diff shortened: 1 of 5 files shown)


zackverham:
@claude can you fix the test naming suggestion provided in the code review on this PR?

karawoo (commented):
Looking at the `isDevVersion` code I am actually uncertain whether we need this function (and `findAvailableVersion`, `packageVersion`) at all? It seems basically designed to generate false positives like this. Real dev versions (e.g. packages installed from github) would not have `Source: Repository` at all. Dev versions from r-universe might but are still not considered `CRANLike` in this context because the repository is not `CRAN`/`RSPM`/`P3M`/`PPM`.

extensions/vscode/src/publish/rLibraryMapper.test.ts
```diff
@@ -495,7 +498,7 @@ describe("library mapper error cases", () => {
     expect(description["Version"]).not.toBe(pkg.Version);
   });
 
-  test("dev version produces empty Source", () => {
+  test("CRAN package newer than mirror still resolves (toManifestPackage)", () => {
```
karawoo: this is in "error cases" but it's not really an error case anymore and is basically a duplicate of the one up at line 72. I think we could just remove

karawoo (approved):

extensions/vscode/src/publish/rLibraryMapper.test.ts
zackverham: @claude can you remove this test?

dotNomad (approved):
This makes sense to me as a work around 👍

PR #4012, opened 2026-04-16
Remove Go publish path and delete ~40k lines of dead code

christierney:
## Summary

- Remove the Go deployment fallback from the extension — all publishing now goes through the TypeScript `connectPublish()` orchestrator
- Remove the Go `POST /api/deployments/{name}` and `POST /api/deployments/{name}/cancel/{localid}` API routes
- Delete ~40,000 lines of dead Go code (excluding vendor/) across 248 files, found via iterative `deadcode` analysis
- Clean up TS/Go naming distinctions (`runTsDeployWithProgress` → `runDeployWithProgress`, `tsDeployProgress.ts` → `deployProgress.ts`)
- Add `try/finally` around `initiateDeployment` to guarantee `PUBLISH_INIT` (deploy button reset) is sent on all error paths

### Go packages fully deleted

`api_client/`, `bundles/`, `clients/` (all subpackages), `config/`, `contenttypes/`, `deployment/`, `executor/`, `inspect/`, `interpreters/`, `publish/` (all subpackages), `schema/`, `state/`, `util/dcf/`

### Go packages trimmed

- `events/` — removed `New()`, all publish Operation constants, dead Phase constants, unused emitter implementations
- `types/` — removed `EventableError` interface, `Optional[T]`, all `Null*` type aliases, `api_errors.go`, dead `ErrorCode` constants
- `util/` — removed `files.go`, `http.go`, `process.go`, `random.go`, `symlink_walker.go`, dead functions from `fspath.go` and `urls.go`
- `services/api/` — removed `BadRequest`, `ProjectDirFromRequest`, dead helpers; relocated credential error types

### What remains in Go

The Go backend now serves only: credential CRUD (3 routes), SSE event streaming (1 route), and binary startup. Five routes total in `api_service.go`.

## Test plan

- [x] `go build ./...` passes
- [x] `go test -short ./...` — all tests pass
- [x] `deadcode ./...` — only test-infrastructure logging functions remain unreachable
- [x] Pre-commit hooks pass (tsc, eslint, vitest, prettier)
- [x] Manual test: deploy to Connect via the extension (TS path)
- [x] Manual test: edit config to be invalid, click deploy — button resets, error shown
- [x] Manual test: credential creation and deletion still work
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4020, opened 2026-04-16
Remove Go SSE server and EventSource connection

christierney:
## Summary

- Refactor `EventStream` from `Readable` + `EventSource` to a pure in-memory `EventEmitter` — all publish events are now injected synthetically via `injectMessage()`
- Remove Go SSE route (`GET /api/events`), SSE infrastructure (emitter, handler, logger, mocks, error codes), and `r3labs/sse/v2` dependency
- Simplify Go `ui.go` to use `NewCLILogger` instead of the SSE multi-handler
- Delete dead `MultiHandler` and SSE-only logging constants
- Remove `eventsource` npm dependency from extension and homeView webview (unused there too)

**-3,288 lines** across 51 files. The Go API server now only serves the 3 credential CRUD routes.

Stacked on #4012.

## Test plan

- [x] `go build ./...` passes
- [x] `go test -short ./...` passes
- [x] `go vet ./...` passes
- [x] `npx tsc --noEmit` passes
- [x] `npm run test-unit` — all 1689 tests pass
- [x] `npx eslint` clean on modified files
- [x] `deadcode ./...` shows no new production-dead code
- [ ] E2E tests in CI
- [ ] Manual smoke test: deploy via TS publish path, verify log streaming works


cmd/publisher/commands/ui.go
```diff
@@ -20,14 +21,8 @@ type UICmd struct {
 
 func (cmd *UICmd) Run(args *cli_types.CommonArgs, ctx *cli_types.CLIContext) error {
 	ctx.Logger.Info("Starting PublishUICmd.Run")
```
christierney, line 23: all these changes are sort of irrelevant since this code is going away imminently, but I was keeping the removal targeted at just SSE code so it needed to replace this with another logger instance

zackverham (approved):

PR #4027, opened 2026-04-17
Fix race where deploy uses config without interpreter defaults

christierney:
## Summary

Fixes first deploy after creating a new deployment failing with "Invalid manifest.json" (missing R version in manifest, missing TOML files from bundle).

**Root cause:** `newDeployment()` writes the config TOML to disk, then calls `updateFileList()` twice to add the config and deployment TOML paths to the `files` list. But `updateFileList()` re-reads from disk, modifies, and writes back — it does not update the in-memory `Configuration` object that `newDeployment()` already captured. That stale object (missing both TOML file paths, missing interpreter defaults like `r.version`) gets pushed directly into the state cache by `showNewDeploymentMultiStep()`. The file watcher will eventually refresh the cache from disk, but if the user clicks Deploy before that completes, `initiateDeployment()` gets the stale config via `findValidConfig()`.

**Fix:** `initiateDeployment()` now reads the config fresh from disk via `loadConfiguration()` and applies interpreter defaults, instead of using the in-memory cache. This guarantees the config reflects the current TOML state regardless of file watcher timing. The stale cache push in `showNewDeploymentMultiStep()` is kept for UI responsiveness — the watcher replaces it shortly after.

- New `loadFreshConfig()` private method on `HomeViewProvider` handles disk read + interpreter hydration + error handling
- `findValidConfig()` warning comment updated to note the broader staleness risk and recommend `loadConfiguration()` for correctness-sensitive paths

## Test plan

- [x] Create a new R Shiny deployment and click "Deploy your project" immediately — should succeed with correct `r.version` in manifest and both TOML files in bundle
- [x] Create a new Python deployment and deploy immediately — should succeed with correct `python.version`
- [x] Deploy an existing deployment (not newly created) — should continue working as before
- [x] Edit a config TOML and deploy before the loading spinner finishes — should use the saved config from disk, not stale cache
- [x] Deploy with a missing or invalid config file — should show appropriate error message

### DIFF
--- a/extensions/vscode/src/state.ts
+++ b/extensions/vscode/src/state.ts
@@ -378,6 +378,12 @@ export class PublisherState implements Disposable {
     return findConfiguration(name, projectDir, this.configurations);
   }
 
+  // WARNING: the returned config may have empty interpreter fields
+  // (r.version, python.version, etc.) if the file watcher recently
+  // re-read the TOML from disk but the async UpdateConfigWithDefaults
+  // hydration hasn't completed yet. Callers that need interpreter
+  // fields for correctness (e.g. publish) must re-apply
+  // UpdateConfigWithDefaults before using them.
   findValidConfig(name: string, projectDir: string) {
     return findConfiguration(name, projectDir, this.validConfigs);
   }
--- a/extensions/vscode/src/state.ts
+++ b/extensions/vscode/src/state.ts
@@ -378,12 +378,11 @@ export class PublisherState implements Disposable {
     return findConfiguration(name, projectDir, this.configurations);
   }
 
-  // WARNING: the returned config may have empty interpreter fields
-  // (r.version, python.version, etc.) if the file watcher recently
-  // re-read the TOML from disk but the async UpdateConfigWithDefaults
-  // hydration hasn't completed yet. Callers that need interpreter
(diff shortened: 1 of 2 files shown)


christierney:
Tricky to test since it's a race but I was able to reproduce it pretty consistently before by creating a new deployment and clicking "deploy" real fast. Now I can't reproduce it.

zackverham (commented):
LGTM - if its a weird race condition it does seem hard to test. 

I suppose we could try to deploy many many many times in a row and confirm nothing breaks? But that feels kind of unnecessary.

zackverham (approved):

christierney:
@zackverham I did some deeper investigation and figured out my initial understanding of the bug was incorrect---I have a new fix now, please rereview.

christierney:
The more deployments you have, the easier this is (was) to reproduce---because the file watcher update had to launch R and Python processes for every single config and wait for them to return, I ended up with like 10+ seconds in which to try deploying my new config. So I got to where I could reproduce it every time. With this latest fix I can no longer reproduce it so I'm confident it is fixed.

homeView.ts still has no unit tests---a known gap.

dotNomad (approved):
Great catch. The fix here makes sense to me 🫡

PR #4031, opened 2026-04-17
Add diagnostic bundle command for support troubleshooting

### LINKED ISSUE #2325: Diagnostic Bundle
I realized we don't have a good way to get Support all the items that might help them diagnose an issue.  I think it'd be great to add a button to save logs from various places under the Help and Feedback section.

Ideally it would collect...
1. Posit Publisher Log
2. Output > Posit Publisher logs
3. If possible the Environment Restore and App Logs from Connect (I don't think there is an API for these in Connect though).
4. Optionally package the app they're deploying, maybe we could ask if they'd like us to include it?
5. VSCode Version and Posit Publisher Version

zackverham:
Adds a "Save Diagnostic Bundle" command under Help and Feedback that collects system info (extension version, IDE, platform), language runtime versions (Python, R, Quarto), and workspace details into a text file.

Closes #2325

Generated with [Claude Code](https://claude.ai/claude-code)


extensions/vscode/src/commands/diagnosticBundle.ts
```diff
@@ -0,0 +1,156 @@
...
+    `Python: ${info.pythonVersion}`,
+    `R: ${info.rVersion}`,
+    `Quarto: ${info.quartoVersion}`,
+    "",
+    "## Workspace",
+    ...info.workspaceFolders.map((f) => `- ${f}`),
+    "",
+    "## Extension Log",
+    "(Open the Posit Publisher output channel for full logs:",
+    ' Run "Posit Publisher: Show Debug Log" from the Command Palette)',
```
christierney, line 151: is this not something we can get programmatically?

PR #4030, opened 2026-04-17
Decommission Go backend

dotNomad:
## Summary

Closes #3818. Finishes the Go-to-TypeScript migration: removes the credential dual-write to the Go backend and deletes the rest of the Go codebase. Publisher is now a pure TypeScript extension.

VSIX size drops from ~10.6 MB to ~2.1 MB (the Go binary was ~8 MB).

## Commits

1. **Stop spawning the Go binary from the extension** (`b3ce536d`) — Delete `src/services.ts`, `src/servers.ts`, binary-spawn helpers in `src/commands.ts`; remove `Service` usage from `extension.ts`; rewire `ShowOutputChannel` to `logger.show()`; stop copying `bin/publisher` during packaging.
2. **Delete Go codebase and replace build/CI with TypeScript-only tooling** (`745130d2`) — Remove `cmd/`, `internal/`, `vendor/`, `go.mod`, `go.sum`, and Go-only scripts (798 files, ~358k line deletions — most from `vendor/`). Rewrite root `Justfile` to TypeScript-only targets. Rewire `extensions/vscode/Justfile` with local `_os`/`_arch` helpers and `jq` for name resolution. Delete Go-only GitHub workflows (`agent.yaml`, `archive.yaml`, `build.yaml`, `upload.yaml`); strip Go steps from mixed workflows; restructure `package.yaml` to a 6-cell `(darwin|linux|windows) × (amd64|arm64)` matrix that merges into a single `dist` artifact.
3. **Delete dead src/ports.ts and drop get-port dependency** (`152962d1`) — Module only acquired ports for the Go subprocess; now unreferenced.
4. **Restore pure-TypeScript docs and launch configs after rebase** (`82881eea`) — Rewrite `CLAUDE.md` / `extensions/vscode/CLAUDE.md` / `CONTRIBUTING.md` / `extensions/vscode/CONTRIBUTING.md` / `.gitignore` / both `launch.json` files to drop Go-era content; add `Removed` entry to CHANGELOG for #3818.

## Follow-up issues

- #4025 — Drop platform-specific VSIX packaging (`vsce --target`). With the Go binary gone, a single universal VSIX works everywhere.
- #4026 — Update the `dev-setup` skill for the pure-TypeScript project.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm run lint` clean
- [x] Unit tests: 1669 pass / 4 todo
- [x] `just test` (Mocha integration + Vitest): all pass
[...]

PR #4044, opened 2026-04-17
Drop platform-specific VSIX packaging (#4025)

### LINKED ISSUE #4025: Drop platform-specific VSIX packaging (vsce --target)
Sub-issue of #3713.

Now that the Go binary is gone (#3818 et al.) and the extension is pure JavaScript/TypeScript, there's no reason to publish a separate VSIX per platform. We should drop `vsce --target` and ship a single universal `.vsix` that works on every platform VS Code runs on.

## Why it existed

The extension previously bundled `bin/publisher` — a Go binary built per platform (`darwin-amd64`, `darwin-arm64`, `linux-amd64`, `linux-arm64`, `win32-x64`, `win32-arm64`). `vsce --target` was required so each VSIX would contain only the matching binary.

## Why it's no longer needed

Per the [VS Code extension docs](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platform-specific-extensions):

> Platform-specific extensions are useful if your extension has platform-specific libraries or dependencies, so you can control the exact binaries that are included in a platform package. A common use case is the use of **native node modules**.

All of our runtime dependencies are now pure JavaScript:

- `@posit-dev/connect-api`, `@posit-dev/connect-cloud-api`
- `@vscode/codicons`
- `ajv`, `ajv-formats`
- `async-mutex`, `axios`, `debounce`, `filenamify`, `js-yaml`, `jsonwebtoken`, `picomatch`, `retry`, `smol-toml`, `tar-stream`, `vscode-uri`

No `.node` bindings are included in the built VSIX. The only native modules in `node_modules` (`rollup`, `fsevents`, `ttf2woff2`) are dev-only and never bundled by esbuild into `dist/extension.js`.

### `extensions/vscode/Justfile`
- Delete `_os`, `_arch`, `package-target-platform` helper recipes.
- Simplify `default`, `install`, `package`, `package-name` to take no `os`/`arch` arguments.
- `package-name` becomes `$name-$version.vsix` (drop the `-{os}-{arch}` suffix).
- `package` recipe: drop the `-t "$target_platform"` arg from the `vsce package` invocation.

### `.github/workflows/package.yaml`
- Collapse the 6-cell `(darwin|linux|windows) × (amd64|arm64)` matrix into a single job.
- Delete the `merge-artifacts` job — a single `dist/` upload is all downstream consumers need.

### `.github/workflows/publish.yaml`
- Simplify to publish one VSIX instead of iterating over platform-specific files.

### `.github/workflows/release.yaml`
- Attach a single `.vsix` to the release artifacts instead of six.

### Verification
- VSIX installs and activates on darwin-arm64 (local smoke test).
- CI package job produces one VSIX.
- `vsce publish` from a dry-run or pre-release flow succeeds with a single file.

dotNomad:
## Summary
- Ship one universal `publisher-$version.vsix` now that the extension is pure TypeScript (follow-up to #4030).
- Collapse the `(os, arch)` matrix in `package.yaml` and `publish.yaml`; drop the `merge-artifacts` job.
- Strip `grep linux-amd64` from e2e and Workbench VSIX lookups — they'd otherwise fail to find the renamed file.

Resolves #4025.

## Test plan
- [x] `just` from the repo root produces exactly one `dist/*.vsix` with no platform suffix (verified locally: `publisher-1.37.19-11-gf905712e.vsix`).
- [x] e2e helper lookup pattern resolves to that single file.
- [x] CI `package` workflow uploads a single `dist` artifact containing one VSIX.
- [x] CI e2e workflow locates and installs the VSIX.
- [ ] Next release (or `vsce publish` dry run) resolves `extension_path` to the single file.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -174,7 +174,7 @@ jobs:
         working-directory: test/e2e
         run: |
           # Install the Publisher extension and set ownership in parallel
-          VSIX_FILENAME=$(ls -Art ../../dist | grep linux-amd64 | tail -n 1)
+          VSIX_FILENAME=$(ls -Art ../../dist/*.vsix | xargs -n1 basename | tail -n 1)
           docker compose exec code-server code-server --install-extension "/home/coder/vsix/${VSIX_FILENAME}" &
 
           # Set proper ownership for workspace files while extension installs
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -6,22 +6,6 @@ permissions:
 jobs:
   vscode:
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - os: darwin
-            arch: amd64
-          - os: darwin
-            arch: arm64
-          - os: linux
-            arch: amd64
-          - os: linux
-            arch: arm64
-          - os: windows
-            arch: amd64
-          - os: windows
-            arch: arm64
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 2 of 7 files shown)


christierney (approved):
🦅 

love it

PR #4068, opened 2026-04-22
fix: update positron-server binary path for new Workbench image

zackverham:
## Summary

- Fix e2e test failures caused by the Workbench Docker image (`rstudio/rstudio-workbench:ubuntu2204`) moving the `positron-server` binary from `bin/positron-server` to `bundled/bin/positron-server`
- Updates the hardcoded path in `install-positron-extension.sh` to match the new image layout
- Confirmed the same path change applies to both the release and preview Workbench images

## Context

All four e2e test jobs (preview, 2023.03.0, 2025.03.0, release) started failing with:

```
bash: line 1: ./bin/positron-server: No such file or directory
```

This is an upstream change in the Workbench Docker image — the failing commit only modified an unrelated notification workflow.

## Test plan

- [ ] E2e tests pass in CI with the updated path


dotNomad (approved):
Thanks for doing this 🚀

PR #4072, opened 2026-04-22
Fix external issue notification workflow

zackverham:
## Summary
- **Wrong team slug**: The workflow used `publisher` as the team slug, but the actual team is `publisher-code-owners`. This caused the GitHub API to return 404 for every user, making everyone appear external.
- **Wrong status code check**: The `GET /orgs/{org}/teams/{team_slug}/members/{username}` endpoint returns `204` for members, not `200`. Even with the correct team, the check would have failed.
- **Broken Slack formatting**: Literal `\n` in bash strings weren't rendered as newlines in Slack messages. Switched to `printf` for proper formatting.

## Test plan
- [ ] Verify an internal team member commenting on an external issue does **not** trigger a Slack notification
- [ ] Verify an external user opening a new issue **does** trigger a Slack notification with correct formatting
- [ ] Verify an external user commenting on an issue **does** trigger a Slack notification with correct formatting

### DIFF
--- a/.github/workflows/external-issue-notify.yaml
+++ b/.github/workflows/external-issue-notify.yaml
@@ -27,13 +27,13 @@ jobs:
           GH_TOKEN: ${{ secrets.ORG_TEAM_READ_TOKEN }}
         run: |
           # Check if the author is a member of posit-dev/publisher
-          # The API returns 200 if they are a member, 404 if not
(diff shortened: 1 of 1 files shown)


dotNomad (approved):
Fantastic catches and fixes 🎉

PR #4083, opened 2026-04-24
Fix Connect Cloud redeploy wiping all secrets

### LINKED ISSUE #4082: Publishing to Connect Cloud deletes secrets by sending `secrets: []` on PATCH
## Bug Description

When publishing a new version of content to Posit Connect Cloud, all secrets on the content are deleted. This happens because the PATCH request payload always includes a `secrets` field, even when secrets haven't changed.

## Root Cause

In [`extensions/vscode/src/publish/cloudContentRequest.ts` at line 407](https://github.com/posit-dev/publisher/blob/d8215f91774ab8a00bab9b214b2894afa50986d5/extensions/vscode/src/publish/cloudContentRequest.ts#L407), a value for `secrets` is always included in the request payload. If the user made no changes to secrets, it sends `secrets: []`, which tells Connect Cloud to remove all secrets from the content.

Connect Cloud update endpoints are all PATCH — if a value is not changing, it should **not** be included in the payload. Sending `secrets: []` is interpreted as "there should be no secrets on this content."

## Expected Behavior

If secrets have not been modified by the user, the `secrets` field should be omitted from the PATCH payload entirely, so existing secrets are preserved.

## Steps to Reproduce

1. Deploy content to Connect Cloud with secrets configured
2. Publish a new version of the content via Posit Publisher without modifying secrets
3. Observe that all secrets are deleted from the content on Connect Cloud

m--:
## Summary

Fixes #4082.

When redeploying content to Connect Cloud, all secrets were being deleted because the PATCH request always included `secrets: []` when no secrets were provided. Connect Cloud interprets this as "remove all secrets."

- Changed `buildSecrets` to return `undefined` instead of `[]` when neither `config.environment` nor the `secrets` parameter has any entries
- When `secrets` is `undefined`, it is omitted from the PATCH payload (JSON serialization drops `undefined` fields), preserving existing secrets on the content
- Added test coverage for the missing cases: omitting secrets on update when none provided, including secrets when provided, and including env vars from config

## Test plan

- [x] Unit tests pass (`npm run test-unit`)
- [x] Manual test: deploy content to Connect Cloud with secrets configured, redeploy without modifying secrets, verify secrets are preserved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed Connect Cloud redeploy wiping previously set secrets. (#4082)
+
 ## [2.0.0]
 
 ### Added
--- a/extensions/vscode/src/publish/cloudContentRequest.test.ts
+++ b/extensions/vscode/src/publish/cloudContentRequest.test.ts
@@ -831,6 +831,26 @@ describe("buildUpdateContentRequest", () => {
     expect(request.secrets).toBeUndefined();
   });
 
+  it("omits secrets when secrets param is empty object (redeploy without new values)", () => {
+    const config: ConfigurationDetails = {
+      $schema: "https://example.com/schema",
+      productType: ProductType.CONNECT_CLOUD,
+      type: ContentType.R_SHINY,
+      entrypoint: "app.R",
+      validate: true,
+    };
+
+    const request = buildUpdateContentRequest(
+      config,
+      "my-app",
+      {},
+      ContentID("content-456"),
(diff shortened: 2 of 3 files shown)


christierney (approved):

christierney:
I reproduced the reported bug, and then confirmed this fix.

Before:

1. deploy app with a secret set
2. result: app has a secret variable
3. redeploy without changing anything
4. result: secret variable is gone

After:

1. deploy app with a secret set
2. result: app has a secret variable
3. redeploy without changing anything
4. result: secret variable is still there

One thing to note: if you configure an `[environment]` section (non-secret vars), or if you add a new secret but don't re-set your existing secret, then your existing secret will still get dropped. Publisher does not remember your secrets after deploying with them.

PR #4093, opened 2026-04-28
Guard prepare-release against accidental version downgrades

dotNomad:
## Summary

The prepare-release workflow rewrites `extensions/vscode/package.json` to whatever version the operator types, with no monotonic check. A typo or a patch release of an older release line cut against current `main` silently rolls `package.json` backwards. This came up while preparing #4090 (`v2.0.1` patch), where the prepare-release flow happily set `package.json` from `2.1.1` (latest pre-release) down to `2.0.1`.

This PR adds a guard:

- `scripts/prepare-release.py` errors when the requested version is lower than the current `package.json` version
- A new `--allow-downgrade` flag overrides the guard, for genuine patch releases of older release lines
- The `prepare-release.yaml` workflow exposes the override as an opt-in `allow_downgrade` boolean input

While here, the script invocation is switched from `yes | ... || true` to `echo y | ...`. The old form swallows nonzero exits (originally to absorb `yes`'s SIGPIPE under bash's `pipefail`), which would have hidden the new guard's error and produced an empty release PR. `echo y` exits cleanly on its own without SIGPIPE, so genuine script failures now fail the workflow step.

## Test plan

- [x] `python3 scripts/test_prepare_release.py` — all 18 tests pass, including 3 new `TestDowngradeGuard` cases covering: downgrade rejected by default, downgrade allowed with `--allow-downgrade`, equal version allowed (not a downgrade)
- [x] YAML validates as a workflow file
- [ ] Reviewer sanity check: open Actions → Prepare Release; confirm `version` and `allow_downgrade` inputs render

## Out of scope

Patch-release tooling that automates branching from a prior tag and cherry-picking fixes — that will be tracked separately as a follow-up issue.

### DIFF
--- a/.github/workflows/prepare-release.yaml
+++ b/.github/workflows/prepare-release.yaml
@@ -7,6 +7,11 @@ on:
         description: "Version to release (e.g., 1.34.0) - must be even minor"
         required: true
         type: string
+      allow_downgrade:
+        description: "Allow setting package.json to a lower version (use only for patch releases of older release lines)"
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   prepare:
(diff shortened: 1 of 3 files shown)


zackverham (approved):
Nice! Yeah, this makes sense to me.

PR #4097, opened 2026-04-29
Improve error message when server is unreachable (#4096)

### LINKED ISSUE #4096: Improve error message when deployment fails due to VPN/network connectivity
## Problem

When deploying to a Posit Connect server that requires VPN access, if you're not connected to the VPN, deployment fails with "HTTP undefined". The linked troubleshooting docs don't mention VPN/network connectivity as a possible cause.

## Suggestion

Either:
- Improve the error message to indicate a network connectivity issue (e.g. "unable to reach server" instead of "HTTP undefined")
- Add VPN/network connectivity as a troubleshooting step in the linked docs
- Both

zackverham:
## Intent

Resolves #4096

When deploying to a Posit Connect server that requires VPN access (or is otherwise unreachable), deployment fails with the unhelpful message "HTTP undefined". This happens because axios throws an error with no `response` property for network-level failures, and `err.response?.status` evaluates to `undefined`.

## Type of Change

- [x] Bug Fix
- [x] Documentation

## Approach

Added network error detection (`isAxiosError(err) && !err.response`) in three places:

1. **`packages/connect-api/src/client.ts`** — `testAuthentication()` now throws a clear `ConnectAPIError("Unable to reach the server...")` instead of "HTTP undefined". Certificate errors (which also lack a `response`) are rethrown as-is so callers can classify them separately.
2. **`extensions/vscode/src/publish/connectPublish.ts`** — `classifyDeploymentError()` now returns `connectionFailed` for no-response axios errors, placed after the certificate check to preserve existing cert error classification.
3. **`extensions/vscode/src/publish/connectCloudPublish.ts`** — Same `connectionFailed` classification in `classifyCloudDeploymentError()`.

Also added a new troubleshooting section to `docs/troubleshooting.md` covering common causes of network connectivity errors (VPN, DNS, firewall, server down, incorrect URL).

## User Impact

Users who encounter network connectivity issues during deployment or credential testing will now see:

> Unable to reach the server. Check your network connection, VPN, and server URL.

instead of the opaque "HTTP undefined".

## Automated Tests

Added tests across all four affected test files:
- `packages/connect-api/src/client.test.ts` — network error produces clear message + `ConnectAPIError` with no `httpStatus`; certificate errors are not caught by the network check
- `extensions/vscode/src/publish/connectPublish.test.ts` — no-response error classifies as `connectionFailed`
- `extensions/vscode/src/publish/connectCloudPublish.test.ts` — same for Cloud deployments
- `extensions/vscode/src/utils/testCredentials.test.ts` — network error propagates through credential testing

## Directions for Reviewers

- The core fix is in `packages/connect-api/src/client.ts` lines 190-205 — verify the cert/network error branching logic
- Check that the troubleshooting doc section in `docs/troubleshooting.md` reads well
- All existing tests continue to pass (94 connect-api, 104 connectPublish, 33 connectCloudPublish, 20 testCredentials)

## Checklist


zackverham:
@claude can you make the improvement you noted under "minor suggestions" in your code review above

zackverham:
@claude can you add a changelog blurb for this PR

extensions/vscode/src/publish/connectCloudPublish.ts
```diff
@@ -671,6 +672,29 @@ function classifyCloudDeploymentError(
...
+  // Network errors (no response received) — server unreachable due to
+  // DNS failure, VPN disconnected, firewall, etc.
+  if (isAxiosError(err) && !err.response) {
+    return {
+      code: "connectionFailed",
+      message:
+        "Unable to reach the server. " +
+        "Check your network connection, VPN, and server URL.",
+    };
+  }
```
jonkeane, line 697: ❤️ seeing this split out since fixing them are sooo different, but we do have some information to get people to the right place

extensions/vscode/src/publish/connectCloudPublish.ts
```diff
@@ -671,6 +672,29 @@ function classifyCloudDeploymentError(
...
 
+  // Certificate verification errors (any step)
+  // Must be checked before the generic network error check below, because
+  // TLS failures also have no `response`.
+  if (isAxiosError(err) && isCertificateError(err)) {
+    return {
+      code: "errorCertificateVerification",
+      message:
+        "Certificate verification failed. " +
+        "Check the server's TLS certificate or disable verification.",
```
christierney: The Connect Cloud API client does not allow for disabling verification, so this advice should probably be adjusted. (Or we could add that ability to the cloud API, but I think it makes sense that we would not want to let the Cloud cert lapse so there should be no reason for it?)

christierney (approved):

extensions/vscode/src/publish/connectCloudPublish.ts
zackverham: hmm - what do you think would be a better message? 

> Certificate verification failed. If you're trying to publish to Connect Cloud, contact support <link> or open an issue <link>. If you're trying to publish to Connect, check the server's TLS certificate. 


Something like that?

christierney: Well this is only the Cloud path so we can tailor the advice to just Cloud. (I think the reason the Cloud path didn't have a check for this specific error originally is probably that we don't expect Cloud to have a bad cert, although anything can happen...)

christierney: So, probably their options are:
- file an issue
- contact support
- check their network (proxy in the way, etc.)

zackverham: ohhh I see what you're saying. `connectCloudPublish.ts` 🤦 

Yeah, Agree. 


@claude can you improve the error message here so that it tailors the advice to just cloud?

PR #4111, opened 2026-05-01
fix: Snowflake SPCS dual authentication

### LINKED ISSUE #3465: feat: fix dual auth for Snowflake SPCS (Connect API key + Snowflake token, and Connect token + Snowflake token)

christierney:
## Summary

Fix publishing to Connect in SPCS _from Publisher outside of SPCS_

- Adds dual-header authentication for Snowflake SPCS: `Authorization: Snowflake Token="..."` for the ingress proxy plus `X-RSC-Authorization: Key <apiKey>` or token-signing headers for Connect
- Updates the credential wizard to collect Snowflake connection first (needed to reach Connect through the proxy), then prompt for Connect auth method (API key or token)
- Legacy snowflake-only credentials remain loadable for display/deletion but will fail auth at request time 
 
Closes #3465

## Approach

Users publishing to Connect in SPCS from a Publisher outside of SPCS need to provide two forms of authentication: Snowflake and Connect. So, which should come first in the new credentials wizard? I opted to put Snowflake first to enable support for Connect token auth, which requires being able to get through the Snowflake ingress in order to reach Connect and set up the token.

- user enters a Snowflake server URL
- Publisher reads Snowflake connections.toml and narrows down the list of options by trying to connect to each one: if we get a 401, we assume the connection worked and Connect then rejected it for being unauthorized. This weeds out some forms of bad Snowflake config, and in any case I expect most users will only have one connection set up anyway, so I'm not too worried about bad connection configs getting through.
- User selects which Snowflake connection they want to use and then picks API Key or Token Auth. Credential setup proceeds as normal, except when we make requests to Connect we include a Snowflake token.

## Related

This PR also fixes a bug in wizard navigation: if for some reason a user completes token auth setup, but then navigates back and enters an API Key instead, this would previously have failed (tried to create a credential with both token and API key auth). Now we reset state so this failure doesn't happen.

Previously we were using `testCredential` in two places:
- server URL validation (expects authentication failure)
- auth validation

This seemed confusing to me, especially with the changes I had to make for snowflake, so I split this into two separate functions, one for each purpose.

## Test plan

- [x] Unit tests pass (1677 passed)
- [x] Manual: create new Snowflake credential with API key on dogfood SPCS instance
- [x] Manual: create new Snowflake credential with token auth on dogfood SPCS instance
- [x] Manual: verify legacy snowflake-only credential loads without crashing
- [x] Manual: verify Snowflake discovery works without Connect auth (401-as-success)
- [x] Manual: verify wizard back-navigation clears stale auth fields

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
@@ -75,6 +75,7 @@ export async function newConnectCredential(
     INPUT_SERVER_URL = "inputServerUrl",
     INPUT_API_KEY = "inputAPIKey",
     INPUT_SNOWFLAKE_CONN = "inputSnowflakeConnection",
+    INPUT_SNOWFLAKE_AUTH_METHOD = "inputSnowflakeAuthMethod",
     INPUT_CRED_NAME = "inputCredentialName",
     INPUT_AUTH_METHOD = "inputAuthMethod",
     INPUT_TOKEN = "inputToken",
@@ -87,6 +88,7 @@ export async function newConnectCredential(
     [step.INPUT_SERVER_URL]: inputServerUrl,
     [step.INPUT_API_KEY]: inputAPIKey,
     [step.INPUT_SNOWFLAKE_CONN]: inputSnowflakeConnection,
+    [step.INPUT_SNOWFLAKE_AUTH_METHOD]: inputSnowflakeAuthMethod,
     [step.INPUT_CRED_NAME]: inputCredentialName,
     [step.INPUT_AUTH_METHOD]: inputAuthMethod,
     [step.INPUT_TOKEN]: inputToken,
@@ -120,8 +122,12 @@ export async function newConnectCredential(
   };
 
   const isValidSnowflakeAuth = () => {
-    // for Snowflake, require snowflakeConnection
-    return isSnowflake(serverType) && isString(state.data.snowflakeConnection);
+    return (
+      isSnowflake(serverType) &&
+      isString(state.data.snowflakeConnection) &&
+      (isString(state.data.apiKey) ||
+        (isString(state.data.token) && isString(state.data.privateKey)))
+    );
   };
 
   // ***************************************************************
@@ -483,7 +489,19 @@ export async function newConnectCredential(
 
     try {
       await showProgress("Reading Snowflake connections", viewId, async () => {
-        const resp = await fetchSnowflakeConnections(serverUrl);
+        const connectAuth =
+          typeof state.data.apiKey === "string" && state.data.apiKey
+            ? { apiKey: state.data.apiKey }
+            : typeof state.data.token === "string" &&
+                typeof state.data.privateKey === "string" &&
+                state.data.token &&
+                state.data.privateKey
+              ? {
+                  token: state.data.token,
+                  privateKey: state.data.privateKey,
+                }
+              : undefined;
+        const resp = await fetchSnowflakeConnections(serverUrl, connectAuth);
         connections = resp.connections;
         connectionQuickPicks = resp.connectionQuickPicks;
       });
@@ -514,9 +532,57 @@ export async function newConnectCredential(
     }
 
     return {
-      name: step.INPUT_CRED_NAME,
+      name: step.INPUT_SNOWFLAKE_AUTH_METHOD,
       step: (input: MultiStepInput) =>
-        steps[step.INPUT_CRED_NAME](input, state),
+        steps[step.INPUT_SNOWFLAKE_AUTH_METHOD](input, state),
+    };
+  }
+
+  // ***************************************************************
+  // Step: Select Connect auth method (Snowflake SPCS only)
+  // ***************************************************************
+  async function inputSnowflakeAuthMethod(
+    input: MultiStepInput,
+    state: MultiStepState,
+  ) {
+    const authMethods = [
+      {
+        label: AuthMethodName.API_KEY,
+        description: "Enter a Connect API key",
+      },
+      {
+        label: AuthMethodName.TOKEN,
+        description: "One-click token authentication",
+      },
+    ];
+
+    const pick = await input.showQuickPick({
+      title: state.title,
+      step: 0,
+      totalSteps: 0,
+      placeholder:
+        "Select how to authenticate with Posit Connect (behind Snowflake)",
+      items: authMethods,
+      activeItem: authMethods[0],
+      buttons: [],
+      shouldResume: () => Promise.resolve(false),
+      ignoreFocusOut: true,
+    });
+
+    authMethod = getAuthMethod(pick.label as AuthMethodName);
+
+    if (isApiKey(authMethod)) {
+      return {
+        name: step.INPUT_API_KEY,
+        step: (input: MultiStepInput) =>
+          steps[step.INPUT_API_KEY](input, state),
+      };
+    }
+
+    return {
+      name: step.INPUT_TOKEN,
+      step: (input: MultiStepInput) => steps[step.INPUT_TOKEN](input, state),
+      skipStepHistory: true,
     };
   }
 
--- a/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
@@ -572,6 +572,8 @@ export async function newConnectCredential(
     authMethod = getAuthMethod(pick.label as AuthMethodName);
 
     if (isApiKey(authMethod)) {
+      state.data.token = undefined;
+      state.data.privateKey = undefined;
       return {
         name: step.INPUT_API_KEY,
         step: (input: MultiStepInput) =>
@@ -579,6 +581,7 @@ export async function newConnectCredential(
       };
     }
 
+    state.data.apiKey = undefined;
     return {
       name: step.INPUT_TOKEN,
       step: (input: MultiStepInput) => steps[step.INPUT_TOKEN](input, state),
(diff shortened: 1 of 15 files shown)


packages/connect-api/src/types.ts
```diff
@@ -44,22 +44,42 @@ interface TokenAuth extends ConnectAPIBaseOptions {
...
+  /** Base64-encoded DER PKCS#1 RSA private key for Connect token-based authentication. */
+  privateKey: string;
 }
 
 interface NoAuth extends ConnectAPIBaseOptions {
   apiKey?: never;
   token?: never;
   privateKey?: never;
-  snowflakeToken?: never;
+  /** Optional Snowflake token for legacy credentials without Connect auth.
```
zackverham, line 70: I guess eventually we'll want to fully deprecate this right? Do we need a strategy / plan for that (even if its just an issue)?

extensions/vscode/src/multiStepInputs/newConnectCredential.ts
```diff
@@ -100,6 +102,24 @@ export async function newConnectCredential(
     return authMethod === AuthMethod.API_KEY;
   };
 
+  async function getSnowflakeToken(
+    state: MultiStepState,
+  ): Promise<string | undefined> {
+    const connName = state.data.snowflakeConnection;
+    if (!isSnowflake(serverType) || typeof connName !== "string") {
```
zackverham, line 109: not for this PR - but using serverType as a global is a little confusing / hard to reason through, I'm not sure why exactly we have things set up that way.

extensions/vscode/src/multiStepInputs/newConnectCredential.ts
```diff
@@ -431,9 +463,11 @@ export async function newConnectCredential(
         const serverUrl =
           typeof state.data.url === "string" ? state.data.url : "";
         try {
-          const testResult = await testCredentials({
+          const snowflakeToken = await getSnowflakeToken(state);
+          const testResult = await testAuthentication({
             url: serverUrl,
             apiKey: input,
+            snowflakeToken,
```
zackverham, line 470: this reads a little odd that the parameter name isn't included as a prefix, like the other parameters here. Not sure if that's intentional or not.

extensions/vscode/src/utils/testCredentials.ts
```diff
@@ -96,49 +91,91 @@ export async function testCredentials(
...
+    return {
+      user: null,
+      url: null,
+      serverType: null,
+      error: toAgentError(err),
+    };
+  }
+
+  let lastUser: User | null = null;
+  const timeoutMs = Math.max(params.timeout ?? 30, 30) * 1000;
```
zackverham: should we document this max timeout value somewhere? Or at least store it as a named constant

zackverham (approved):

packages/connect-api/src/types.ts
christierney: eh, perhaps? It feels like a tough one when we don't have much insight into how often people are upgrading. The thing I want to avoid is anyone who created an old-style snowflake credential updating publisher and suddenly it crashes trying to load their saved creds. Realistically, there probably aren't many users out there like this...

extensions/vscode/src/multiStepInputs/newConnectCredential.ts
christierney: these wizards are full of globals like this and I agree that it's not great. I don't have a specific plan for refactoring them

extensions/vscode/src/multiStepInputs/newConnectCredential.ts
christierney: nope, agreed this is odd. I'll update.

christierney: actually no, this makes sense---we're initializing an object and in this case the variable name and the property name are identical, so expanding this would look like:

```
snowflakeToken: snowflakeToken,
```

It's ES6 shorthand syntax.

extensions/vscode/src/utils/testCredentials.ts
christierney: yeah I'll add a constant. I don't think we (publicly) document any of the other timeouts we have scattered around currently.

christierney: Nobody actually uses the ability to override the default, either---this is just dead code. I'll simplify and just use a constant and we can always add back the ability to override it if we want it later.

PR #4125, opened 2026-05-04
Add Claude permission allowlist for safe just/npm commands

dotNomad:
Adds a `permissions.allow` block to `.claude/settings.json` so Claude Code can run common build, test, lint, and format commands in this repo without prompting on each invocation. Scoped narrowly using explicit recipes only, with a few targeted wildcards.

See the Claude documentation on permissions: https://code.claude.com/docs/en/permissions

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -22,5 +22,41 @@
         ]
       }
     ]
+  },
+  "permissions": {
+    "allow": [
+      "Bash(just)",
+      "Bash(just -l)",
+      "Bash(just default)",
+      "Bash(just version)",
+      "Bash(just pre-release)",
+      "Bash(just clean)",
+      "Bash(just format)",
+      "Bash(just check-format)",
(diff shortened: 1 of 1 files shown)


.claude/settings.json
```diff
@@ -22,5 +22,41 @@
...
+      "Bash(just vscode configure)",
+      "Bash(just vscode package)",
+      "Bash(just vscode package-name)",
+      "Bash(just vscode package-path)",
+      "Bash(npm install)",
+      "Bash(npm run format)",
+      "Bash(npm run check-format)",
+      "Bash(npm test *)",
+      "Bash(npm run lint *)",
+      "Bash(npm run typecheck *)",
```
dotNomad, line 56: The wildcarding here is worth a double check. Everything else is really narrow.

zackverham (approved):
yeah this looks great, good call.

Issue #4123, opened 2026-05-04
docs: add nodejs content type to configuration documentation and CHANGELOG


dotNomad:
The final piece of the #3964 Node.js epic: update user-facing docs and the CHANGELOG now that everything ships.

## `docs/configuration.md`

1. Under `#### type`, add `nodejs` to the bullet list, alphabetically between `jupyter-voila` and `python-bokeh`.

2. Add a new `## Node.js settings` section after `## Quarto settings`, parallel in style to `## Python settings` / `## R settings`, covering:

   - **There is no `[node]` TOML section.** `type = "nodejs"` is the only marker — Node.js content has no user-configurable Publisher-side settings. This was a deliberate decision in #3996; don't introduce one here.
   - Both `package.json` and `package-lock.json` must be present in the project and included in `files`. Connect installs with `npm ci`, so a missing lock file is a hard failure — run `npm install` to generate one.
   - If `engines.node` is set in `package.json`, Connect uses it to pick the runtime version; otherwise Connect uses its default.
   - **Server gating:** requires Connect with an Advanced or Evaluation license, and the admin must enable Node.js support. Publisher's preflight surfaces a Node.js-specific error when the server reports `nodejs.enabled = false`.
   - **Connect Cloud is not supported** — `type = "nodejs"` with `product_type = "connect_cloud"` is rejected at load time.

   Example snippet (note: no `[node]` line):

   ```toml
   type = "nodejs"
   entrypoint = "index.js"
   files = ["*.js", "package.json", "package-lock.json"]
   ```

## `CHANGELOG.md` (root)

Under `## [Unreleased]`, add an `### Added` subsection above the existing `### Fixed`:

> - Support for deploying Node.js applications (Express, Fastify, Hono, etc.) to Posit Connect. (#3964)

Reference the epic (#3964), not the implementation sub-issues. Edit only the root `CHANGELOG.md` — leave `extensions/vscode/CHANGELOG.md` alone.

Issue #3964, opened 2026-04-10
feat: add support for NodeJs app deployment


mconflitti-pbc:
**Experimental feature for Connect**

Need minimal support in publisher to be able to identify and deploy nodejs apps

Connect requires the presence of both package.json and package-lock.json and relies on the information there for running the app over the bundle's manifest.

dotNomad:
Closing as complete 🎉 

Released in version [`v2.6.0`](https://github.com/posit-dev/publisher/releases/tag/v2.6.0)

PR #4129, opened 2026-05-05
feat(detector): add NodejsAppDetector for Node.js project detection

### LINKED ISSUE #3997: feat: Node.js project detector
## Context

Publisher uses a detector chain (`ContentTypeDetector` implementations in `src/inspect/detectors/`) to identify project types. Each detector's `inferType` method receives a base directory and optional entrypoint, returning `PartialConfig[]`. We need a detector that identifies Node.js API projects.

Unlike Python detectors which scan `.py` files for framework imports, Node.js detection is based on the presence of `package.json` and resolving an entrypoint file. Connect determines the specific framework from `package.json` dependencies at deploy time — Publisher doesn't need to distinguish Express from Fastify, etc.

## Goal

Detect Node.js API projects and resolve their entrypoint so users see a concrete file (e.g., `index.js`) in their deployment config, not `package.json`.

## Approach

Create a `NodejsAppDetector` that:

1. Checks if `package.json` exists (the definitive "this is a Node.js project" signal)
2. Resolves the application entrypoint

**Entrypoint resolution** follows Connect's own fallback chain (from `run_app.js`):

- `package.json` `"main"` field
- Parse `package.json` `"scripts.start"` — extract filename from `node <file>` patterns (keep this parser simple: only handle literal `node` commands, skip `nodemon`/`npx`/`tsx` etc.)
- Look for conventional files on disk: `server.js`, `app.js`, `index.js`

**When an `entrypoint` argument is provided** (e.g., user already has a config pointing at a specific file): validate that the file exists and has a valid Node.js extension (`.js`, `.mjs`, `.cjs`, `.ts`, `.mts`, `.cts`), rather than requiring it to match the `package.json`-resolved entrypoint. This matches how Python detectors work — the caller-specified entrypoint is used directly.

**Detector ordering**: Register after all Python framework detectors and before `StaticHTMLDetector`. Note that all detectors run (no short-circuit) — a project with both `package.json` and `app.py` will produce results from both detectors. The user sees all options. Ordering affects sort priority via `sortConfigs()`, so Python results sort higher when both match.

Utility functions `fileExistsAt` and `readFileText` in `src/interpreters/fsUtils.ts` are useful here.

## Design decisions

- **Single content type `NODEJS_API`** — no framework-level differentiation (Express, Fastify, etc.). Connect only has one app mode (`nodejs-api`).
- **No import scanning** — detection is purely structural (package.json + entrypoint existence), not content-based.
- **Simple `scripts.start` parsing** — only handle `node <file>` patterns. Complex scripts (`npx ts-node`, `nodemon`, shell pipelines) are skipped, falling through to conventional file detection.

## Testing considerations

- Detection with each entrypoint source: `main` field, `scripts.start`, conventional files
- Fallback chain order: `main` takes priority over `scripts.start` which takes priority over conventional files
- No match when `package.json` doesn't exist
- No match when no entrypoint resolves
- No match when `entrypoint` argument has a non-Node extension (e.g., `.py`)
- When `entrypoint` argument is provided, matches if file exists on disk (independent of `package.json` metadata)
- `scripts.start` edge cases: flags (`node --inspect app.js`), non-node commands (`nodemon index.js`), complex patterns (`npx ts-node app.ts`)
- Coexistence with Python detectors on projects that have both `package.json` and `.py` files

## Out of scope

- Framework-specific detection (Express vs Fastify vs etc.)
- Subdirectory/monorepo detection
- Yarn/pnpm lock file support (only npm/`package-lock.json` for now)

dotNomad:
## Summary

- Adds `NodejsAppDetector`, a 1:1 port of Connect's `resolveEntrypoint` from `nodejs/run_app.js`, so Publisher and Connect agree on which file Connect will run.
- Resolution priority mirrors Connect exactly: caller-provided entrypoint → `package.json` `main` → `scripts.start` (`/\bnode\s+(.+)/` regex, first non-flag arg with valid extension) → conventional fallback (`server.js, app.js, index.js, server.ts, app.ts, index.ts`).
- Wired into `detectorRunner.ts` between the Python framework detectors and `StaticHTMLDetector`.

Closes #3997. Part of epic #3964.

## Test plan

- [x] `npm run test-unit --prefix extensions/vscode` — all green, including 35 new tests in `nodejs.test.ts` (1716 total).
- [x] `npm run lint --prefix extensions/vscode` — clean.
- [ ] Manual end-to-end deploy of a Node.js project (deferred: homeView-side rendering and publish pipeline arrive in later issues of #3964; this PR's behavior is observable via unit tests).

## Notes for reviewers

- **CHANGELOG / docs intentionally deferred** to #4123, scheduled to land alongside shipped end-to-end behavior at the close of the #3964 epic. No user-visible behavior ships in this PR alone.
- **Detector vs runtime divergences** (intentional): caller entrypoint with bad extension or missing file → `[]` (Connect runtime warns and proceeds); malformed `package.json` → silent `[]` (Connect runtime warns and falls through to conventional files). Rationale: detection is identification, not best-effort execution; we'd rather surface Unknown than guess.
- **Test fixture style** uses real filesystem via `os.tmpdir()` + `fs.mkdtempSync`, mirroring `bundler.test.ts` rather than mocking `fs/promises`.
- **Subdir paths from `main`** (e.g., `"main": "src/index.js"`) are preserved with forward-slash normalization. This is a deliberate scope expansion vs. existing detectors (which only emit basenames).
- `ts-node` parity: the regex matches inside `ts-node app.ts` because of the `\b` word boundary at `-`→`n`. This is Connect's exact behavior; pinned by an explicit test so a future "fix" can't silently diverge.

### DIFF
--- a/extensions/vscode/src/inspect/detectorRunner.ts
+++ b/extensions/vscode/src/inspect/detectorRunner.ts
@@ -17,6 +17,7 @@ import {
   newStreamlitDetector,
   newBokehDetector,
 } from "./detectors/pythonApp";
+import { NodejsAppDetector } from "./detectors/nodejs";
 import { StaticHTMLDetector } from "./detectors/html";
 import { PlumberDetector } from "./detectors/plumber";
 import { RMarkdownDetector } from "./detectors/rmarkdown";
@@ -40,6 +41,7 @@ function createDetectors(): ContentTypeDetector[] {
     newPanelDetector(),
     newStreamlitDetector(),
     newBokehDetector(),
+    new NodejsAppDetector(),
     new StaticHTMLDetector(),
   ];
 }
--- a/extensions/vscode/src/inspect/detectors/nodejs.test.ts
+++ b/extensions/vscode/src/inspect/detectors/nodejs.test.ts
@@ -72,6 +72,16 @@ describe("NodejsAppDetector", () => {
         ]);
       }
     });
+
+    test("caller-provided entrypoint wins when package.json main points elsewhere", async () => {
+      writePackageJson({ main: "other.js" });
+      writeFile("other.js", "");
+      writeFile("server.js", "");
+      const configs = await detector.inferType(baseDir, "server.js");
+      expect(configs).toEqual([
+        { type: ContentType.NODEJS, entrypoint: "server.js" },
+      ]);
+    });
   });
 
   describe("package.json main field", () => {
@@ -209,6 +219,30 @@ describe("NodejsAppDetector", () => {
         { type: ContentType.NODEJS, entrypoint: "server.js" },
       ]);
     });
+
+    test("falls through from missing main to scripts.start", async () => {
+      writePackageJson({
+        main: "missing.js",
+        scripts: { start: "node app.js" },
+      });
+      writeFile("app.js", "");
+      const configs = await detector.inferType(baseDir);
+      expect(configs).toEqual([
+        { type: ContentType.NODEJS, entrypoint: "app.js" },
+      ]);
+    });
+
+    test("matches node within ts-node (mirrors Connect parity)", async () => {
+      writePackageJson({ scripts: { start: "ts-node app.ts" } });
+      writeFile("app.ts", "");
+      const configs = await detector.inferType(baseDir);
+      // The \bnode\s+ regex finds the boundary between '-' and 'n', so this
+      // emits app.ts. Mirrors run_app.js exactly. If Connect's behavior changes,
+      // this test should fail loudly so the detector can be re-aligned.
+      expect(configs).toEqual([
+        { type: ContentType.NODEJS, entrypoint: "app.ts" },
+      ]);
+    });
   });
 
   describe("conventional file fallback", () => {
@@ -286,6 +320,15 @@ describe("NodejsAppDetector", () => {
       const configs = await detector.inferType(baseDir);
       expect(configs).toEqual([
         { type: ContentType.NODEJS, entrypoint: "index.js" },
+      ]);
+    });
+
+    test("falls through from non-node scripts.start to conventional fallback", async () => {
+      writePackageJson({ scripts: { start: "nodemon app.js" } });
+      writeFile("server.js", "");
+      const configs = await detector.inferType(baseDir);
+      expect(configs).toEqual([
+        { type: ContentType.NODEJS, entrypoint: "server.js" },
       ]);
     });
   });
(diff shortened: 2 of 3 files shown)


dotNomad:
> * **Line 66 in `parseStartScript`**: The current logic finds the first arg that is both not a flag AND has a valid extension. What happens if the command is `node --require ./setup.js app.js` — would it pick `./setup.js` (if it exists) over `app.js`? This might be intentional to match Connect, just want to confirm.

The detector picks up `/setup.js` because `find` selects the first non-flag argument with a valid extension very similar to Connect.

extensions/vscode/src/inspect/detectorRunner.ts
zackverham, line 37: not for this PR, but I wonder if we should standardize how we construct these detectors. I don't know why we use `new` vs using a constructor function.

extensions/vscode/src/inspect/detectors/nodejs.ts
```diff
@@ -0,0 +1,136 @@
...
+    type: ContentType.NODEJS,
+    entrypoint: toRelForwardSlash(baseDir, abs),
+  };
+}
+
+function isRecord(v: unknown): v is Record<string, unknown> {
+  return typeof v === "object" && v !== null;
+}
+
+function readMain(pkg: unknown): string | undefined {
```
zackverham, line 57: not a huge deal - it might be nice to have docstrings on some of these helpers just to clarify what they're doing / how they're used in the full detector

extensions/vscode/src/inspect/detectors/nodejs.test.ts
```diff
@@ -0,0 +1,335 @@
...
+    test("emits server.js when no main or scripts.start", async () => {
+      writePackageJson({});
+      writeFile("server.js", "");
+      const configs = await detector.inferType(baseDir);
+      expect(configs).toEqual([
+        { type: ContentType.NODEJS, entrypoint: "server.js" },
+      ]);
+    });
+
+    test("prefers server.js over app.js", async () => {
```
zackverham, line 258: is there a reason for this preference order? Or is it just convention

zackverham (approved):
🔥

extensions/vscode/src/inspect/detectorRunner.ts
dotNomad: Yeah that would be good to do as a quick cleanup.

extensions/vscode/src/inspect/detectors/nodejs.test.ts
dotNomad: It matches the order of Connect's discovery

extensions/vscode/src/inspect/detectorRunner.ts
dotNomad: Opened #4149 to capture it

PR #4141, opened 2026-05-07
fix: only deploy docs on release or manual dispatch

### LINKED ISSUE #4140: docs: only publish changes on publisher release
We only want to update docs when a specific publisher release goes out, not on every merge to main. 

We may want to lock this down only after the site is good and stable in a v1 form - so that we can get everything tuned up to start.

zackverham:
## Summary
- Changes the docs deployment workflow to trigger on releases and manual dispatch instead of every push to main that touches `docs/`

Closes #4140

## Test plan
- [ ] Verify workflow no longer triggers on pushes to main
- [ ] Verify workflow can be manually dispatched via GitHub Actions UI
- [ ] Verify workflow triggers on a published release

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,11 +1,9 @@
 name: Deploy docs
 
 on:
-  push:
-    branches: [main]
-    paths:
-      - "docs/**"
-      - ".github/workflows/docs.yaml"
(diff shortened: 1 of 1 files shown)


dotNomad (approved):
Makes a lot of sense to me :shipit:

Issue #3996, opened 2026-04-15
feat: Node.js data model, content type & schema


dotNomad:
## Context

Connect added an experimental feature with a new `nodejs-api` app mode for deploying Node.js API applications (Express, Fastify, etc.). Publisher needs the foundational data model and schema to represent Node.js projects before any detection, bundling, or publishing work can begin.

## Goal

Establish the `nodejs-api` content type throughout Publisher's type system and configuration schema so that subsequent issues can build detection, manifest generation, and publishing on top of it.

## Approach

Follow the existing patterns for Python and R content types:

- A `NODEJS_API` content type enum value that maps to/from Connect's `"nodejs-api"` app mode
- A `NodeConfig` type representing the Node.js runtime section in deployment configs, with fields for `version`, `packageFile`, and `packageLockFile`
- `node?: NodeConfig` on `ConfigurationDetails` and a `node?: Record<string, never>` marker on `PartialConfig` (same marker pattern as `python` and `r`)
- Schema updates to `posit-publishing-schema-v3.json` allowing a `[node]` section when `type` is `"nodejs-api"`, with the conditional in the top-level `allOf` (same placement as Python/R type conditionals)
- App mode mapping in both directions, content type string label, and addition to `isAPIContentType` (Node.js APIs require the `allow-apis` license)

## Design decisions

- **No `packageManager` field on `NodeConfig`**, unlike `PythonConfig`/`RConfig`. Connect's Node.js runtime only supports `npm ci` — the package manager is not user-configurable.
- **Connect Cloud is not supported.** The `node` property should only be defined in `config_connect`, not `config_connect_cloud`. The `forceProductTypeCompliance` function should strip `node` for Connect Cloud configs. `NODEJS_API` should NOT be added to `connectCloudSupportedTypes` in `configLoader.ts`.
- The Go schema (`internal/schema/schemas/`) should be kept in sync even though Go doesn't use the Node.js fields.

## Testing considerations

- App mode mapping round-trips correctly
- Schema validates a proper Node.js config, requires `[node]` when type is `nodejs-api`, and rejects `[node]` for Connect Cloud product type
- Config compliance strips `node` for Connect Cloud
- Config loader deserializes `[node]` section correctly (the generic camelCase converter should handle this — verify with a round-trip test)
- `isAPIContentType` returns true for the new type

## Out of scope

- Detection logic #3997
- Normalization and interpreter defaults #3998 
- Manifest generation #3999 
- Publishing flow changes #4000

[...]

PR #4150, opened 2026-05-08
feat(nodejs): normalize Node.js configs and bundle package files

### LINKED ISSUE #3998: feat: Node.js normalization & interpreter defaults
## Context

After a detector produces a `PartialConfig` with a `node: {}` marker, the normalization step (`src/inspect/normalize.ts`) fills in runtime details and the interpreter defaults system ensures empty config fields get populated when configs are loaded. For Python/R this involves subprocess calls (`python --version`, `R --version`). For Node.js, all information comes from reading `package.json` — no subprocess needed.

## Goal

When a Node.js project is detected, fill in the `[node]` config section with version and package file information from `package.json`, and integrate into the interpreter defaults system so existing configs get populated correctly.

### Normalization

Add a Node.js block in `normalize.ts` (following Python and R blocks) that runs when `cfg.node` is present:

- Read `package.json` and extract `engines.node` for the version constraint. Leave as empty string if absent.
- Check for `package-lock.json` existence and add both package files to the files list.
- Set `packageLockFile` to `"package-lock.json"` unconditionally — lock file absence is caught at publish-time preflight (issue 6), not normalization. This parallels Python always setting `packageFile: "requirements.txt"`.
- Wire the resulting `NodeConfig` into the `NormalizedConfig` return value alongside `python`, `r`, and `quarto`.

**Important**: The existing R-detection logic in `normalize.ts` checks for `renv.lock` on all content types that are not HTML and not Python. Without a guard, `NODEJS_API` would get an unwanted R config. The Node.js content type needs to be excluded from this check.

### Config writing

The generic `convertKeysToSnakeCase` converter should handle `packageFile` -> `package_file` and `packageLockFile` -> `package_lock_file`. Verify with round-trip tests.

### Interpreter defaults

Add Node.js to the `InterpreterDefaults` and `InterpreterDetectionResult` types and the `getInterpreterDefaults` function. The "detection" is a `package.json` read (not a subprocess), returning empty defaults when the file doesn't exist. This runs for all projects (low cost — single file read) and results are only consumed when `config.node` exists, via `UpdateConfigWithDefaults`.

Note: `InterpreterDetectionResult` (in `src/interpreters/index.ts`) and `InterpreterDefaults` (in `src/api/types/interpreters.ts`) are structurally separate types that happen to be compatible via TypeScript structural typing. Both need the `node` field using `NodeConfig`. This introduces a new import from `api/types/configurations.ts` into the interpreters module — acceptable since Node.js has no separate interpreter config type.

Existing test factories (`interpreterDefaultsFactory` in `src/test/unit-test-utils/factories.ts`) and mocks in `state.test.ts` will need updating to include the `node` field.

## Design decisions

- **No local Node.js interpreter detection.** Version comes from `engines.node` in `package.json`, not `node --version`. Connect picks the nearest installed version regardless.
- **`packageLockFile` always set to `"package-lock.json"`** even if the file doesn't exist. Mirrors Python's approach — validated at publish time.
- **Node.js defaults run for all projects.** Consistent with Python/R detection running for all projects. Cost is negligible (one file read).

## Testing considerations

- Normalization with `engines.node` present and absent
- `package-lock.json` present and absent (both paths: file list population and config field)
- R-detection guard: a Node.js project should NOT get an R config attached
- Config writing round-trip: write `[node]` section, read it back, verify fields match
- `UpdateConfigWithDefaults` fills empty `node` fields and preserves non-empty ones
- Existing test infrastructure compiles after adding `node` to defaults types

## Out of scope

- Local Node.js binary detection (`node --version`)
- Yarn/pnpm lock file alternatives
- Node Packages sidebar view (separate follow-up issue)

dotNomad:
## Summary

Adds Node.js normalization that adds `package.json` and `package-lock.sjon` to the files list in the configuration.

- Guard the existing renv.lock R-detection check to exclude `ContentType.NODEJS`, so a Node.js project doesn't get an unwanted R config attached.

- No `[node]` section is populated — `type = "nodejs"` is sufficient signal, consistent with the design decision in #3996. Connect picks the runtime from `package.json`'s `engines.node` at build time, so no local Node.js detection or `InterpreterDefaults` plumbing is required here.

Resolves #3998.

Builds on the `NodejsAppDetector` from #4129.

## Test plan

- [x] `npx vitest run src/inspect/normalize.test.ts` — all 20 tests pass, including three new tests:
  - adds `package.json` and `package-lock.json` to files list for NODEJS type
  - does not trigger renv.lock R detection for NODEJS type
  - does not set python or r config for NODEJS type
- [x] `npm run test-unit` — 1708 tests pass; the one pre-existing failure (`@posit-dev/mock-connect` import) is unrelated and reproduces on `main`.
- [x] `just lint` — clean.

## Out of scope

- Changelog entry — deferred until the full Node.js deployment feature lands.
- Yarn / pnpm lockfile alternatives.
- Local Node.js binary detection.
- Node Packages sidebar view.

### DIFF
--- a/extensions/vscode/src/inspect/normalize.ts
+++ b/extensions/vscode/src/inspect/normalize.ts
@@ -95,6 +95,7 @@ export async function normalizeConfig(
   if (
     !needR &&
     cfg.type !== ContentType.HTML &&
+    cfg.type !== ContentType.NODEJS &&
     !isPythonContentType(cfg.type)
   ) {
     // For non-HTML, non-Python content: check for renv.lock
@@ -124,6 +125,13 @@ export async function normalizeConfig(
     if (result.config.packageFile) {
       files.push(`/${result.config.packageFile}`);
     }
+  }
+
+  // Node.js: bundle package.json and package-lock.json. There is no [node]
+  // section to populate — Connect picks the runtime from package.json's
+  // engines.node at build time.
+  if (cfg.type === ContentType.NODEJS) {
+    files.push("/package.json", "/package-lock.json");
   }
 
   const comments = initialComment.split("\n");
--- a/extensions/vscode/src/inspect/normalize.test.ts
+++ b/extensions/vscode/src/inspect/normalize.test.ts
@@ -306,6 +306,44 @@ describe("normalizeConfig", () => {
     expect(result.alternatives).toBeUndefined();
   });
 
+  test("adds package.json and package-lock.json to files list for NODEJS type", async () => {
+    const cfg: PartialConfig = {
+      type: ContentType.NODEJS,
+      entrypoint: "index.js",
+    };
+    setupDefaultMocks();
+
+    const result = await normalizeConfig(cfg, "/project");
+    expect(result.files).toContain("/index.js");
+    expect(result.files).toContain("/package.json");
+    expect(result.files).toContain("/package-lock.json");
+  });
+
+  test("does not trigger renv.lock R detection for NODEJS type", async () => {
+    const cfg: PartialConfig = {
+      type: ContentType.NODEJS,
+      entrypoint: "index.js",
+    };
+    // renv.lock exists, but NODEJS type should skip renv.lock check
+    mockAccess.mockResolvedValue(undefined);
+    mockReadFile.mockRejectedValue(new Error("ENOENT"));
+
+    const result = await normalizeConfig(cfg, "/project");
+    expect(result.r).toBeUndefined();
+  });
+
+  test("does not set python or r config for NODEJS type", async () => {
+    const cfg: PartialConfig = {
+      type: ContentType.NODEJS,
+      entrypoint: "index.js",
+    };
+    setupDefaultMocks();
+
+    const result = await normalizeConfig(cfg, "/project");
+    expect(result.python).toBeUndefined();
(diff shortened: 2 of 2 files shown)


extensions/vscode/src/inspect/normalize.ts
```diff
@@ -126,6 +127,13 @@ export async function normalizeConfig(
     }
   }
 
+  // Node.js: bundle package.json and package-lock.json. There is no [node]
+  // section to populate — Connect picks the runtime from package.json's
+  // engines.node at build time.
+  if (cfg.type === ContentType.NODEJS) {
+    files.push("/package.json", "/package-lock.json");
```
zackverham, line 134: is there a chance that these files won't actually exist? Or is it safe to assume they'll always be there?

zackverham (approved):

extensions/vscode/src/inspect/normalize.ts
dotNomad: There is a chance they won't be there, but this pushes them to the files list in the configuration, so even if they don't exist it is not bad to have them in the configuration.

Ideally we have a pre-flight check or some warning to say that these don't exist, but that is for a future PR.

PR #4151, opened 2026-05-08
feat(nodejs): generate Node.js manifest without packages block

### LINKED ISSUE #3999: feat: Node.js manifest & bundling
## Context

Connect expects Node.js bundles to contain a `manifest.json` with `"appmode": "nodejs-api"`, a `"nodejs": {}` section, and an entrypoint in `metadata`. The bundle must include `package.json` and `package-lock.json`. Connect reads dependencies directly from these files — unlike Python/R, the manifest's `packages` section stays empty.

## Goal

Generate correct Connect-compatible manifests for Node.js projects and ensure bundles contain the right files.

### Manifest type and generation

Add a `nodejs?: object` field to the `Manifest` type in `src/bundler/types.ts`. Use `object` rather than `Record<string, never>` — the latter causes type errors when assigning `{}`.

In `manifestFromConfig.ts`, emit `nodejs: {}` when `cfg.node` is present. This follows the existing spread pattern used for `python`, `jupyter`, and `quarto` sections. The `packages` section stays empty — Connect reads `package.json`/`package-lock.json` from the bundle directly, unlike Python/R where every package is enumerated in the manifest.

The `metadata.entrypoint` is already set from `cfg.entrypoint` (the JS/TS file resolved by the detector). No changes to `primaryField()` — Node.js doesn't use `primary_rmd` or `primary_html`.

### Bundling

No changes needed to the bundler:

- `node_modules/` is already in `STANDARD_EXCLUSIONS`
- File collection, checksumming, and tar.gz creation are generic
- No synthetic file generation needed (unlike Python where `requirements.txt` may be generated)

### Expected manifest output

```json
{
  "version": 1,
  "metadata": {
    "appmode": "nodejs-api",
    "entrypoint": "index.js"
  },
  "nodejs": {},
  "packages": {},
  "files": {
    "index.js": { "checksum": "abc123..." },
    "package.json": { "checksum": "def456..." },
    "package-lock.json": { "checksum": "ghi789..." }
  }
}
```

## Design decisions

- **`nodejs: {}` is an empty object** — Connect requires its presence but currently reserves it for future use. No version or package manager info goes here.
- **No package enumeration** — Connect's Node.js runtime reads dependencies from `package.json`/`package-lock.json` directly, unlike Python/R where the manifest `packages` section lists every dependency.

## Testing considerations

- `manifestFromConfig` with Node.js config produces `"appmode": "nodejs-api"`, `"nodejs": {}`, correct entrypoint, and empty `packages`
- `manifestFromConfig` without Node.js config does NOT include `nodejs` section (no regression)
- Verify `node_modules/` is excluded from bundles

## Out of scope

- Package enumeration in the manifest
- Pre-built `node_modules/` inclusion for `environment_manager: "none"` mode

dotNomad:
## Summary

Avoids writing a consistently empty `packages: {}` field for Node.js content types since Connect always reads from `package.json` and `package-lock.json`

- Mark `Manifest.packages` optional in `bundler/types.ts`.

- Add unit + golden tests for the Node.js manifest path; existing golden tests cover the regression case (R/Python/Quarto/HTML manifests still emit `"packages": {}`).

- Add an end-to-end bundler test that walks a Node.js project through `manifestFromConfig` + `createBundle` and asserts the bundled `manifest.json` has the package files with checksums, no `packages` block, and excludes `node_modules/`.

Resolves #3999

## Expected manifest output for Node.js

```json
{
  "version": 1,
  "metadata": {
    "appmode": "nodejs",
    "entrypoint": "index.js"
  },
  "files": {
    "index.js": { "checksum": "..." },
    "package.json": { "checksum": "..." },
    "package-lock.json": { "checksum": "..." }
  }
}
```

No `python`, `jupyter`, `quarto`, `environment`, `platform`, or `packages` blocks. (Connect's Go deserializer treats a missing field as the zero value, so omitting these is safe.)

## Test plan

- [x] `npx vitest run src/bundler/manifestFromConfig.test.ts src/bundler/manifestFromConfig.golden.test.ts src/bundler/bundler.test.ts` — 96 tests pass, including:
  - 4 new unit tests under `Node.js section` (appmode/entrypoint, packages omitted, no other blocks, regression: non-NODEJS still emits `packages: {}`)
  - 1 new golden test (`testdata/nodejs.json`)
  - 1 new bundler integration test (Node.js project end-to-end)
- [x] Verified the bundler integration test fails when `manifestFromConfig` is reverted to emit `packages: {}` for NODEJS — confirms it catches the regression.
- [x] `npm run test-unit` — 1713 tests pass; the lone failure (`@posit-dev/mock-connect` import in `integrationRequests.e2e.test.ts`) reproduces on `main` and is unrelated.
- [x] `just lint` — clean.
- [x] Audited callers of `manifest.packages`: only written (never read-without-write) and only on R-only code paths in `publish/publishShared.ts`. Making it optional is safe.

## Out of scope

- Pre-built `node_modules/` inclusion when `environment.environment_management.nodejs = false`.
- Changelog entry — deferred until the full Node.js deployment feature lands.

### DIFF
--- a/extensions/vscode/src/bundler/bundler.test.ts
+++ b/extensions/vscode/src/bundler/bundler.test.ts
@@ -177,6 +177,51 @@ describe("createBundle", () => {
     expect(entries.has("node_modules/pkg/index.js")).toBe(false);
   });
 
+  it("bundles a Node.js project with package files and no packages block", async () => {
+    makeFile("index.js", "console.log('hi');");
+    makeFile("package.json", '{"name":"app","version":"1.0.0"}');
+    makeFile("package-lock.json", '{"lockfileVersion":3}');
+    makeFile("node_modules/foo/index.js", "module.exports = {};");
+
+    const manifest = manifestFromConfig({
+      $schema: "" as never,
+      productType: ProductType.CONNECT,
+      type: ContentType.NODEJS,
+      entrypoint: "index.js",
+      validate: true,
+    });
+
+    const result = await createBundle({
+      projectPath: tmpDir,
+      manifest,
+      filePatterns: ["/index.js", "/package.json", "/package-lock.json"],
+    });
+
+    // Package files are present in the manifest with valid checksums
+    expect(result.manifest.files["index.js"]?.checksum).toMatch(/^[0-9a-f]+$/);
+    expect(result.manifest.files["package.json"]?.checksum).toMatch(
+      /^[0-9a-f]+$/,
+    );
+    expect(result.manifest.files["package-lock.json"]?.checksum).toMatch(
+      /^[0-9a-f]+$/,
+    );
+
+    // node_modules/ is excluded
+    expect(
+      Object.keys(result.manifest.files).some((k) =>
+        k.startsWith("node_modules/"),
+      ),
+    ).toBe(false);
+
+    // The bundled manifest.json has no packages block (Node.js manifests omit it)
+    const entries = await extractTarEntries(result.bundle);
+    const manifestContent = JSON.parse(
+      entries.get("manifest.json")!.data.toString(),
+    );
+    expect(manifestContent.metadata.appmode).toBe("nodejs");
+    expect(manifestContent.packages).toBeUndefined();
+  });
+
   it("creates a bundle from a single file path", async () => {
     makeFile("app.py", "main app");
     makeFile("other.py", "other file");
--- a/extensions/vscode/src/bundler/types.ts
+++ b/extensions/vscode/src/bundler/types.ts
@@ -11,7 +11,7 @@ export type Manifest = {
   jupyter?: ManifestJupyter;
   quarto?: ManifestQuarto;
   environment?: ManifestEnvironment;
-  packages: Record<string, ManifestPackage>;
+  packages?: Record<string, ManifestPackage>;
   files: Record<string, ManifestFile>;
   integration_requests?: ManifestIntegrationRequest[];
 };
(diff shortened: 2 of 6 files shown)


extensions/vscode/src/bundler/types.ts
```diff
@@ -11,7 +11,7 @@ export type Manifest = {
   jupyter?: ManifestJupyter;
   quarto?: ManifestQuarto;
   environment?: ManifestEnvironment;
-  packages: Record<string, ManifestPackage>;
+  packages?: Record<string, ManifestPackage>;
```
dotNomad, line 14: `packages` was technically always optional at least from Connect's point of view. Node.js is the first content type that doesn't write one since it is always empty.

extensions/vscode/src/bundler/bundler.test.ts
```diff
@@ -177,6 +177,51 @@ describe("createBundle", () => {
...
+      ),
+    ).toBe(false);
+
+    // The bundled manifest.json has no packages block (Node.js manifests omit it)
+    const entries = await extractTarEntries(result.bundle);
+    const manifestContent = JSON.parse(
+      entries.get("manifest.json")!.data.toString(),
+    );
+    expect(manifestContent.metadata.appmode).toBe("nodejs");
+    expect(manifestContent.packages).toBeUndefined();
```
zackverham, line 222: just saying it back to you to make sure I'm following - we don't want / need a packages block in the manifest.json because we infer packages from package-lock.json, which is referenced in the files section of the manifest. Is that right?

zackverham (approved):

extensions/vscode/src/bundler/bundler.test.ts
dotNomad: Yes that is right. Connect doesn't support listing Node.js packages in the `manifest.json` at all. npm packages are installed using `package.json` and `package-lock.json` 👍 

So here we aren't adding `packages` to the `manifest.json` specifically.

PR #4174, opened 2026-05-15
fix: inlining style and icon files to load in Workbench in Firefox

### LINKED ISSUE #4013: bug: Loading CSS in Workbench in Firefox does not work
Currently there is an issue with the latest version of Firefox where the CSS for the extension isn't loaded due to a `SSL_ERROR_BAD_CERT_DOMAIN` network error.

Based on analysis this seems to be an issue with Firefox service worker interception of VS Code / Positron webview resources.

fizzboop:
## Intent
Resolves #4013

## Type of Change
- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach
It was suggested to follow a similar approach that Positron did by inlining the CSS file so the service worker in Firefox doesn't need to intercept a request. This worked, however, I noticed the icons need to be handled the same way. With only the styles, it looked like the following:

<details>
<summary>Without the Icon Fix</summary>
image
</details>

## User Impact
The style and icons of the Publisher on Workbench in Firefox appeared like so: 

<details>
<summary>Bug</summary>
image
</details>

Now both the styles and icons appear
<details>
<summary>Fixed Style and Icons</summary>
image
</details>

## Automated Tests
WIP

## Directions for Reviewers
1. Build the extension: `just package` from `extensions/vscode/`.
2. Install the resulting `.vsix` in a Positron-on-Workbench session opened in Firefox
3. Open the Publisher sidebar and confirm styles + icons render correctly
4. There should be no requests for `codicon.css` or `posit-publisher-icons.css`, and the Console shows no CSP violations.

## Checklist

### DIFF
--- a/extensions/vscode/scripts/inline-icon-fonts.mjs
+++ b/extensions/vscode/scripts/inline-icon-fonts.mjs
@@ -1,0 +1,49 @@
+#!/usr/bin/env node
+// Reads each icon-font CSS file under dist/ and rewrites its @font-face
+// `url("./<font>?<hash>")` reference to a base64 `data:` URI, writing the
+// result to a sibling `*.inlined.css`. The extension's webview HTML reads
+// these prebuilt files and injects them as inline <style> blocks so
+// Firefox's broken service-worker interception (in Posit Workbench) can't
+// fail the stylesheet/font fetches.
+// See https://github.com/posit-dev/publisher/issues/4013.
+import { readFileSync, writeFileSync } from "node:fs";
+import { basename } from "node:path";
+
+const targets = [
+  {
+    cssPath: "dist/codicons/codicon.css",
+    fontPath: "dist/codicons/codicon.ttf",
+    mimeType: "font/ttf",
+    outPath: "dist/codicons/codicon.inlined.css",
+  },
+  {
+    cssPath: "dist/posit-publisher-icons.css",
+    fontPath: "dist/posit-publisher-icons.woff2",
+    mimeType: "font/woff2",
+    outPath: "dist/posit-publisher-icons.inlined.css",
+  },
+];
+
+function escapeRegExp(str) {
+  return str.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
+for (const { cssPath, fontPath, mimeType, outPath } of targets) {
+  const cssContent = readFileSync(cssPath, "utf8");
+  const fontBase64 = readFileSync(fontPath).toString("base64");
+  const fontBasename = basename(fontPath);
+  const pattern = new RegExp(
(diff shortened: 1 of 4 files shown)


extensions/vscode/scripts/inline-icon-fonts.mjs
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env node
```
zackverham, line 1: cool! This is an elegant way to bypass the issue. 

Is there any need for tests on this file do you think? Or is the fact that things are working in CI sufficient for us?

zackverham (approved):

extensions/vscode/scripts/inline-icon-fonts.mjs
fizzboop: i think CI passing would be good signal that things are working as expected. the script will error out if there are any issues with the fonts.

fizzboop:
looks like fantasticon is broken on [Windows](https://github.com/tancredi/fantasticon/issues/470). they look similar to these [failures](https://github.com/posit-dev/publisher/actions/runs/25939213841/job/76252350728?pr=4174) in CI.  the root cause is upstream by fantasticon building a glob path with backslashes on Windows that newer glob versions treat as escape characters.

fizzboop:
A few options to approach this:

1. swap to @twbs/fantasticon, Bootstrap's maintained fork that pins glob ^7.2.3 (where backslash paths still work), but this feels like a fork-dependency risk.

2. commit a prebuilt .css and .woff2 and drop build-font from the CI build chain. fantasticon would no longer run in CI which was the cause of the Window related failures.

3. patch-package for fantasticon in `node_modules`. this seems super brittle and we could run into more glob related issues.

dotNomad:
I think we can fix this without changing `fantasticon` which we've already bounced back and forth

https://github.com/posit-dev/publisher/pull/3435

I'll do some digging today

dotNomad:
@fizzboop I created #4199 off of this pull request showing a potential solution

It moves `npm run build-font and npm run inline-icon-fonts` back to `vscode:prepublish` and has `homeView.ts` split between the inline and non-inline paths.

Inline is used with the actual `vsix` where the inline files are available. In the extension dev host the non-inline versions are used so the tests and local development don't need to do `fantasticon`.

If you like that solution we can pull that in.

dotNomad (approved):
Tested the change I pulled in locally and in Workbench on Firefox

Image

The icons loaded on both and CI passed the previous erroring point. I figure the trade off of having `homeView.ts` know whether to look at the inline vs non-inline font sources was worth avoiding larger refactoring.

PR #4181, opened 2026-05-16
ci(docs): add lychee link checking to docs preview workflow

### LINKED ISSUE #4136: docs: add link checking
with the docs up on a public site, we probably want to add link checking and do minimal due dilligence to make sure that the site is published in a good state.

m--:
## Summary

Resolves #4136

- Adds [lychee](https://github.com/lycheeverse/lychee) link checking to the docs PR preview workflow. Lychee runs after the preview deploy step, so a broken link fails the PR check without blocking the preview itself — authors can still visually review their changes.
- Configures lychee via `docs/lychee.toml`: HTML-mode scanning of `docs/_site`, anchor-fragment validation enabled, `_site/site_libs` and `_site/llms*.txt` excluded (the former is vendored Quarto assets, the latter are plain-text dumps that duplicate HTML coverage and lose backtick context). Uses `GITHUB_TOKEN` to authenticate against the GitHub API for higher rate limits.
- Repairs pre-existing doc-content issues surfaced during setup:
  - `configuration.md` — wrap the `my-account-vanity` placeholder URL in backticks so it isn't extracted from prose and DNS-checked.
  - `configuration.md` — convert the bare publishing-cli-apps reference into a real markdown link and update the anchor to `#understanding-entrypoints` (the upstream section was renamed).
  - `configuration.md` — drop the `#engines` fragment from the npm docs link; the npm docs site strips that fragment when redirecting to a versioned path, so it's non-functional regardless of where it's clicked from.


docs/configuration.md
```diff
@@ -233,7 +233,7 @@ _Only valid when `product_type` is `connect`_
 
 Both `package.json` and `package-lock.json` must be present and listed under `files`. Connect uses `npm ci` to install dependencies; if `package-lock.json` is missing, run `npm install` to generate it.
 
-If `package.json` specifies an [`engines.node`](https://docs.npmjs.com/cli/configuring-npm/package-json#engines) constraint, Connect uses it to select the runtime version. Otherwise Connect uses the latest version of Node.js available on the server.
+If `package.json` specifies an [`engines.node`](https://docs.npmjs.com/cli/configuring-npm/package-json) constraint, Connect uses it to select the runtime version. Otherwise Connect uses the latest version of Node.js available on the server.
```
m--, line 236: If we want to link to the `#engines` fragment, it looks like we'd need to pick a specific version of the docs to link to — this redirects to the "latest" docs, but drops the fragment. This seemed like a fair compromise, but I'm happy to switch it.

zackverham: makes sense to me - I think that seems fine.

zackverham (approved):

Issue #4094, opened 2026-04-28
Automate patch releases of older release lines


dotNomad:
## Background

Today's release tooling (`scripts/prepare-release.py`, `.github/workflows/prepare-release.yaml`, `tag-on-release-merge.yaml`, `nightly-prerelease.yaml`) is built around two flows:

1. **Production releases** — operator dispatches `prepare-release.yaml` with an even-minor version; the workflow branches off `main`, runs the script, opens a PR; merge triggers tag + release/publish.
2. **Nightly pre-releases** — daily cron picks the highest semver tag, computes the next odd-minor pre-release, commits the version bump to `main`, tags it.

Patch releases of an older release line (e.g. `2.0.1` while `2.1.x` pre-releases are in flight) don't fit either flow. As surfaced during prep for #4090:

- The release branch ends up cut from `main`, which contains everything since the last release — not just the targeted fix. We got lucky with `2.0.1` (only docs + nightly bumps + the fix), but the system doesn't enforce the "only the fix" guarantee.
- `prepare-release.py` would silently set `package.json` from `2.1.1` down to `2.0.1`. PR #4093 adds a `--allow-downgrade` guard, but that only catches the symptom.
- A maintainer doing a patch today has to manually: cut a branch from the prior tag, cherry-pick the fix, run `prepare-release.py` locally, push, merge, etc.

## Proposal

A new manually dispatched workflow, `prepare-patch-release.yaml`, that automates the mechanical parts:

```yaml
on:
  workflow_dispatch:
    inputs:
      version:        # e.g. "2.0.1" — must be even minor
      base_tag:       # optional; defaults to highest v{major}.{minor}.* tag
      pr_numbers:     # comma-separated, e.g. "4082,4087"
```

Steps it would take:

1. **Validate** — version format + even minor; `base_tag` exists; `release/v$VERSION` branch and `v$VERSION` tag don't already exist.
2. **Derive `base_tag` if omitted** — from `2.0.1`, look up the highest tag matching `v2.0.*`. Saves an input and prevents off-by-one mistakes.
3. **Branch** — `git checkout -b release/v$VERSION $base_tag`.
4. **Cherry-pick each PR** — for each PR number, resolve to its squash-merge SHA on `main` (`gh pr view N --json mergeCommit -q .mergeCommit.oid`) and `git cherry-pick` it. PR numbers are friendlier than raw SHAs and let the workflow render a nice "Backports: #4082, #4087" line in the release PR body.
[...]

PR #4184, opened 2026-05-18
feat: Add prepare-patch-release workflow for older release lines (#4094)

zackverham:
## Summary

Implements #4094 — automates patch releases of older release lines.

- New `prepare-patch-release.yaml` workflow with `version`, `base_tag` (optional), and `pr_numbers` inputs. Branches from the prior tag, cherry-picks specified PRs, runs `prepare-release.py`, and opens a **draft** PR for review (not to be merged).
- New `tag-patch-release.yaml` workflow that tags the release branch directly, closes the review PR, and triggers the release pipeline.
- Adds `--from-cherry-pick` flag to `prepare-release.py` that skips the empty-Unreleased interactive prompt (since cherry-picked commits already carry their CHANGELOG entries).
- On cherry-pick conflict, the workflow pushes the partial branch and opens a **draft** PR with step-by-step resolution instructions.
- Updates `verify-tag-ancestry` to allow tags on `release/v*` branches (not just main).
- Documents when to use `prepare-release` vs `prepare-patch-release` in CONTRIBUTING.md.

### Flow

1. Run **Prepare Patch Release** → creates branch, cherry-picks, opens draft PR for review
2. Team reviews the draft PR (never merges it — it targets main only for diff visibility)
3. Run **Tag Patch Release** → tags the branch directly, closes the PR, triggers release pipeline

## Test plan

- [ ] Validate YAML syntax (done locally via `yaml-lint`)
- [ ] Confirm `prepare-release.py` tests pass with the new flag (`pytest scripts/` — 77 tests pass)
- [ ] Smoke-test on a throwaway version (e.g., cut a `v0.0.1` patch from a `v0.0.0` baseline tag on a fork)
- [ ] Verify conflict path by cherry-picking a PR that touches CHANGELOG in a conflicting way


.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,267 @@
...
+          When this PR is merged, a GitHub Action will automatically:
+          1. Create the \`v$VERSION\` tag
+          2. Trigger the release workflow to build and publish
+
+          ---
+          🤖 Generated by the prepare-patch-release workflow
+          EOF
+          )" \
+            --head "release/v$VERSION" \
+            --base main
```
dotNomad: Doesn't this still target `main`? Meaning that the changed `package.json` with the older version will be overriding the newer version on main and contain the newer code too not just the fix we want.

.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,267 @@
...
+          \`\`\`
+
+          Once resolved, mark this PR as **Ready for review**.
+
+          ---
+          🤖 Generated by the prepare-patch-release workflow
+          EOF
+          )" \
+            --head "release/v$VERSION" \
+            --base main \
```
dotNomad, line 270: Similar note here about the `--base main`. Don't we need to target the new release branch we created and keep that separate from `main`?

.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,267 @@
...
+            echo "Cherry-picking $SHA (PR #$PR_NUM)..."
+            if ! git cherry-pick "$SHA" 2>/dev/null; then
+              echo "::warning::Cherry-pick conflict on PR #$PR_NUM (SHA: $SHA)"
+              echo "conflict=true" >> "$GITHUB_OUTPUT"
+              echo "conflict_pr=$PR_NUM" >> "$GITHUB_OUTPUT"
+              echo "conflict_sha=$SHA" >> "$GITHUB_OUTPUT"
+              # Abort so we can push the branch in its pre-conflict state
+              git cherry-pick --abort
+              break
+            fi
```
dotNomad, line 147: If we hit a conflict it looks like we completely bail. I think that makes sense, we would have to do this manually.

.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,267 @@
...
+
+            BACKPORTED_PRS="${BACKPORTED_PRS}- #${PR_NUM} ($SHA)\n"
```
dotNomad: It looks like we don't include PRs with conflicts and those after in this list.

.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,267 @@
...
+
+      - name: Install dependencies and sync lockfile
+        if: steps.cherry-pick.outputs.conflict != 'true'
+        run: npm install
+
+      - name: Commit changes
+        if: steps.cherry-pick.outputs.conflict != 'true'
+        run: |
+          git add CHANGELOG.md extensions/vscode/CHANGELOG.md extensions/vscode/package.json package-lock.json
+          git commit -m "Prepare patch release v$VERSION"
```
dotNomad, line 174: Should we check to see if there are actually changes before committing like we do in `prepare-release.yaml`?

dotNomad (commented):
I may need a walk through of this one. Reading through `prepare-patch-release.yaml` I'm a bit lost on how it works. I think I caught a couple of things, but I could be misunderstanding the setup a bit.

.github/workflows/prepare-patch-release.yaml
zackverham: yeah, I think so. I don't think there's a great way around this, if we can't auto-cherry pick then we'll need to manually intervene.

.github/workflows/prepare-patch-release.yaml
zackverham: Probably a good idea. I think the assumption is that there will be changes if we're cherry-picking commits onto an older release line, but it would be a good sanity check.

.github/workflows/prepare-patch-release.yaml
zackverham: oh that's a good point, yeah. I'm not entirely sure where the PR should target then.

.github/workflows/prepare-patch-release.yaml
zackverham: can you say more about this? if we hit a conflict I think we bail out, right?

.github/workflows/prepare-patch-release.yaml
zackverham: Update: I think the move is to create a branch that will never get merged with the cherry-picked commits, and then tag from there. The PR should indicate that it isn't for merging. 

I'm making that change now.

.github/workflows/prepare-patch-release.yaml
dotNomad: Yeah we bail out. Sorry this comment was vague. It was mainly about not logging the additional PRs to backport.

.github/workflows/release.yaml
```diff
@@ -7,9 +7,9 @@ on:
 permissions:
   contents: write # Required to create GitHub releases
 jobs:
-  # Security: Verify the tag points to a commit on main branch
+  # Security: Verify the tag points to a commit on main or a release branch
   verify-tag:
-    name: Verify tag is on main branch
+    name: Verify tag is on main or release branch
```
dotNomad, line 12: Could update the name in `publish.yaml` as well.

.github/actions/verify-tag-ancestry/action.yaml
```diff
@@ -1,18 +1,32 @@
...
+        # For patch releases, allow tags on release/v* branches
+        # Tag "v2.4.1" → branch "release/v2.4.1"
+        TAG_NAME="${GITHUB_REF_NAME}"
+        RELEASE_BRANCH="release/${TAG_NAME}"
+        git fetch origin "$RELEASE_BRANCH" 2>/dev/null || true
+        if git rev-parse "origin/$RELEASE_BRANCH" >/dev/null 2>&1 && \
+           git merge-base --is-ancestor "$COMMIT_SHA" "origin/$RELEASE_BRANCH"; then
+          echo "✅ Verified: Tag points to commit on release branch $RELEASE_BRANCH"
+          exit 0
+        fi
```
dotNomad, line 29: This is great, but prior to merge we need to setup a branch protection rule for branches named `release/*`.

Right now this trusts any tag pointed at a branch `release/v{tag}` with no guarding. We can limit who can push to those branches, and setup blocking force-pushes.

.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,294 @@
...
+            # Validate PR number is numeric
+            if ! [[ "$PR_NUM" =~ ^[0-9]+$ ]]; then
+              echo "::error::Invalid PR number: $PR_NUM (must be numeric)"
+              exit 1
+            fi
+
+            echo "Processing PR #$PR_NUM..."
+
+            # Resolve squash-merge SHA
+            SHA=$(gh pr view "$PR_NUM" --json mergeCommit -q .mergeCommit.oid)
```
dotNomad, line 132: A Claude analysis of this PR pointed this out. It assumes squash merges with a single `mergeCommit.oid`.

We can also do rebase merging (maybe we turn that off for this?) and if we do that on purpose or not `mergeCommit.oid` is only the last replayed commit, so a multi-commit PR would have earlier commits dropped.

It seems like the best would be to add a comment and disable squash merges so this can work? What do you think @zackverham

.github/workflows/prepare-patch-release.yaml
```diff
@@ -0,0 +1,294 @@
...
+          🤖 Generated by the prepare-patch-release workflow
+          EOF
+
+          gh pr create \
+            --title "Release v$VERSION (patch) [DO NOT MERGE]" \
+            --body-file /tmp/pr-body.md \
+            --head "release/v$VERSION" \
+            --base main \
+            --label "do-not-merge" \
+            --draft
```
dotNomad, line 223: Out of curiosity why do we still make a pull request if we are tagging a branch?

dotNomad: If it is only for a convenient diff we can use GitHub's difference between SHAs view

dotNomad (commented):
I think we have a few things we need to do prior to merging. Curious on your thoughts on some of these questions @zackverham

.github/actions/verify-tag-ancestry/action.yaml
zackverham: ah, yeah. You're right. hmm.

.github/workflows/prepare-patch-release.yaml
zackverham: yeah we can for sure use a diff between SHAs view - the pull request gives a convenient diff, and would allow for commenting / others to approve the change prior to merging. But it is kind of shoehorning the PR tools for something that isn't a PR.

.github/workflows/prepare-patch-release.yaml
zackverham: are we able to to disable squash merges just for an isolated workflow? I see your point, that's a good catch.

.github/workflows/prepare-patch-release.yaml
dotNomad: I'm fine either way. We could potentially look into branch protection rules on merging from `release/v{tag}` branches too if we want. Just voicing it

zackverham: yeah totally - I'm not entirely sure what the right approach is. Maybe a branch protection rule is the right way forward.

.github/workflows/prepare-patch-release.yaml
dotNomad: I just disabled rebase merging 👍

dotNomad (approved):
After some repo setting changes this is looking good.

PR #4194, opened 2026-05-19
docs: document @types/vscode and engines.vscode pinning policy

dotNomad:
## Summary
- Adds a section to `CONTRIBUTING.md` (under *Updating Dependencies*) describing how we pin `engines.vscode` and `@types/vscode`, why `vsce package` requires their ranges to line up, and how to bump them.
- Codifies the rule from #4193: pin both to the same exact version (no caret) across `extensions/vscode/package.json` and `test/extension-contract-tests/package.json` so Dependabot can't drift them apart.

## Test plan
- [ ] Read the rendered diff and confirm the policy matches what we want contributors (and future-us) to follow.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -350,9 +350,7 @@ The VS Code extension declares two related versions:
 - `engines.vscode` in `extensions/vscode/package.json` — the minimum VS Code version users need to install the extension.
 - `@types/vscode` in `extensions/vscode/package.json` and `test/extension-contract-tests/package.json` — the API surface the TypeScript code compiles against.
 
-`vsce package` fails if the declared `@types/vscode` range is higher than the declared `engines.vscode` range (it refuses to ship an extension that compiles against APIs newer than the minimum runtime it claims to support). Keeping these two in lockstep avoids that failure and keeps `engines.vscode` honest — if it's lower than `@types/vscode`, the extension will install on VS Code versions that lack APIs the code calls, and crash at runtime.
-
-We pin both to the **same exact version, with no caret**:
+`vsce package` fails if `@types/vscode` is higher than `engines.vscode` — it refuses to ship an extension that compiles against APIs newer than the minimum runtime it claims to support. Pin both to the same version across all three places:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -342,3 +342,29 @@ Any significantly difficult dependency updates should have an issue created to
...
 Updates to dependencies should be done in a separate PR.
+
+### `engines.vscode` and `@types/vscode`
+
+The VS Code extension declares two related versions:
+
+- [`engines.vscode`](https://code.visualstudio.com/api/references/extension-manifest) in `extensions/vscode/package.json` — the minimum VS Code version users need to install the extension.
+- `@types/vscode` in `extensions/vscode/package.json` and `test/extension-contract-tests/package.json` — the API surface the TypeScript code compiles against.
+
+`vsce package` fails if `@types/vscode` is higher than `engines.vscode` — it refuses to ship an extension that compiles against APIs newer than the minimum runtime it claims to support. Pin `@types/vscode` to an exact version (so the code can't drift onto newer APIs) and use a caret on `engines.vscode` (so the extension installs on that VS Code version and any newer one). Use the same version number in all three places:
```
m--: Can we link to docs for `vsce package` if this is the first time we've talked about it in this doc?

CONTRIBUTING.md
```diff
@@ -342,3 +342,29 @@ Any significantly difficult dependency updates should have an issue created to
...
+```
+
+```jsonc
+// test/extension-contract-tests/package.json
+"devDependencies": {
+  "@types/vscode": "1.105.0"
+}
+```
+
+When adding code that needs a VS Code API not present in the current pinned version, bump all three places to the lowest version that exposes the API you need (check the VS Code API release notes), then run `npm install`. Raising the pin raises the minimum VS Code version users need, so don't bump further than required.
```
m--: could link "VS Code API release notes" to https://code.visualstudio.com/updates/, which will redirect to the latest version it looks like (?)

m-- (approved):

PR #4199, opened 2026-05-20
fix: keep fantasticon out of esbuild-base to unbreak Windows CI

dotNomad:
## Intent

Unbreak the Windows ``vscode / native`` CI jobs on #4174.

## Type of Change

- - [x] Tooling 

## Approach

``fantasticon@4.1.0`` is broken on Windows: it calls ``path.join(inputDir, "**/*.svg")`` then passes the result to ``glob@13``. On Windows the joined path uses backslashes, which ``glob`` treats as escape characters, so it matches zero files and fantasticon throws ``No SVGs found in ./assets/icons``.

#4174 moved ``build-font`` (and the new ``inline-icon-fonts``) into ``esbuild-base``. Because ``pretest`` runs ``esbuild-base``, the Windows native test matrix started invoking fantasticon for the first time and failed. macOS / Linux passed; Windows didn't.

This PR moves both back to ``vscode:prepublish`` only, which runs Ubuntu-only via the ``package`` job. ``esbuild-base`` is restored to its pre-#4174 shape and no longer touches fantasticon.

To keep the Workbench/Firefox fix from #4013 working, ``homeView.getWebviewContent()`` now picks its strategy based on whether the prebuilt ``*.inlined.css`` files exist:

- Packaged extension (``.vsix``): inlined files are present, embed them as ``<style>`` blocks with ``font-src data:`` — same behavior as #4174.
- F5 Extension Development Host: inlined files are absent, fall back to ``<link rel=\"stylesheet\">`` via webview URIs with ``font-src \${webview.cspSource}``. The Firefox/Workbench service-worker interception bug doesn't apply in the Extension Development Host, so plain stylesheet links work.

Trade-off: ``homeView`` has two output shapes guarded by a single ``existsSync`` check, but the duplication is small and the CSP differs honestly between contexts.

## User Impact

None. Users continue to get the #4013 fix in the packaged extension. Windows CI is green again.

## Automated Tests

No new tests — the change is build-config plus a runtime fallback. Validation is CI passing on this branch (which is what failed on #4174).

## Directions for Reviewers

1. Confirm the Windows ``vscode / native (windows-latest)`` and ``windows-11-arm`` jobs go green on this PR.
2. Confirm ``package / vscode`` still produces a working ``.vsix`` (it runs on Ubuntu, so fantasticon executes there).
3. Optional: ``just vscode package`` locally, install the ``.vsix``, open the sidebar in Workbench/Firefox, verify icons and styles still render (the #4174 behavior).

## Checklist

  - No CHANGELOG entry — this is internal build tooling, not user-facing.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -643,10 +643,10 @@
     ]
   },
   "scripts": {
-    "vscode:prepublish": "npm run --prefix webviews/homeView build && npm run esbuild-base -- --minify",
+    "vscode:prepublish": "npm run --prefix webviews/homeView build && npm run esbuild-base -- --minify && npm run build-font && npm run inline-icon-fonts",
     "copy-codicons": "node scripts/copy-codicons.mjs",
     "inline-icon-fonts": "node scripts/inline-icon-fonts.mjs",
-    "esbuild-base": "npm run copy-codicons && npm run build-font && npm run inline-icon-fonts && tsc --noEmit --project tsconfig.json && esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
+    "esbuild-base": "npm run copy-codicons && tsc --noEmit --project tsconfig.json && esbuild ./src/extension.ts --bundle --outfile=dist/extension.js --external:vscode --format=cjs --platform=node",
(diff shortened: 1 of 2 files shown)


fizzboop (approved):
This makes sense to me. :shipit:

PR #4207, opened 2026-05-22
feat: externalbrowser auth for snowflake

### LINKED ISSUE #3466: feat: support publishing to snowflake SPCS with token

christierney:
## Intent

Add support for browser-based authentication to snowflake.

Closes #3466

## Type of Change

- - [ ] Bug Fix 
- - [x] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

[`snowflake-sdk`](https://github.com/snowflakedb/snowflake-connector-nodejs) already implements externalbrowser auth. However it does not give us direct access to a token. But, by analogy with https://docs.snowflake.com/en/developer-guide/snowpark-container-services/tutorials/advanced/tutorial-8-access-public-endpoint-programmatically#option-3-send-requests-to-the-service-endpoint-programmatically-by-using-a-session-token (using an internal function on the python connector to get a session token) I figured out that we can get and use a session token from the `serialize()` utility. This is not guaranteed to be the same across versions so we will need to be careful about upgrading this library.

`snowflake-sdk` has built-in connection caching to prevent repeated browser auth flows. By default it writes tokens to disk, but you can supply your own credential manager. I implemented one that is backed by vscode SecretStorage.

## User Impact

Users with a snowflake connection configured to use browser auth can now use it to log in to snowflake for publishing to SPCS Connect.

```toml
[default]
account = "duloftf-posit-software-pbc-dev"
user = "chris@posit.co"
authenticator = "externalbrowser"
```

If you initiate the new credential flow and enter a snowflake URL, we initiate the browser auth flow automatically.

## Automated Tests

Added unit tests of the new code, mocking SecretStorage and snowflake-sdk.

## Directions for Reviewers

If you have a working snowflake connection, you can change your authenticator type to "externalbrowser" and add a new credential for / publish to https://kf2jaayqaqc-duloftf-posit-software-pbc-dev.snowflakecomputing.app/.

## Checklist

### DIFF
--- a/extensions/vscode/src/snowflake/secretStorageCredentialManager.ts
+++ b/extensions/vscode/src/snowflake/secretStorageCredentialManager.ts
@@ -1,0 +1,30 @@
+// Copyright (C) 2026 by Posit Software, PBC.
+
+import { SecretStorage } from "vscode";
+
+const KEY_PREFIX = "snowflake-token:";
+
+// Snowflake SDK's credential manager validation requires methods to be own properties
+// of the object (checked via Object.hasOwnProperty), not inherited from the prototype.
+// We use property assignments in the constructor to ensure they pass validation.
+export class SnowflakeSecretStorageCredentialManager {
+  readonly write: (key: string, token: string) => Promise<null>;
+  readonly read: (key: string) => Promise<string | null>;
+  readonly remove: (key: string) => Promise<null>;
+
+  constructor(private readonly secrets: SecretStorage) {
+    this.write = async (key: string, token: string): Promise<null> => {
+      await this.secrets.store(KEY_PREFIX + key, token);
+      return null;
+    };
+
+    this.read = async (key: string): Promise<string | null> => {
+      return (await this.secrets.get(KEY_PREFIX + key)) ?? null;
+    };
+
+    this.remove = async (key: string): Promise<null> => {
+      await this.secrets.delete(KEY_PREFIX + key);
+      return null;
+    };
+  }
+}
--- a/extensions/vscode/src/snowflake/tokenProviders.ts
+++ b/extensions/vscode/src/snowflake/tokenProviders.ts
@@ -189,6 +189,7 @@ class ExternalBrowserTokenProvider implements TokenProvider {
       account: this.account,
       username: this.user,
       authenticator: "EXTERNALBROWSER",
+      clientStoreTemporaryCredential: true,
     });
 
     await conn.connectAsync();
--- a/extensions/vscode/src/snowflake/tokenProviders.ts
+++ b/extensions/vscode/src/snowflake/tokenProviders.ts
@@ -5,6 +5,7 @@ import fs from "fs";
 
 import axios from "axios";
 import jwt from "jsonwebtoken";
+import snowflake from "snowflake-sdk";
 
 import type { SnowflakeConnectionConfig } from "./types";
 
@@ -15,7 +16,7 @@ export interface TokenProvider {
 
 /**
  * Creates a TokenProvider based on the authenticator field of the connection config.
- * Supported authenticators: "snowflake_jwt", "oauth" (case-insensitive).
+ * Supported authenticators: "snowflake_jwt", "oauth", "externalbrowser" (case-insensitive).
  */
 export function createTokenProvider(
   connection: SnowflakeConnectionConfig,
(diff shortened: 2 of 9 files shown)


extensions/vscode/src/snowflake/tokenProviders.ts
```diff
@@ -166,3 +169,52 @@ class OAuthTokenProvider implements TokenProvider {
...
+  private readonly account: string;
+  private readonly user: string;
+
+  constructor(connection: SnowflakeConnectionConfig) {
+    this.account = connection.account;
+    this.user = connection.user;
+  }
+
+  async getToken(_hostname: string): Promise<string> {
+    const conn = snowflake.createConnection({
```
christierney, line 188: the credential manager (above) and `clientStoreTemporaryCredential` here means the user doesn't get a new browser flow every time, but it still seems not great to me to create a new connection every time we need a token. This actually applies to the JWT and OAuth flows two (both making repeated API calls to snowflake). I'm going to follow this up with a separate PR to address that.

christierney:
follow-up: #4208

extensions/vscode/src/snowflake/secretStorageCredentialManager.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2026 by Posit Software, PBC.
+
+import { SecretStorage } from "vscode";
+
+const KEY_PREFIX = "snowflake-token:";
```
zackverham, line 5: is the KEY_PREFIX something that we are setting our own convention for? Or is this a snowflake thing?

extensions/vscode/src/snowflake/tokenProviders.ts
```diff
@@ -166,3 +169,52 @@ class OAuthTokenProvider implements TokenProvider {
...
+
+    const sessionToken = extractSerializedToken(JSON.parse(conn.serialize()));
+
+    if (!sessionToken || typeof sessionToken !== "string") {
+      throw new Error(
+        "missing session token in externalbrowser connection state",
+      );
+    }
+
+    // Do NOT call conn.destroy() — it sends a logout request to the server
```
zackverham, line 205: does this do anything weird w/ connection lifecycles? Like are we leaving stale / hanging connections floating around that will cause issues?

extensions/vscode/src/snowflake/tokenProviders.ts
```diff
@@ -166,3 +169,52 @@ class OAuthTokenProvider implements TokenProvider {
...
+    // and invalidates the session token we just extracted.
+
+    return sessionToken;
+  }
+}
+
+/**
```
zackverham, line 212: I think the documentation here is sufficient - just thinking through the implications: 

Say there's some requirement where we * have * to update the SDK version, but snowflake also removes this internal utility that makes browser auth work. Would we just have to announce that external auth is broken on a particular release of publisher and ask for patience while we sort things out?

zackverham (approved):

extensions/vscode/src/snowflake/secretStorageCredentialManager.ts
christierney: it's to distinguish the keys we are storing in vscode secret storage from other unrelated keys. See e.g.:

https://github.com/posit-dev/publisher/blob/7d32fa1df75ae3388043afff06b26bb139f1fb86/extensions/vscode/src/credentials/storage.ts#L8
https://github.com/posit-dev/publisher/blob/7d32fa1df75ae3388043afff06b26bb139f1fb86/extensions/vscode/src/credentials/storage.ts#L72-L82

extensions/vscode/src/snowflake/tokenProviders.ts
christierney: As far as I can tell, the session will eventually time out. I don't think it leaves any actual "connection" as in an open network connection, I think it is just that the server has not been told to invalidate the session token.

As part of the followup work though I think it would make sense to call destroy() when we close the extension.

extensions/vscode/src/snowflake/tokenProviders.ts
christierney: Ideally as part of the change to update the SDK we'd also find a new workaround (or better yet, snowflake would introduce a supported solution). Otherwise, yes, we'd have to make the call to break externalauth, if updating the sdk were that important.

Issue #4210, opened 2026-05-22
integration tests for SPCS


christierney:
Right now the code for authenticating with Snowflake is only end-to-end tested manually. It has been prone to breaking in the past. We should talk to the platform-partnerships team about setting up a test account that we can use in CI.

I think a single test that adds a (or just uses an existing) credential and deploys a simple app would be sufficient.

We could do this for every auth type but, especially if we consolidate all the auth modes to use the snowflake-sdk, one would probably be sufficient.

PR #4215, opened 2026-05-26
refactor: cache snowflake connections

christierney:
## Intent

Consolidate all Snowflake auth mechanisms into one path using session tokens from `snowflake-sdk` (official nodejs client library) and add caching so we do not constantly make new token requests.

## Type of Change

- - [ ] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [x] Refactor 
- - [ ] Tooling 

## Approach

Deleted the custom code for generating JWT and exchanging OAuth tokens. Deleted the TokenProvider interface/implementations and moved the snowflake token code into CredentialsService (to make caching easier).

_NOTE: this changes the way the JWT flow works. Now it uses session tokens instead OAuth access tokens. This makes it consistent with the other auth flows, and my testing shows it still works for publishing._

Added a cache to CredentialsService guarded by a mutex. Now when we fetch a snowflake token (multiple times per deployment and multiple times when the extension refreshes), we will reuse an existing snowflake connection and its underlying token. We check the connection for validity and replace it if it is no longer valid. This works alongside snowflake-sdk's own caching/refreshing logic; the `isValidAsync()` call transparently refreshes the token if possible .

_NOTE: the single mutex means that, for a user with multiple snowflake credentials, every getSnowflakeToken call will block behind the user logging in, the client making a network call to snowflake, etc. I think this is fine: most users will probably have 0 or 1 snowflake credential; if they have more than one, either they will be browser-based (so we need to wait on the user to log in anyway, and doing them one at a time makes sense) or they will be key or token based so the user will not notice they are connecting one at a time. I think this is better than having more complicated locking logic._

Added better error handling and timeouts around snowflake code.

_NOTE: I opted not to make anything `Disposable` or try to `destroy()` all our cached connections at shutdown. It would potentially slow down extension shutdown for, I think, little benefit. [Snowflake docs](https://docs.snowflake.com/en/developer-guide/node-js/nodejs-driver-connect#terminating-a-connection) don't make this sound like a requirement._
 
## User Impact

Should be transparent, but the number of network calls out to Snowflake will be greatly reduced.

## Automated Tests

Added lots of automated test coverage.

I have manually tested all three styles of snowflake auth and did my best to test the caching behavior (I don't have a good way to manually invalidate tokens, but I did some tests where I left the extension running for hours and came back to deploy again.)

## Directions for Reviewers

This might be easier to review one commit at a time.

## Checklist

Not user-facing.


extensions/vscode/src/credentials/service.ts
```diff
@@ -294,4 +252,293 @@ export class CredentialsService {
...
+  }
+
+  /**
+   * Destroys a Snowflake connection, resolving once the SDK reports completion
+   * or {@link SNOWFLAKE_DESTROY_TIMEOUT_MS} elapses — whichever comes first.
+   * Never rejects: destroy is best-effort cleanup of an already-invalid
+   * connection, so a failure or hang is logged and swallowed rather than
+   * propagated to the caller (who is about to rebuild the connection anyway).
+   */
+  private destroyConnection(
```
zackverham, line 348: if we destroy the connection, does that also invalidate the cache key? Should we clean up something in the cache here as well?

extensions/vscode/src/credentials/service.ts
```diff
@@ -294,4 +252,293 @@ export class CredentialsService {
...
+        }
+
+        if (isValid) {
+          logger.debug(`Snowflake connection cache hit for ${cacheKey}`);
+          return cached;
+        }
+
+        // Connection is no longer valid - destroy it and remove from cache
+        await this.destroyConnection(cached, cacheKey);
+        this.snowflakeConnectionCache.delete(cacheKey);
```
zackverham, line 429: ah, I see - we delete after the async destroyConnection call. That makes sense.

extensions/vscode/src/credentials/service.ts
```diff
@@ -294,4 +252,293 @@ export class CredentialsService {
...
+      throw new Error(
+        `missing session token in ${authenticator} connection state`,
+      );
+    }
+
+    return sessionToken;
+  }
+
+  /**
+   * Extracts the session token from the private state returned by
```
zackverham, line 492: do we have any guards in place that stop us from accidentally unpinning the SDK and breaking this? I guess probably typescript checks would fail if the SDK internal no longer existed.

If the internal SDK implementation exists but breaks in some way, do you think our tests would catch that?

extensions/vscode/src/credentials/service.test.ts
```diff
@@ -572,3 +578,626 @@ describe("connectAPIOptionsFromCredential", () => {
...
+      });
+
+      // Start two concurrent calls
+      await Promise.all([
+        service.getSnowflakeToken(testConnection),
+        service.getSnowflakeToken(testConnection),
+      ]);
+
+      // Connection should only be created once (mutex ensures serial access)
+      expect(snowflake.createConnection).toHaveBeenCalledTimes(1);
```
zackverham, line 823: does this only trigger once because we hit the cache?

extensions/vscode/src/snowflake/sdkConfig.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { configureSnowflakeSDK } from "./sdkConfig";
+
+describe("configureSnowflakeSDK", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("configures snowflake-sdk exactly once with a credential manager", () => {
+    configureSnowflakeSDK(new mockSecretStorage());
```
zackverham, line 20: would it be worth calling the configure function multiple times to validate that the singleton is working as expected?

zackverham (approved):
All looks great to me! New approach makes sense.

extensions/vscode/src/credentials/service.ts
christierney: yeah, in fact there originally wasn't even a separate destroy function but I needed to add error handling and timeouts.

extensions/vscode/src/credentials/service.ts
christierney: Not at this time. :/ We need to figure out e2e tests. #4210

extensions/vscode/src/credentials/service.test.ts
christierney: correct, the second getSnowflakeToken call is a cache hit

extensions/vscode/src/snowflake/sdkConfig.test.ts
christierney: no, in fact the test comment is a little misleading here---nothing prevents it from being called more than once, it's just documented that you shouldn't.

christierney: I'm updating the test name; I don't think we need to enforce in code that this only gets called once---convention is enough. (Calling it twice wouldn't break anything, it would just be unnecessary)

PR #4220, opened 2026-06-02
feat: deploy markdown files via Quarto

### LINKED ISSUE #4219: feat: deploy markdown
If I have a .md file, Publisher doesn't know what to do with it. I can't publish it. I should be able to render it with Quarto right?

zackverham:
## Summary
- Adds `.md` to the recognized entrypoint file extensions so markdown files appear in the deployment picker
- Adds `.md` to the Quarto detector's suffix list so it's inspected as Quarto content
- Adds `.md` fallback handling (uses `markdown` engine, no Python/R needed) when `quarto inspect` is unavailable

Closes #4219


zackverham:
Just validated this - seems to work fine. Going to open it for full review. 

Screenshot 2026-06-05 at 2 31 33 PM

zackverham:
@claude lets add a test for uppercase entrypoint handling as well.

extensions/vscode/src/inspect/detectors/quarto.ts
```diff
@@ -89,18 +89,23 @@ export class QuartoDetector implements ContentTypeDetector {
...
         configs.push(cfg);
       }
+    } else {
+      const entrypointPaths = await this.findEntrypoints(baseDir);
+      for (const epPath of entrypointPaths) {
+        const cfg = await this.configFromInspect(baseDir, epPath);
+        if (cfg) {
+          configs.push(cfg);
+        }
+      }
```
dotNomad: This is a bit weird.

The only spot I can find where we call `inferType` with no entrypoint is through `inspectProject` in `selectNewOrExistingConfig.ts#getConfigurationInspections()` which only happens when we have a deployment selected, but there is an error of some kind like:

- creating a new configuration for a destination (shown when there is no matching config for a deployment)
- when the selected deployment's config is broken

So the guard in `findEntrypoints` rarely ever fires. What was the intention there?

zackverham: The intention was mainly as a forward-thinking protective change, in case our assumptions about the callers change over time - I'm happy to back it out though, it isn't strictly related to markdown support.

extensions/vscode/src/inspect/detectors/quarto.ts
```diff
@@ -89,18 +89,23 @@ export class QuartoDetector implements ContentTypeDetector {
...
     const configs: PartialConfig[] = [];
-    const entrypointPaths = await this.findEntrypoints(baseDir);
 
-    for (const epPath of entrypointPaths) {
-      const relEntrypoint = path.basename(epPath);
-      if (entrypoint && relEntrypoint !== entrypoint) {
-        continue;
-      }
+    if (entrypoint) {
+      const epPath = path.join(baseDir, entrypoint);
```
dotNomad: Since the path to not passing an entrypoint is only for the error cases of the config erroring or missing I think we can simplify this back to the logic we had and remove the `findEntrypoints` cases where we avoid generic markdown files. In those cases the user is always hitting deploy for those files. There isn't a case where the user isn't selecting that and passing the `entrypoint` down to an inspector.

zackverham:
Sanity-checked deploying both a standalone markdown document and a quarto content item. Both worked as expected.

dotNomad (approved):
Very clean! Thank you for the back and forth and discussion on this one. I think we got it to a good place. :shipit:

PR #4229, opened 2026-06-03
Fix docs deployment not triggering on new releases

### LINKED ISSUE #4228: docs didn't bump version
https://posit-dev.github.io/publisher/ - current version still shows 2.4.0, not 2.6.0.

zackverham:
## Summary

- Fixes #4228 — docs site was stuck showing version 2.4.0 because the deploy workflow never triggered automatically on releases.
- **Root cause:** The `release: [published]` trigger never fired because the Release workflow creates the GitHub release using `GITHUB_TOKEN`, and [events created by `GITHUB_TOKEN` do not trigger other workflows](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow).
- Switches to `workflow_run` on the Release workflow completion, with a guard to skip pre-release versions (odd minor numbers).

## Test plan

- [ ] Manually trigger the docs workflow (`workflow_dispatch`) to verify it still works
- [ ] On the next production release (even minor), confirm the docs workflow triggers automatically via `workflow_run`
- [ ] Verify pre-release tags (odd minor) do _not_ trigger a docs deployment

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,8 +1,9 @@
 name: Deploy docs
 
 on:
-  release:
-    types: [published]
+  workflow_run:
+    workflows: ["Release"]
+    types: [completed]
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -15,17 +16,34 @@ concurrency:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.conclusion == 'success'
     steps:
       - uses: actions/checkout@v6
```
dotNomad, line 22: Should we check out the specific `github.event.workflow_run.head_sha` using `with` and `ref` for `actions/checkout@v6`?

.github/workflows/docs.yaml
dotNomad, line 59: I had some ideas on how to simplify this and make it a bit more consistent with other workflows

The code block below does a few things:
- does two jobs `check-release` and `deploy` to separate the logic a bit
- uses `github.event.workflow_run.head_sha` when checking out so we are getting the tagged SHA
- re-uses `./scripts/is-pre-release.bash` so it matches `just pre-release` that the `release.yaml` workflow uses
- it reduces the number of `if` blocks a bit

<details>
  <summary>Code Block</summary>
  
```
jobs:
  check-release:
    runs-on: ubuntu-latest
    outputs:
      skip: ${{ steps.check.outputs.skip }}
    steps:
      - uses: actions/checkout@v6
        with:
          ref: ${{ github.event.workflow_run.head_sha }}
          # Full history/tags so is-pre-release.bash's `git describe` works.
          fetch-depth: 0

      - name: Check if pre-release
        if: github.event_name == 'workflow_run'
        id: check
        run: |
          if [ "$(./scripts/is-pre-release.bash)" = "true" ]; then
            echo "skip=true" >> "$GITHUB_OUTPUT"
            echo "Skipping docs deploy for pre-release"
          fi

  deploy:
    needs: check-release
    runs-on: ubuntu-latest
    if: >-
      needs.check-release.outputs.skip != 'true' &&
      (github.event_name == 'workflow_dispatch' ||
       github.event.workflow_run.conclusion == 'success')
    steps:
      - uses: actions/checkout@v6
        with:
          # Render the released commit, not wherever main points now.
          ref: ${{ github.event.workflow_run.head_sha }}

      - uses: quarto-dev/quarto-actions/setup@v2

      - name: Render
        run: quarto render docs
        env:
          GH_TOKEN: ${{ github.token }}

      - uses: peaceiris/actions-gh-pages@v4.1.0
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: docs/_site
          keep_files: true
```

</details>

dotNomad (approved):
Incredible!

PR #4231, opened 2026-06-03
Replace HaaLeo/publish-vscode-extension with inline publish steps

zackverham:
## Summary
- Removes the dependency on `HaaLeo/publish-vscode-extension@v2` (a Node.js-based action with runtime version constraints)
- Replaces it with inline bash steps that call `@vscode/vsce` and `ovsx` directly via `npx`
- Auth is handled via environment variables (`VSCE_PAT` / `OVSX_TOKEN`) which is how both tools natively authenticate

Proves the viability of https://github.com/posit-dev/posit-gh-actions/pull/2 - which would expose a common drop-in replacement GHA for the HaaLeo dependency.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -38,12 +38,16 @@ jobs:
           echo "extension_path=dist/${name}-$(just version).vsix" >> "$GITHUB_OUTPUT"
 
       - name: Publish to Visual Studio Marketplace
-        uses: HaaLeo/publish-vscode-extension@v2
-        with:
-          registryUrl: https://marketplace.visualstudio.com
-          pat: ${{ secrets.VSCE_PAT }}
-          extensionFile: ${{ steps.get-extension-path.outputs.extension_path }}
-          skipDuplicate: true
+        env:
+          VSCE_PAT: ${{ secrets.VSCE_PAT }}
+          EXTENSION_FILE: ${{ steps.get-extension-path.outputs.extension_path }}
+        run: |
+          set -euo pipefail
+          if [[ ! -f "$EXTENSION_FILE" ]]; then
+            echo "::error::extensionFile not found at: $EXTENSION_FILE"
+            exit 1
+          fi
+          npx --yes @vscode/vsce@2 publish --packagePath "$EXTENSION_FILE" --skip-duplicate
 
   publish-open-vsx:
     needs:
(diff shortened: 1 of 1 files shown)


zackverham:
@claude can you fix the issues you identified in code review above

.github/workflows/publish.yaml
```diff
@@ -66,8 +70,13 @@ jobs:
...
+        env:
+          OVSX_PAT: ${{ secrets.OPEN_VSX_TOKEN }}
+          EXTENSION_FILE: ${{ steps.get-extension-path.outputs.extension_path }}
+        run: |
+          set -euo pipefail
+          if [[ ! -f "$EXTENSION_FILE" ]]; then
+            echo "::error::extensionFile not found at: $EXTENSION_FILE"
+            exit 1
+          fi
+          npx --yes ovsx@0 publish "$EXTENSION_FILE" --skip-duplicate
```
dotNomad: Should this use `@1`?

.github/workflows/publish.yaml
```diff
@@ -38,12 +38,16 @@ jobs:
...
+        env:
+          VSCE_PAT: ${{ secrets.VSCE_PAT }}
+          EXTENSION_FILE: ${{ steps.get-extension-path.outputs.extension_path }}
+        run: |
+          set -euo pipefail
+          if [[ ! -f "$EXTENSION_FILE" ]]; then
+            echo "::error::extensionFile not found at: $EXTENSION_FILE"
+            exit 1
+          fi
+          npx --yes @vscode/vsce@2 publish --packagePath "$EXTENSION_FILE" --skip-duplicate
```
dotNomad: Could we use `@3`? Not sure if the CLI has changed much.

dotNomad (approved):
This looks good, but I think we should investigate the version bumps

.github/workflows/publish.yaml
zackverham: Done — bumped to `@vscode/vsce@3` (latest is 3.9.1). The CLI interface is the same for our usage (`publish --packagePath ... --skip-duplicate`).

.github/workflows/publish.yaml
zackverham: Done — bumped to `ovsx@1` (1.0.0 just shipped). The publish interface is unchanged.

.github/workflows/publish.yaml
zackverham: (claude wrote this, but useful breadcrumb)

PR #4234, opened 2026-06-05
clean up generated renv.lock

### LINKED ISSUE #4233: When Publisher creates a `renv` and `renv.lock` it is then treated as the source of truth
If a user that isn't using `renv` for a project deploys using Publisher, a `renv` folder and `renv.lock` are created.

This can be re-created with the `shinyapp` we have in `test/sample-content` by deleting the `renv` and `renv.lock` (if you have one) and deploying.

Once that `renv.lock` is created Publisher treats that as the source of truth on re-deploys which is a problem for users who expecting to not use `renv` at all.

Their `renv.lock` never gets updated by Publisher (since it was already made and is now being used without being re-made).

This causes a problem if you then add a new `library(...)` call. The `renv.lock` wont have that new library, so the re-deploy with fail.

karawoo:
## Intent

closes #4233 

When a user without renv deploys an R project, Publisher wrote `renv.lock` and `renv/` into the project directory. On re-deploys, Publisher treated that on-disk lockfile as source of truth and never regenerated it, so adding a new `library()` call caused deploys to fail with a stale lockfile.

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Added an optional `renvTargetDir` parameter to `scanRPackages` that decouples where renv runs (init/hydrate/snapshot) from where dependencies are scanned. `buildManifest` now creates an OS temp directory, passes it as `renvTargetDir`, reads the lockfile for package metadata, then deletes the temp dir. Nothing persists in the user's project.

## User Impact

The `renv.lock` and `renv/` directories no longer persist on disk. Re-deploys correctly pick up new dependencies in the project. Projects that _do_ use renv are unaffected.

Note: we will not clean up any renv.lock files that publisher previously generated before this change.

## Automated Tests

- New tests in `rPackages.test.ts` for the `renvTargetDir` parameter.
- New `publishShared.test.ts` covering the ephemeral branch

I also manually tested that the files do not persist and that adding dependencies and re-deploying works. Also tested that projects that do use renv don't have their lockfiles removed.

## Checklist

### DIFF
--- a/extensions/vscode/src/interpreters/rPackages.ts
+++ b/extensions/vscode/src/interpreters/rPackages.ts
@@ -179,18 +179,25 @@ export async function scanRPackages(
   rPath: string,
   saveName?: string,
   positronR?: PositronRSettings,
+  renvTargetDir?: string,
 ): Promise<void> {
   const lockfileName = saveName || DEFAULT_RENV_LOCKFILE;
   validateSaveName(lockfileName);
+
+  // renvTargetDir separates where renv runs (init/hydrate/snapshot) from where
+  // deps are scanned. Defaults to projectDir so existing callers are unchanged.
+  const targetDir = renvTargetDir ?? projectDir;
 
   const repoURL = repoURLFromOptions(positronR);
 
   // Normalize paths for the R script (use forward slashes)
   const normalizedProjectPath = projectDir.split(path.sep).join("/");
+  const normalizedTargetPath = targetDir.split(path.sep).join("/");
   const normalizedLockfile = lockfileName.split(path.sep).join("/");
 
   // Validate strings before injecting into R code
   validateRStringLiteral(normalizedProjectPath, "Project path");
+  validateRStringLiteral(normalizedTargetPath, "renv target path");
   validateRStringLiteral(normalizedLockfile, "Lockfile name");
   validateRStringLiteral(repoURL, "Repository URL");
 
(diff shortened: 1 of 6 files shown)


extensions/vscode/src/interpreters/rPackages.ts
```diff
@@ -179,18 +179,25 @@ export async function scanRPackages(
   rPath: string,
   saveName?: string,
   positronR?: PositronRSettings,
+  renvTargetDir?: string,
 ): Promise<void> {
   const lockfileName = saveName || DEFAULT_RENV_LOCKFILE;
   validateSaveName(lockfileName);
 
+  // renvTargetDir separates where renv runs (init/hydrate/snapshot) from where
+  // deps are scanned. Defaults to projectDir so existing callers are unchanged.
```
zackverham: what existing callers are you thinking about? Wouldn't all callers be passing renvTargetDir now?

zackverham:
Approach totally makes sense to me! Just one question about what things you're expecting to call the function without passing the renvTargetPath.

zackverham (approved):

extensions/vscode/src/interpreters/rPackages.ts
karawoo: In the case where a user explicitly clicks to generate a lockfile from the UI I think we do want to put the renv stuff in the project directory:

https://github.com/posit-dev/publisher/blob/970b2e29240ece07ff0adf3c4c4c1db2aa65040d/extensions/vscode/src/views/homeView.ts#L1242-L1246

zackverham: Ah! There's the caller - got it. Good eye.

karawoo: though the comment could be worded better, it's very specific to this change but in the future reading the code "existing callers" doesn't really make sense, they're all "existing". i'll update the wording.

PR #4260, opened 2026-06-22
fix: clarify Connect uses npm in Node.js lockfile preflight error

dotNomad:
## Intent

A publisher who builds a Node.js project with yarn or pnpm has no `package-lock.json`, so the publish preflight fails with an error that asks for `package-lock.json` without explaining that Connect specifically requires `npm`. This adds that clarification.

## Type of Change

- [x] Bug Fix

## Approach

`extensions/vscode/src/publish/connectPublish.ts`: the `missingOnDiskHint` for `package-lock.json` in the Node.js preflight now leads with "Connect installs Node.js dependencies with npm." The message is unconditional - no `yarn.lock`/`pnpm-lock.yaml` detection - so it reads correctly in every case. Only the missing-on-disk hint changed; the "not included in files" hint is untouched.

## Directions for Reviewers

Configure a Node.js deployment whose project directory has `package.json` but no `package-lock.json`, then publish. The preflight error names npm as the package manager Connect uses.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- The "missing package-lock.json" error when publishing Node.js content now states that Connect installs Node.js dependencies with npm, helping publishers who build with yarn or pnpm understand why a `package-lock.json` is required. (#4253)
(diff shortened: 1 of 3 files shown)


zackverham (approved):
lgtm - up to you on if this should end up in the changelog, its small enough that its probably fine without.

dotNomad:
> lgtm - up to you on if this should end up in the changelog, its small enough that its probably fine without.

I'm leaning toward leaving it in as a reference if we add `pnpm` or other packaging support. A bit of a before and after.

Issue #4262, opened 2026-06-23
correctly handle redirects on bundle upload


christierney:
To support streaming of large bundles, we had to disable Axios' automatic following of redirects. (#4249)

I believe prior to 2.0.0, publisher did not handle redirects (#18, #2372), and as far as I know we have not done specific testing of redirect handling with 2.x. However, this is still arguably a regression.

Other Connect clients have added custom redirect handling:

- https://github.com/posit-dev/posit-sdk-py/issues/385
- https://github.com/rstudio/rsconnect/issues/674
- https://github.com/rstudio/rsconnect/issues/1297

We should investigate whether something similar would work here.

For bundle uploads specifically, we should start a new stream from disk every time we make an upload request; we must not buffer the bundle in memory nor drop the POST body on retry.

joshyam-k:
@christierney is this one that you're going to work on next? trying to decide whether to move this to your teams board or keep it in intake. Thanks!

christierney:
I can take this one.

christierney:
Closing this as wontfix after investigating. Summary: the redirect cases that motivated
this issue are origin changes, and forwarding `Authorization` across an origin change is
the wrong behavior — the fix for the real-world scenarios belongs at credential-creation
time (#4315), not in publish-time redirect following.

## What actually happens today

Every Publisher request to Connect carries `Authorization` (API key or signed token).
On `main`, only the bundle upload sets `maxRedirects: 0` (to keep axios on its native
transport so large bundles truly stream — see #4249); every other request goes through
axios's default `follow-redirects` transport. So when a stored credential URL redirects:

- **Host change** (e.g. a server migration): `follow-redirects` strips `Authorization`
  when the redirect crosses to a different, non-subdomain host, so the very first
  publish request (`GET /__api__/v1/user`) follows the redirect unauthenticated, gets a
  401, and the publish fails at "Testing authentication." We never reach the upload.
- **http→https upgrade on the same host** (Connect's `[HTTPRedirect]`):
  `follow-redirects` preserves `Authorization` on an https upgrade, so the preflight
  requests all succeed and the publish fails at the bundle upload with an opaque
  `Unexpected response from the server (30x: ...)`.
- Token-auth (signed) credentials break on any path-changing redirect regardless,
  because the request signature is computed once over the original path.

## Why redirect following isn't the right fix

Following these redirects "correctly" would mean replaying the streamed bundle body
*and* re-supplying `Authorization` to a new origin. The reference implementations
[...]

PR #4263, opened 2026-06-23
refactor: bundles are staged to tmp file and streamed during upload

### LINKED ISSUE #4249: Cannot upload 4.5 GiB bundle
## Problem statement

Trying to uplpad an application whose total uncompressed size is 4.5 GiB

## Steps to reproduce

1. Create a small/simple pyshiny app
2. Add some large random data files and include them in the application's files when publishing, such that total size is 4.5 GiB
3. Attempt to publish to Connect on-prem or Connect Cloud

**Expected behavior:** 

Application is deployed

**Actual behavior:** 

Fails at "Create Bundle" step reliably with the below error

Image

## System Details

```
## System Information

- **Posit Publisher Version**: 2.6.0
- **Platform**: linux x64
- **IDE**: Positron 1.106.0
```

## Publishing information

- **Content type:**  Pyshiny app, likely not important though

## Publisher output and logs

```
2026-06-16T16:53:50Z Collecting package descriptions
2026-06-16T16:53:50Z Local Quarto not in use
2026-06-16T16:53:50Z Local R not in use
2026-06-16T16:53:50Z Local Python version 3.13.6
2026-06-16T16:53:50Z Checking configuration against server capabilities
2026-06-16T16:53:50Z Testing authentication
2026-06-16T16:53:51Z Publishing with credentials
2026-06-16T16:53:51Z Checking server capabilities
2026-06-16T16:53:51Z Verifying existing content
2026-06-16T16:53:51Z Verifying content is not locked
2026-06-16T16:53:51Z Verifying app mode is the same
2026-06-16T16:53:51Z Configuration OK
2026-06-16T16:53:51Z Preparing files
2026-06-16T16:53:51Z Creating bundle from source directory
2026-06-16T16:55:33Z Bundle includes
```

christierney:
## Intent

Don't buffer bundle files in memory. Instead, write them to disk (tmpdir) and stream them during the HTTP POST.

File checksum for token auth is calculated at the same time as the bundle is written to disk, avoiding the need to process the stream twice.

Automatic redirect-following is disabled to avoid axios buffering the file in memory anyway. We should follow up with custom redirect handling (see rsconnect, posit-sdk-py for examples).

Fixes #4249

## Type of Change

- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach

Instead of writing the bundle file to a buffer in memory and passing that around, write it to a tmp file on disk. In case of error, delete the tmp file (but the OS should clean it up if it is missed).

During the write, also calculate the md5sum, which we need for token auth requests.

At upload time, instead of passing the in-memory buffer, read the file from disk.

Set axios `maxRedirects: 0`, because otherwise it uses the `follow-redirects` transport and buffers the file in memory anyway, leading to an OOM error at a later stage of the deploy. [Axios docs](https://axios.rest/pages/advanced/file-posting.html) confirm:

> DANGER
> 
> When uploading a readable stream in Node.js, set maxRedirects: 0 to prevent the follow-redirects package from buffering the entire stream in RAM.

Filed #4262 to track. I think this is an acceptable regression because publisher has previously not handled redirects, and it's not clear to me that the automatic axios handling was correct for this use case either. (I don't think we ever tested this.)

While testing I noticed that cancellation (via an `AbortSignal`) was not properly threaded through all the create/upload bundle steps, so I fixed that too. Now you can cancel during bundle creation or bundle upload (Connect and Cloud paths) and the operation will cancel and the temp file will get cleaned up immediately.

## User Impact

Users can now upload large bundles without Publisher erring. Whether Connect or Connect Cloud times out or errs is another question, but I was at least able to upload larger bundles than before in my own testing.

## Automated Tests

Updated existing tests and added new tests for bundle creation and cleanup.

## Manual Tests

I created a simple test app with a large (4.5G) random data file. (Random to avoid effective gzip compression.) Prior to this change, Publisher would err as in the linked issue. After this change, I successfully uploaded the file to Connect. Cloud eventually returned an error, but it was a rejection of the file size by the server, not a Publisher problem:
Screenshot 2026-06-22 at 16 21 51

I also observed a Connect instance (Dogfood) fail during a long upload with a Gateway Timeout, which Publisher displayed correctly.

## Checklist


extensions/vscode/src/publish/connectPublish.ts
dotNomad, line 618: After the bundle upload is successful (here and in `connectCloudPublish.ts` we could remove the bundle temp directory immediately. We don't need it after this point, and the rest of the publish flow can take a bit of time.

I don't think this is strictly necessary, but thinking out loud about the option.

extensions/vscode/src/publish/connectPublish.ts
```diff
@@ -810,6 +825,13 @@ export async function connectPublish({
       // Don't mask the original error
     }
     throw err;
+  } finally {
+    // Remove the streamed bundle's temp directory on every exit path.
+    if (bundleTmpDir) {
+      await fs.promises
+        .rm(bundleTmpDir, { recursive: true, force: true })
+        .catch(() => {});
+    }
```
dotNomad: Prior to removing the bundle temp directory we should probably destroy the bundle stream we create with `fs.createReadStream`. If something goes wrong with uploading the bundle we could hit an issue where we cannot delete the temp directory because it is still being read / is open.

We could also use `maxRetries` and `retryDelay` on the `rm` call to assist with this possible edge case.

The same suggestion goes for `connectCloudPublish.ts` too.

dotNomad: Additionally since the cleanup is the same for both call sites I wonder if we could create a `cleanupBundle` helper or something to remove some duplication.

packages/connect-api/src/auth.ts
```diff
@@ -70,3 +71,23 @@ export function signRequest(
     "X-Auth-Signature": signature,
   };
 }
+
+/**
+ * Computes per-request authentication headers for Connect token auth.
+ * Returns an object with Date, X-Content-Checksum, X-Auth-Token, and X-Auth-Signature headers.
+ */
+export function signRequest(
```
dotNomad, line 79: Nice re-use here 👍

extensions/vscode/src/publish/publishShared.ts
```diff
@@ -390,9 +396,15 @@ export async function buildBundleArchive(
...
+      signal,
     });
 
-    return { bundle: result.bundle, manifest: result.manifest };
+    return {
+      bundlePath: result.bundlePath,
+      bundleSize: result.bundleSize,
+      bundleChecksum: result.bundleChecksum,
+      manifest: result.manifest,
+    };
```
dotNomad: Any reason not to return the entire `result`? We could re-use the `BundleResult` type and consumers of this could extract the values they want.

extensions/vscode/src/bundler/archive.ts
```diff
@@ -26,102 +35,192 @@ export async function createArchive(
...
+      signal,
+    );
+  } catch (err) {
+    // Don't leak a partial bundle if archive creation fails — including when
+    // the deployment is canceled mid-creation, which makes writeArchive throw.
+    // Cleanup is best-effort — never let a removal failure mask the original
+    // error.
+    await fs.promises
+      .rm(tmpDir, { recursive: true, force: true })
+      .catch(() => {});
```
dotNomad, line 71: Does this delete the entire temp directory? Do we want to do that, or just remove the single directory we created?

dotNomad (commented):
I had a few suggestions and a couple questions

extensions/vscode/src/publish/connectPublish.ts
christierney: Yes, good call.

extensions/vscode/src/publish/publishShared.ts
christierney: No particular reason!

extensions/vscode/src/bundler/archive.ts
christierney: `tmpDir` is the result of `mkdtemp`, which creates a unique directory with a random suffix, e.g. `/tmp/posit-publisher-bundle-abc123`. It's specific to this bundle.

extensions/vscode/src/publish/publishShared.ts
dotNomad: Could be an easy cleanup then. Totally fine to ignore for this PR though

extensions/vscode/src/bundler/archive.ts
dotNomad: Ah okay I just misread this 👍 Thank you for the clarification

dotNomad (approved):
This looks great! Thank you for those adjustments 🎉

PR #4267, opened 2026-06-25
fix: add required field to test to fix typecheck failure

christierney:
Fix a typecheck failure that keeps tripping up Claude. Also add recipes so Claude will stop trying to run `npx tsc` directly.


dotNomad (approved):
Easy approve

PR #4283, opened 2026-07-06
fix(e2e): match Positron activitybar overflow button by aria-haspopup

dotNomad:
Newer Positron builds changed the button's `role` from `button` to `tab`, so the selector stopped matching.


m-- (approved):
thank you!

PR #4303, opened 2026-07-22
fix: follow HTTP redirects client-wide, including bundle upload (#4262)

### LINKED ISSUE #4262: correctly handle redirects on bundle upload

christierney:
## Summary

Fixes #4262.

PR #4263 (fixing #4249) made bundle uploads stream from a staged temp file instead of buffering in memory, and set `maxRedirects: 0` on the two upload requests so axios stays on its native http/https transport that truly streams (its default `follow-redirects` transport buffers the whole body in RAM to replay on a redirect, which would reintroduce the OOM #4249 fixed). The regression: a 3xx on upload now surfaces as a raw `Request failed with status code 302` — triggered by `http://`→`https://` upgrades and renamed hosts (e.g. `colorado.rstudio.com` → `colorado.posit.co`).

It also turned out that **every other** Connect request still used axios's `follow-redirects` transport, which on 301/302 rewrites POST→GET, drops the body, and never re-signs the token-auth signature against the redirected path. So redirect handling was broken client-wide, not just for uploads.

## Approach

Manual, client-wide redirect handling in both `@posit-dev/connect-api` and `@posit-dev/connect-cloud-api`:

- A small `redirects.ts` module (duplicated per package — they share no dependency) installs an axios **response error interceptor**. Each instance is created with `maxRedirects: 0` (native streaming transport); a 3xx rejects via `validateStatus` and the interceptor re-issues the request **through the same instance**, so token signing and cookie forwarding re-run per hop.
- Follows **301/302/307/308**, preserving method and body on all of them (libcurl `CURLOPT_POSTREDIR` semantics, matching posit-sdk-py). **303** and Location-less 3xx are deliberately not followed — the original error propagates. Hop limit is **5**, after which a descriptive error names the last Location.
- **Credentials are forwarded** across cross-origin redirects. The redirect target comes from the user's own configured server, so this matches the sibling clients (rsconnect, posit-sdk-py). `http`→`https` scheme upgrades and host renames both work.

## Stream-factory approach (never buffer the bundle)

`uploadBundle` now takes a stream **factory** `() => Readable` instead of a `Readable`. On each redirect hop the interceptor destroys the consumed stream and calls the factory to open a **fresh read stream from disk** — the bundle is re-streamed, never held in memory and never dropped. The API client owns the stream lifecycle end-to-end, so the extension's `cleanUpBundle` now only removes the temp directory. Connect Cloud's presigned-URL upload gets the same treatment via a dedicated redirect-following instance (S3-style presigned endpoints can region-redirect).

Cancellation is unchanged: the original `AbortSignal` rides on the config through every hop, and axios rejects promptly on abort.

## Testing

- `packages/connect-api`: 157 tests (new `redirects.test.ts` + a full redirect matrix through `ConnectAPI` — GET/POST across 301/302/307/308, relative/absolute/cross-origin Location, token re-signing against the target pathname, fresh-stream-per-hop, hop limit, 303, missing Location, abort mid-chain, non-redirect errors). Typecheck clean.
- `packages/connect-cloud-api`: 93 tests incl. instance + presigned-upload redirect coverage. Typecheck clean.
- `extensions/vscode`: full unit suite (1844) green; lint clean.
- `test/connect-api-contracts`: 69 contract tests green (normal-path request shape unchanged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed deploying large projects (multiple GiB) failing at the "Create Bundle" step. The bundle is now streamed to disk and uploaded as a stream instead of being held in memory, so bundles are no longer limited by available memory. (#4249)
+- Publisher now follows HTTP redirects when talking to Connect, including during bundle upload — for example when a server redirects `http://` to `https://` or has moved to a new hostname. Bundle uploads re-stream from disk on each redirect, so large bundles still never need to be held in memory. (#4262)
 - Turning off `positPublisher.verifyCertificates` now skips certificate checks as expected, so you can deploy to Connect servers with self-signed or otherwise untrusted certificates. (#4265)
 
 ## [2.8.0]
--- a/extensions/vscode/src/publish/connectCloudPublish.test.ts
+++ b/extensions/vscode/src/publish/connectCloudPublish.test.ts
@@ -168,15 +168,6 @@ vi.mock("../toml/configCompliance", () => ({
   forceProductTypeCompliance: vi.fn(),
 }));
 
-// stream/promises.finished listens for the stream 'close' event which fires
-// via process.nextTick — not faked by vi.useFakeTimers(). If finished() were
-// real, cleanUpBundle's extra await would fire after vi.runAllTimersAsync()
-// returns, causing the log-watching timer setup to miss the run-all window.
-// Mock it to resolve immediately so the async chain stays in sync.
-vi.mock("stream/promises", () => ({
-  finished: vi.fn().mockResolvedValue(undefined),
-}));
-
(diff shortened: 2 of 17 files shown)


hekhuisk (approved):

christierney:
rethinking this approach

Issue #4204, opened 2026-05-21
feat: remove token/key choice in new credential flow


christierney:
Now that token authentication works everywhere[^1], it would be simpler and less confusing to remove the "Token auth or API Key" choice in the new credential flow and just default to token. We have heard that customers are sometimes confused by the choice and think we don't support token auth, even though it is present as a choice and just works.

We need to continue to support API Key credentials that already exist.

We may want to consider having some way to still make a new API Key credential as a fallback in case somebody finds a situation where Token auth is not working. Maybe a separate command they can invoke from the command palette?

[^1]: We think. Verify that this is true! Search for open issues. We have tested and proved it works in Snowflake and Workbench. Try it with a proxy setup.

nealrichardson:
I'd personally love to see this accompanied by #3878. So all the user has to do is give you the Connect server URL, if it can do OAuth it does, if it can't, it just does the token auth as you say here.

Issue #3858, opened 2026-03-20
Add integration test for token auth against a real Connect server


zackverham:
Tracked in #3713.

The TypeScript `ConnectAPI` client now supports token-based authentication (RSA key-pair signing), but there is no integration or E2E test that validates the full token auth handshake against a real Connect server.

## Context

PR #3857 ported Go token auth to the TypeScript client with solid unit test coverage (signature round-trip verification, header computation, constructor validation). However, an integration test against a live Connect server would confirm the full flow works end-to-end:

1. Provision a token credential (RSA key pair + token ID) on a Connect server
2. Authenticate using the TypeScript `ConnectAPI` client with `token` + `privateKey`
3. Make authenticated API calls (e.g., `testAuthentication()`, `getSettings()`)
4. Verify the requests succeed with proper authorization

## Considerations

- May require Docker/Connect server configuration changes to support token provisioning in the E2E environment
- Could be a Cypress E2E test or a standalone integration test
- Existing E2E infrastructure only covers API key auth (PCS) and OAuth (PCC)

Issue #3709, opened 2026-03-12
E2E Cypress test gaps for deployment TOML migration


christierney:
## Context

As part of the Go-to-TypeScript deployment TOML migration (#3707), we analyzed existing e2e Cypress test coverage against the manual test plan. This issue tracks the identified gaps that would be good candidates for automation.

## Existing Coverage

| Test File | What It Covers |
|---|---|
| `common.cy.js` | Smoke test: extension loads, sections visible |
| `credentials.cy.js` | PCS/PCC credential CRUD |
| `deployments.cy.js` | PCS static deploy + PCC Shiny deploy (create → deploy → verify TOML) |
| `embedded-deployments.cy.js` | FastAPI deploys from workspace root + subdirectory |
| `error-err-config.cy.js` | Invalid config + missing config error display |
| `multi-stepper-negative.cy.js` | OAuth cancellation + deployment creation cancellation |
| `integration-requests.cy.js` | Integration Requests section visibility (PCS vs PCC) |
| `open-connect-content.cy.js` | Open Connect Content command (currently skipped/flaky) |

### High Priority

**1. Associate GUID ("Link to Existing Content")**
No existing test covers the Associate GUID flow (`showAssociateGUID`). This directly exercises the migrated `patchDeploymentRecord` code path. A test would:
- Create a deployment
- Run the Associate GUID command
- Enter a valid GUID
- Verify the deployment TOML is updated with `id`, `dashboard_url`, `direct_url`, `logs_url`

**2. Redeployment (deploy → modify → redeploy)**
Existing tests deploy once but never test the redeploy cycle. This would validate that:
- After initial deploy, the deployment record gets updated with `id` and URLs
- Modifying the config and redeploying works
- The deployment state transitions correctly (NEW → has content ID)

### Medium Priority

**3. Deployment Selection/Switching**
`error-err-config.cy.js` selects deployments from the quick pick, but there's no test that:
- Creates multiple deployments
- Switches between them via the deployment selector
- Verifies the UI updates to show the correct deployment's details

This exercises `loadDeployment` (the migrated `getSelectedContentRecord` path in `state.ts`).

**4. Deployment Record TOML Field Verification After Deploy**
`deployments.cy.js` verifies the *config* TOML fields but doesn't check the deployment record TOML after a successful deploy. Specifically:
- `id` (content GUID) is populated
- `dashboard_url`, `direct_url`, `logs_url` are populated and well-formed
- `server_url` and `server_type` match the credential used

[...]

Issue #3376, opened 2026-01-07
feat: Integrate credential storage with `rsconnect-python`


juleswg23:
## Description

This is possibly an over-ambitious request, but I am trying to replicate behavior from the publishing pipeline in RStudio.

I'd like to be able to create a new credential at the command line with `rsconnect-python`, and have Publisher pick this up when I go to deploy. The motivation is that I find it easier to onboard users with an instruction to execute, rather than a series of steps and screenshots for use of the UI, especially for an operation they will only have to do once.

From my understanding, one solution to this may make credential storage less secure, given that both tools (Publisher and rsconnect-python) would have to be able to read and/or write the credential independently.

In the off chance this is already supported I apologize!

dotNomad:
I think if we did this it pulls us in a direction of a universal credential store for Connect publishing.

If we pull in `rsconnect-python` creds, but now those don't sync with Publisher creds (because it uses `SecretStore` in Positron / VS Code now `rsconnect-python` has older, different credentials which could be confusing.

I think its better to keep these credentials as independent for now.

PR #4314, opened 2026-07-27
feat(auth): Connect OAuth authentication (#3878)

### LINKED ISSUE #3878: Use Connect's OAuth flow to authenticate
Rather than token auth or API key. Check for existence of the `$CONNECT_SERVER/.well-known/oauth-authorization-server`, and if it exists, do the OAuth flow and store the credential. Also probably need to be able to detect when you need to re-trigger the auth, like catch the error when the token is expired and re-auth if using this auth method.

mconflitti-pbc:
Closes #3878.

## What

Adds OAuth sign-in for self-hosted Posit Connect credentials, mirroring the mechanism shipped in `rsconnect-python` (`rsconnect/oauth.py`): RFC 8414 discovery, RFC 7591 dynamic client registration, Authorization Code + PKCE, device code, and bearer tokens with refresh.

When creating a Connect credential, the auth-method picker shows **two options**:

- **Sign in with a browser** (Recommended)
- **API Key** (manually enter an API key)

"Sign in with a browser" routes to **OAuth** when the server advertises it (via the `.well-known` probe) and to the existing **token** flow when it doesn't. The user never sees the OAuth-vs-token distinction — both open the browser and no token is ever shown. **Existing API-key / token+key credentials are untouched** (no migration).

## Connect version requirements

- **OAuth** (discovery + DCR + Auth Code/PKCE + refresh) requires **Posit Connect 2026.02.0+**. Against older servers, the browser sign-in transparently uses the existing token flow.
- The **device-code** flow requires **Posit Connect 2026.06.0+** (the device authorization endpoint).

## How it works

- **Discovery probe** on the entered URL (`/.well-known/oauth-authorization-server`) decides whether the browser sign-in uses OAuth or the token flow.
- **Redirect transport is chosen by environment, not by socket bind.** `ConnectOAuthActivator.authorize()` tries, in order:
  1. **`positPublisher.useDeviceCodeAuth`** — explicit user override, honored first so it always wins.
  2. **Posit Workbench relay** (`RS_SERVER_URL` + `RS_SERVER_ADDRESS` set) — real Auth Code + PKCE in browser-served VS Code / Positron. See below.
  3. **Loopback** (`http://127.0.0.1:<port>/callback`) — desktop UI with no `remoteName`, i.e. extension host and browser share a machine. Zero Connect config: DCR accepts a loopback redirect and authorize matches it with RFC 8252 port flexibility.
  4. **Device code** — the universal fallback.

  Client registration moved *into* each flow, since each registers its own `redirect_uri`.
- **Bearer + refresh** live in `@posit-dev/connect-api`: a `BearerAuth` option sends `Authorization: Bearer`, and on a 401 calls a caller-supplied refresh callback, retries once, else throws `SessionExpiredError`. Streamed bodies (bundle upload) are never replayed.
- **Single refresh/persist chokepoint:** `connectAPIOptionsFromCredential` emits the bearer variant with a callback that refreshes (re-registering on `invalid_client`, keeping the current refresh token when the server doesn't rotate it), persists rotated tokens, and returns the new access token — so all existing ConnectAPI call sites get silent refresh for free.
- **Publish re-auth:** refresh is transparent; if it fails, the deploy prompts interactive re-auth (reusing `oauthClientId`), persists, and retries once.
- **Credential name** defaults to the server hostname (editable), matching the token flow.
- **Loopback callback page** is a self-contained branded card (inlined Posit Connect logo, success/error status icon on its own line, light/dark support, friendly copy) served on `127.0.0.1` — no external resource requests.

## Browser Positron / VS Code on Posit Workbench

Workbench serves both IDEs through the browser, where a loopback socket binds fine on the extension host but is unreachable from the user's browser. Workbench exposes a **generic OAuth relay** — the same one Workbench's own VS Code extension uses (`rstudio-workbench-vscode-ext` `src/utils/oauthClient.ts`) — which makes the real Authorization Code + PKCE flow work there:

```
redirect_uri = <RS_SERVER_URL>/oauth_redirect_callback      # browser-reachable
poll         = <RS_SERVER_ADDRESS>/oauth_code?state=...     # host-reachable
```

`/oauth_redirect_callback` records `state → code` in Workbench's `oauth2_state` table on redirect; `/oauth_code?state=` returns it exactly once (404 until it lands). Both routes are registered unconditionally (`ServerOpenIDAuth.cpp:394-395`), and the mechanism is identical for Positron and VS Code.

Because the Workbench URL is not a loopback URI, **a Connect administrator must allow `https://<workbench-url>/oauth_redirect_callback`** as a redirect URI for this path to be used. Publisher handles both directions automatically:

- Probes the relay before opening any browser, so an unreachable deployment falls back to device code immediately rather than after a long poll.
- Falls back to device code when Connect rejects the Workbench redirect URI at registration, logging the URI an admin would need to allow.
- **Surfaces** a terminal failure (user denied, timed out) instead of silently re-prompting in another flow.

`positPublisher.useDeviceCodeAuth` is now documented as a **backup** toggle rather than the fix for browser IDEs, since automatic selection covers those environments.

## Credential model

Adds `oauthClientId` + `tokenExpiresAt`, mutually exclusive with `apiKey` and `token`+`privateKey`. Storage is migration-safe (older records backfill the new fields to `""`; envelope version unchanged).

## Redirect design note (why no Connect code change)

No change to Connect's *default* allowlist is needed. Allowlisting `vscode://`/`positron://` deep links would not help the browser/Workbench case: a custom scheme has no local desktop app to catch it in a browser tab; Connect's existing `https://vscode.dev/redirect` default only bounces to Microsoft's `vscode://` scheme, not forks like Positron; and `env.asExternalUri` in a web IDE resolves to a dynamic, per-deployment origin that can't be a static allowlist entry. (Explored and dismissed in now-closed posit-dev/connect#41662.) The Workbench relay solves the web case instead, using per-deployment allowlisting an admin already controls.

## Testing

- Unit (Vitest): discovery 200/404/HTML/network; DCR (+ conditional device grant, `invalid_client` re-register); PKCE S256; auth-code / device / refresh exchanges (+ refresh-token rotation kept when omitted); credential validation + option order; storage backfill; `update`/`refreshOAuthToken`; connect-api bearer header + 401→refresh→retry + refresh-failure + streamed-body guard; stepper (browser→OAuth, browser→token, API-key with/without OAuth).
- Workbench module: env detection (query/trailing-slash normalization, partial/non-HTTP rejection), redirect-URI construction, relay reachability probe (400/404/429/503 reachable; HTML/ECONNREFUSED not), code polling (retry statuses, denial, unexpected status, timeout).
- Activator transport selection: desktop→loopback, loopback-bind-failure→device, web-outside-Workbench→device, remote→device, override wins even in Workbench; Workbench relay happy path (registered redirect URI, authorize params, raw-string open, poll, exchange) plus all three fallbacks and the terminal-failure passthrough.
- `eslint src` clean, typecheck 0 errors, Prettier clean, esbuild bundle builds. Extension Vitest **107 files / 1918 pass** (+4 todo, full suite including the Quarto/R integration + smoke suites); connect-api Vitest **138 pass**.
- e2e: legacy API-key credential flow selectors switched from row index to aria-label so they're robust to the new picker.

## Out of scope / follow-ups

- Connect Cloud SSO; shared credential store with rsconnect-python (#3376).
- Fully removing the token/key choice (#4204).
- e2e/Cypress coverage for the browser-OAuth path (coordinate with #3709) and extending #3858.
- Manual smoke test against a live Workbench-served Positron/VS Code session with the Workbench redirect URI allowed on Connect (unit tests cover routing + HTTP mechanics, not a live proxy round-trip).

🤖 Draft — opening for early review.

### DIFF
--- a/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
@@ -46,7 +46,7 @@ import {
   OAuthAuthResult,
   OAuthMetadata,
 } from "src/auth/oauth";
-import { logSignInDiagnostic } from "src/logging";
+import { logger } from "src/logging";
 
 // Internal auth mechanisms. OAUTH and TOKEN are both surfaced to the user as a
 // single "sign in with a browser" choice — the user never sees a token, so the
@@ -552,12 +552,12 @@ export async function newConnectCredential(
       // an error popup — see isCancellation for why a dismissal must not take
       // this path.
       if (isCancellation(e)) {
-        logSignInDiagnostic("OAuth sign-in was dismissed before it completed.");
+        logger.debug("OAuth sign-in was dismissed before it completed.");
       } else {
         // describeError, not getMessageFromError: the latter returns "" for any
         // throw it doesn't recognize.
         const reason = describeError(e);
-        logSignInDiagnostic(`OAuth sign-in failed: ${reason}`);
+        logger.debug(`OAuth sign-in failed: ${reason}`);
         window.showErrorMessage(
           `OAuth sign-in did not complete: ${reason}. You can use an API key instead.`,
         );
--- a/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
@@ -45,6 +45,10 @@ import {
   OAuthMetadata,
 } from "src/auth/oauth";
 
+// Internal auth mechanisms. OAUTH and TOKEN are both surfaced to the user as a
+// single "sign in with a browser" choice — the user never sees a token, so the
+// distinction is an implementation detail resolved by whether the server
+// advertises OAuth.
 enum AuthMethod {
   API_KEY = "apiKey",
   TOKEN = "token",
@@ -53,20 +57,8 @@ enum AuthMethod {
 
 enum AuthMethodName {
   API_KEY = "API Key",
-  TOKEN = "Token Authentication (Legacy)",
-  OAUTH = "OAuth (Recommended)",
+  BROWSER = "Sign in with a browser",
 }
-
-const getAuthMethod = (authMethodName: AuthMethodName) => {
-  switch (authMethodName) {
-    case AuthMethodName.API_KEY:
-      return AuthMethod.API_KEY;
-    case AuthMethodName.TOKEN:
-      return AuthMethod.TOKEN;
-    case AuthMethodName.OAUTH:
-      return AuthMethod.OAUTH;
-  }
-};
 
 export async function newConnectCredential(
   viewId: string,
@@ -113,10 +105,6 @@ export async function newConnectCredential(
 
   const isApiKey = (authMethod: AuthMethod) => {
     return authMethod === AuthMethod.API_KEY;
-  };
-
-  const isOAuth = (authMethod: AuthMethod) => {
-    return authMethod === AuthMethod.OAUTH;
   };
 
   async function getSnowflakeToken(
@@ -359,20 +347,14 @@ export async function newConnectCredential(
   // Step: Select authentication method (Connect only)
   // ***************************************************************
   async function inputAuthMethod(input: MultiStepInput, state: MultiStepState) {
-    // Offer OAuth (recommended) only when the server advertised it during the
-    // URL step; otherwise present the token/API-key choices as before.
+    // Two choices, always: a browser sign-in (recommended) and a manual API
+    // key. The browser option is OAuth when the server advertises it, otherwise
+    // the token flow — both open the browser and never expose a token, so the
+    // user sees one consistent "sign in with a browser" option either way.
     const authMethods = [
-      ...(oauthMetadata
-        ? [
-            {
-              label: AuthMethodName.OAUTH,
-              description: "Sign in through your browser",
-            },
-          ]
-        : []),
       {
-        label: AuthMethodName.TOKEN,
-        description: "Generate a token in your browser",
+        label: AuthMethodName.BROWSER,
+        description: "Recommended — no API key to copy",
       },
       {
         label: AuthMethodName.API_KEY,
@@ -386,14 +368,12 @@ export async function newConnectCredential(
       totalSteps: 0,
       placeholder: "Select authentication method",
       items: authMethods,
-      // First item is the default: OAuth when available, else token auth.
+      // Browser sign-in is the default.
       activeItem: authMethods[0],
       buttons: [],
       shouldResume: () => Promise.resolve(false),
       ignoreFocusOut: true,
     });
-
-    authMethod = getAuthMethod(pick.label as AuthMethodName);
 
     // Clear auth fields from other methods so back-navigation can't leave a
     // credential with mixed material.
@@ -408,17 +388,8 @@ export async function newConnectCredential(
       state.data.privateKey = undefined;
     };
 
-    if (isOAuth(authMethod)) {
-      state.data.apiKey = undefined;
-      clearToken();
-      return {
-        name: step.INPUT_OAUTH,
-        step: (input: MultiStepInput) => steps[step.INPUT_OAUTH](input, state),
-        skipStepHistory: true,
-      };
-    }
-
-    if (isApiKey(authMethod)) {
+    if (pick.label === AuthMethodName.API_KEY) {
+      authMethod = AuthMethod.API_KEY;
       clearToken();
       clearOAuth();
       return {
@@ -428,8 +399,18 @@ export async function newConnectCredential(
       };
     }
 
-    // Token authentication.
+    // Browser sign-in: OAuth when supported, else the token flow.
     state.data.apiKey = undefined;
+    if (oauthMetadata) {
+      authMethod = AuthMethod.OAUTH;
+      clearToken();
+      return {
+        name: step.INPUT_OAUTH,
+        step: (input: MultiStepInput) => steps[step.INPUT_OAUTH](input, state),
+        skipStepHistory: true,
+      };
+    }
+    authMethod = AuthMethod.TOKEN;
     clearOAuth();
     return {
       name: step.INPUT_TOKEN,
--- a/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
+++ b/extensions/vscode/src/multiStepInputs/newConnectCredential.ts
@@ -354,7 +354,7 @@ export async function newConnectCredential(
     const authMethods = [
       {
         label: AuthMethodName.BROWSER,
-        description: "Recommended — no API key to copy",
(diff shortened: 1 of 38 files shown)


mconflitti-pbc:
This simply implements the ability to use oauth to create a credential. We can follow up with changes to the UX regarding just automatically always following this cascade: oauth -> legacy token auth -> api key. But determining the proper ergonomics of this change across all environments are weeds I didn't want to get into here. Need thoughts from @kellobri and @nealrichardson

nealrichardson:
I am excited for this. My feedback is all about the UX, much of which is already in Publisher. We can address most of it after.

One concrete idea we could consider here though: "Token Authentication (Legacy)" sounds kinda gross. What if instead of always making that be an option, you probe for the .well-known resource, and if it exists, the options are "oauth" and "api key". If it does not, your options are "token" (no legacy) and "api key". We could further conceal the specifics and present both "oauth" and "token" as "browser" auth, because "token" is purely an implementation detail, the user never actually sees a token, it feels pretty much to them like the oauth flow.

mconflitti-pbc:
> I am excited for this. My feedback is all about the UX, much of which is already in Publisher. We can address most of it after.
> 
> One concrete idea we could consider here though: "Token Authentication (Legacy)" sounds kinda gross. What if instead of always making that be an option, you probe for the .well-known resource, and if it exists, the options are "oauth" and "api key". If it does not, your options are "token" (no legacy) and "api key". We could further conceal the specifics and present both "oauth" and "token" as "browser" auth, because "token" is purely an implementation detail, the user never actually sees a token, it feels pretty much to them like the oauth flow.

yeah that is a good tradeoff imo for this first pass. will do that!

nealrichardson:
The latest commit message says something about device code flow and copying codes, that's not what we're doing right? That's not what your screen recording showed.

mconflitti-pbc:
> The latest commit message says something about device code flow and copying codes, that's not what we're doing right? That's not what your screen recording showed.

It supports both. this is following rsconnect-python's lead which allows you to pass a flag to use a device code to confirm in cases where loopback is not local. Workbench I presume? Will update the video.

mconflitti-pbc:
Video shows both flows. The extension should detect which environment its in to automatically switch to device code, but the setting is there to force it for testing and is an escape hatch since it should work in all cases.


https://github.com/user-attachments/assets/3ac8a2dc-aa1d-417c-a465-fc496174a7af

nealrichardson:
> > The latest commit message says something about device code flow and copying codes, that's not what we're doing right? That's not what your screen recording showed.
> 
> It supports both. this is following rsconnect-python's lead which allows you to pass a flag to use a device code to confirm in cases where loopback is not local. Workbench I presume? Will update the video.

Workbench should be able to do the regular browser oauth flow, I would think--it works for MCP servers on Connect.

> The extension should detect which environment its in to automatically switch to device code

What does this mean?

nealrichardson:
Otherwise I like the new flow!

mconflitti-pbc:
@nealrichardson There is this check here: https://github.com/posit-dev/publisher/pull/4314/changes#diff-8ac7b54b1c4f782516bd7d6984a65b5f1165c427460cee70d03f4274da156955R113-R124 and used here: https://github.com/posit-dev/publisher/pull/4314/changes#diff-8ac7b54b1c4f782516bd7d6984a65b5f1165c427460cee70d03f4274da156955R131-R146

I would be surprised if using the loopback address works in workbench. I guess maybe there is something else going on we need here. Not sure how to test that in workbench specificaly before merging. The extension is spinning up a local server to do the authorization callback so maybe that just works but Connect would need the public address to redirect to.

nealrichardson:
> @nealrichardson There is this check here: https://github.com/posit-dev/publisher/pull/4314/changes#diff-8ac7b54b1c4f782516bd7d6984a65b5f1165c427460cee70d03f4274da156955R113-R124 and used here: https://github.com/posit-dev/publisher/pull/4314/changes#diff-8ac7b54b1c4f782516bd7d6984a65b5f1165c427460cee70d03f4274da156955R131-R146
> 
> I would be surprised if using the loopback address works in workbench. I guess maybe there is something else going on we need here. Not sure how to test that in workbench specificaly before merging. The extension is spinning up a local server to do the authorization callback so maybe that just works but Connect would need the public address to redirect to.

In the MCP use case, the initiator of the OAuth flow includes the `redirect_uri`, which is the Workbench URL. 

It must be possible for you to install a dev version of the extension into one of our workbench instances and try it out. (And if that's hard to do, we should fix that.)

extensions/vscode/src/multiStepInputs/newConnectCredential.ts
```diff
@@ -38,26 +38,28 @@ import {
...
 } from "src/auth/ConnectAuthTokenActivator";
-
+import {
+  ConnectOAuthActivator,
+  discoverOAuthMetadata,
+  OAuthAuthResult,
+  OAuthMetadata,
+} from "src/auth/oauth";
+
+// Internal auth mechanisms. OAUTH and TOKEN are both surfaced to the user as a
```
joshyam-k, line 51: is there a situation where a user would actually want to know which path was taken?

joshyam-k (commented):
apart from the open question around WB, this looks really good and the flow looks really smooth

hekhuisk (approved):

extensions/vscode/src/multiStepInputs/newConnectCredential.ts
mconflitti-pbc: Not that I can think of tbh!

mconflitti-pbc:
Tested successfully in workbench. It still falls back to device code since our connect has not explicitly added the proper workbench oauth relay uri to its allowlist. This is okay though since device code works just as well. Will open separate ticket to address updating that and making sure it is properly documented.

So again, to confirm the cascade is:
Browser auth -> if connect supports oauth: (oauth token generation -> device code), if connect does not support oauth: legacy token method -> or user selects manual api key entry

mconflitti-pbc:
@nealrichardson does this address remaining concerns?

nealrichardson:
I still think that the fact you're struggling to get the OAuth callback working on Workbench is a sign that we're doing something wrong here, but we don't have to hold this PR to sort that out, this is definitely a huge step forward!

mconflitti-pbc:
> I still think that the fact you're struggling to get the OAuth callback working on Workbench is a sign that we're doing something wrong here, but we don't have to hold this PR to sort that out, this is definitely a huge step forward!

The missing piece is updating the config in connect.posit.it so that the workbench redirect uri is in the allowlist. This code checks for failure of that allowlist and falls back to avoid confusion.

Separate conversation started around issues around redirect that are outside the scope of this publisher implementation. It works when the allowlist is correct and falls back seamlessly to device code if not. https://positpbc.slack.com/archives/C04FPQK3H9C/p1785433295023519

PR #4319, opened 2026-07-30
feat(llm): expose agent deploy tools via vscode.lm and Positron commands

### LINKED ISSUE #4305: Deploy to Connect from an AI agent (Language Model Tools + Positron agent commands)
Add tooling so an AI agent (Positron Assistant, GitHub Copilot, etc.) can inspect a project, confirm a plan, and deploy it to Posit Connect / Connect Cloud end-to-end.

- `planDeployment` (read), `deployContent` (write + deploy), `addCredential` (opens the credential UI)
- Secrets never enter tool arguments or model context
- Exposed **both** as `vscode.lm` Language Model Tools (tagged `positron-assistant`) and as Positron `positron.ai` agent commands (`contributes.commands[].agent`), for VSCode + Positron compatibility

mconflitti-pbc:
## Summary

Adds tooling so an AI agent (Positron Assistant, GitHub Copilot, and other tool-capable chat clients) can inspect a project, confirm a deployment plan with the user, and deploy it to Posit Connect / Connect Cloud end-to-end — secrets never enter tool arguments or model context.

- **`planDeployment`** (read-only) — inspects a project directory: content-type/entrypoint candidates, interpreter defaults, existing configs/records, and available credentials (redacted — name/url/serverType only).
- **`deployContent`** (write + deploy) — creates a config + deployment record if none exist, runs the real deploy, and returns a structured status (`success | needs-credential | needs-content-type | failed`).
- **`addCredential`** (initiate-only) — opens the interactive New Credential UI so the user completes authentication themselves; never asked for or passed a secret. Takes optional `serverUrl`/`target`/`authMethod` hints:
  - A `serverUrl` alone implies `target: "connect"` (Connect Cloud never has a URL), skipping the platform picker.
  - When target is confidently `"connect"` and a URL is known, browser sign-in starts automatically instead of showing the auth-method picker (unless the caller explicitly asked for `"apiKey"`).
  - A caller-supplied `serverUrl` is validated in the background (format, dedup, live connectivity — same checks as the interactive prompt) and, on success, the URL confirmation prompt is skipped entirely instead of asking the user to retype something they already gave. Falls back to the normal (still pre-filled) prompt if validation fails. Snowflake routing and the "no auto-signin hint for Snowflake" rule are unaffected either way.
  - Returns `{status: "added", credentialName}` or `{status: "canceled"}` instead of firing-and-forgetting, so `deployContent`'s `needs-credential` flow can wait for the result and retry automatically on `"added"`.

Each tool is exposed **two ways** for host compatibility:

1. As `vscode.lm` `LanguageModelTool`s (tagged `positron-assistant`) — works today in any host that forwards these. Tool descriptions carry the full sequencing/status/argument context needed to use them correctly on their own, since this path can't assume a skill is present to supply that context.
2. As agent-compatible VSCode commands (`posit.publisher.agent.*`) declared via `contributes.commands[].agent`, landing in Positron's `positron.ai` allow-list. Both upstream blockers for this path have since merged: [posit-dev/positron#15077](https://github.com/posit-dev/positron/pull/15077), [posit-dev/assistant#1810](https://github.com/posit-dev/assistant/pull/1810). Positron invokes these commands with a **single object argument** keyed by the declared `agent.args` names (confirmed empirically from a real tool call) rather than positional arguments — the handlers were fixed to match.

**Enabling refactor:** `runDeployWithProgress` and `HomeViewProvider.initiateDeployment` now return a structured `DeployOutcome` (`success | failed | canceled`) instead of `void`, so the deploy tool can report real status back to the agent. A new `deployProject()` selects the target in the sidebar before deploying so the user watches progress on it.

Tracking issue: #4305

## Review fixes

Addressed feedback from the initial review round (@dotNomad, @mconflitti-pbc) and a follow-up critical review:

- **Workspace escape**: `planDeployment`/`deployContent` now reject directories that resolve outside the workspace (`..`, absolute paths) before touching disk, via a shared `resolveWithinWorkspace()` helper (also used by `initiateDeployment`, replacing its duplicate inline check).
- **Connect Cloud `productType`**: `deployContent` now stamps new configs with `productType` derived from the credential's target, and rejects reusing an existing config whose `productType` doesn't match.
- **`enableConnectCloud` enforcement**: previously only checked at credential creation. `planDeployment` now excludes Connect Cloud credentials from its listing while the setting is off; `deployContent` rejects a Cloud credential before writing any files.
- **URL-prompt auto-skip**: added a `trustServerUrl` flag so only `addCredential` (agent-initiated) skips the URL-confirmation prompt. Human entry points (the "+" button, "Add credential for this deployment") still show it even when passing a pre-fill hint.
- **Cancellation**: `deployContent`'s `CancellationToken` is now threaded through `deployProject` → `initiateDeployment` → `runDeployWithProgress`, aborting the deploy the same way dismissing the progress notification does.
- **OAuth retry outcome**: fixed `runDeployWithProgress` always returning the initial failure even when `onError` performed a successful re-auth retry.
- **`addCredential` "unavailable" status**: fixed the disabled-platform check being thrown from inside a multi-step "step" callback, where the framework's generic error handler swallowed it and reported a generic "canceled" instead of the real reason (found via manual testing with Connect Cloud disabled).
- Documented the new `"unavailable"` status in both the `vscode.lm` tool description and the agent command's `returns` doc in `package.json`.

## Test plan

- [x] `tsc --noEmit` clean
- [x] ESLint clean on all changed source
- [x] Unit tests: `planDeploymentTool`, `deployContentTool`, `addCredentialTool`, `newCredential`, `newConnectCredential` (including the auto-proceed and Snowflake-routing cases), `deployProgress`, `homeViewSelectDeployment`, `workspaces` — all passing
- [x] `just test-extension-contracts` — 86/86 passing (registration + single-object-argument → `run()` mapping for both exposure paths)
- [x] Rebased onto latest `main`
- [x] Manual end-to-end verification in Positron (plan → confirm → `needs-credential` → `addCredential` with a supplied `serverUrl` → auto-proceed past the URL prompt → browser sign-in → deploy retries automatically)
- [x] Manual verification: disabling `positPublisher.enableConnectCloud` and asking the agent to deploy to Connect Cloud now surfaces a clear "Connect Cloud is disabled" reason instead of a misleading "authorization was canceled"

## Latest review fixes

- **Runtime input validation**: deployContent now rejects missing directory, entrypoint, or credentialName before touching the workspace, and the command path no longer relies on an unsafe DeployContentInput assertion.
- **Canonical project paths**: normalized directories are used consistently for inspection, config and record lookup/creation, and deployment, preventing ./app versus app mismatches.
- **Entrypoint safety**: deployment requires an exact match to a detected entrypoint and reports available candidates instead of selecting an arbitrary inspection result.
- **Structured failures**: TOML, inspection, file-write, and deployment exceptions are converted into meaningful failed tool results.
- **Type safety**: removed the progress return assertion and tightened the command argument normalization boundary.
- Added regression tests covering all of the above.

## Latest verification

- Targeted tooling tests: 64 passing.
- Extension contract tests: 86 passing.
- TypeScript, ESLint, and just package: passing.
- Full unit suite: 1,952 passing, 11 skipped, with 16 environment-dependent Quarto/R/renv/network integration failures.


mconflitti-pbc:
Demo in VSCode with copilot:

https://github.com/user-attachments/assets/0226932f-ca20-447f-a0a1-cd3386154a9b

likely could be improved if there was a skill providing a bit more guidance. Can test in positron once this gets into the build: https://github.com/posit-dev/positron/pull/15077 (have not tried building this locally yet)

nealrichardson:
> Can test in positron once this gets into the build: [posit-dev/positron#15077](https://github.com/posit-dev/positron/pull/15077) (have not tried building this locally yet)

If you add `"update.positron.channel": "dailies"` to your positron settings.json, you can opt into the daily builds, which should have that PR in it by now.

mconflitti-pbc:
@samclark2015 when you have a chance, can you take a look to confirm I am doing this properly given the work you did here? https://github.com/posit-dev/positron/pull/15077 

I am on positron daily release channel and just updated but not seeing tools loading in from the publisher extension.

samclark2015:
Hey, Matt! The Positron PR got merged in yesterday, but last night's Positron daily build failed to publish. I'm troubleshooting that right now, and I'll ping you once it's up and ready.

mconflitti-pbc:
> Hey, Matt! The Positron PR got merged in yesterday, but last night's Positron daily build failed to publish. I'm troubleshooting that right now, and I'll ping you once it's up and ready.

That explains it! Thanks

juliasilge (commented):
Hello! I am here with a drive-by review for visibility for folks. 😅 

Assistant is _removing_ the list of all of these commands (the full allow list) from the system prompt. This means that while it would be possible for Assistant to successfully call these commands, they won’t be anywhere in the prompting at all; effectively Assistant won’t know about them. We are working with the Core AI team on a way for Positron itself to tell Assistant about commands it can run at appropriate times. This highlights that we also need a way for extensions to do the same.

cc just for awareness as you discuss options @dhruvisompura and @jcheng5

mconflitti-pbc:
> Hello! I am here with a drive-by review for visibility for folks. 😅
> 
> Assistant is _removing_ the list of all of these commands (the full allow list) from the system prompt. This means that while it would be possible for Assistant to successfully call these commands, they won’t be anywhere in the prompting at all; effectively Assistant won’t know about them. We are working with the Core AI team on a way for Positron itself to tell Assistant about commands it can run at appropriate times. This highlights that we also need a way for extensions to do the same.
> 
> cc just for awareness as you discuss options @dhruvisompura and @jcheng5

@juliasilge Thanks for the heads up, Julia! We are planning to include that necessary tool context in the "deploy to connect" skill. As long as the tools get registered and can be called then I think this should be sufficient. We will test that out, but wanted to confirm this approach seems reasonable given the changes you noted above.

juliasilge:
The commands still can be called successfully, yes. The changes we're making are around how they get into the conversation context.

dhruvisompura:
> > Hello! I am here with a drive-by review for visibility for folks. 😅
> 
> > 
> 
> > Assistant is _removing_ the list of all of these commands (the full allow list) from the system prompt. This means that while it would be possible for Assistant to successfully call these commands, they won’t be anywhere in the prompting at all; effectively Assistant won’t know about them. We are working with the Core AI team on a way for Positron itself to tell Assistant about commands it can run at appropriate times. This highlights that we also need a way for extensions to do the same.
> 
> > 
> 
> > cc just for awareness as you discuss options @dhruvisompura and @jcheng5
> 
> 
> 
> @juliasilge Thanks for the heads up, Julia! We are planning to include that necessary tool context in the "deploy to connect" skill. As long as the tools get registered and can be called then I think this should be sufficient. We will test that out, but wanted to confirm this approach seems reasonable given the changes you noted above. 

This seems reasonable to me given that we're planning to do something similar in Positron. The tool will be callable with any command. We're not changing the registration of the tool.

mconflitti-pbc:
Screenshot 2026-08-05 at 4 56 23 PM

test/extension-contract-tests/src/globalSetup.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2026 by Posit Software, PBC.
+
+import { rm } from "fs/promises";
```
dotNomad, line 3: Nit: Since there isn't a `tsconfig` configuration file setup here I'm getting warnings about `@types/node` locally. Not a huge deal, but setting up a `tsconfig.json` with `"types": ["node", "vitest/globals"]` would be handy.

Something like this:

```
{
  "compilerOptions": {
    "target": "ES2023",
    "module": "preserve",
    "moduleResolution": "bundler",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "noEmit": true,
    "lib": ["ES2023", "DOM"],
    "types": ["node", "vitest/globals"],
    // Mirrors the `src` alias in vitest.config.ts. `vscode`/`positron` are
    // deliberately not mapped to their mocks: the mocks are partial runtime
    // stubs, so the extension source reached through this alias needs to type
    // check against the real `@types/vscode` API, not against the stub.
    "paths": {
      "src/*": ["../../extensions/vscode/src/*"]
    }
  },
  "include": [
    "vitest.config.ts",
    "src/**/*.ts",
    // Declares the ambient `positron` module that src/utils/vscode.ts imports.
    "../../extensions/vscode/src/@types/positron.d.ts"
  ],
  "exclude": ["src/mocks/*.conformance.ts"]
}
```

test/extension-contract-tests/src/globalSetup.ts
```diff
@@ -0,0 +1,21 @@
...
+const snowflakeLog = path.join(process.cwd(), "snowflake.log");
+
+async function removeSnowflakeLog() {
+  await rm(snowflakeLog, { force: true });
+}
+
+export default async function setup() {
+  await removeSnowflakeLog();
+  return removeSnowflakeLog;
+}
```
dotNomad, line 24: Nit: Reading the `globalSetup` docs for Vitest I think the `exports` method (as opposed to the default) is a bit clearer. 

```
export async function setup() {
  await removeSnowflakeLog();
}

export async function teardown() {
  await removeSnowflakeLog();
}
```

extensions/vscode/package.json
```diff
@@ -79,6 +79,135 @@
       }
     ],
     "commands": [
+      {
+        "command": "posit.publisher.agent.planDeployment",
+        "title": "Plan a Deployment",
+        "category": "Posit Publisher",
+        "agent": {
+          "description": "Inspect a project directory to determine what can be deployed to Posit Connect or Connect Cloud. Returns detected content type and entrypoint candidates, interpreter defaults, any existing deployment configurations and records, and the list of available server credentials (names and URLs only, never secrets). Use this before deploying to build a plan to confirm with the user.",
```
dotNomad, line 87: I want to make sure I'm understanding this right - this is a Positron specific contribution. I'm not seeing anything in the VSCode contributes for commands like this: https://code.visualstudio.com/api/references/contribution-points#contributes.commands

extensions/vscode/package.json
```diff
@@ -79,6 +79,135 @@
...
+          ],
+          "returns": "A JSON object with the inspected projectDir, interpreter defaults, deployable candidates, existing configurations and deployments, and redacted credentials."
+        }
+      },
+      {
+        "command": "posit.publisher.agent.deployContent",
+        "title": "Deploy Content to Connect",
+        "category": "Posit Publisher",
+        "agent": {
+          "description": "Deploy a project directory to Posit Connect or Connect Cloud. Creates a deployment configuration and record if they do not exist, then publishes and returns the deployed content URLs. Requires the name of an existing stored credential, never a secret directly. If no suitable credential exists it returns status 'needs-credential' — call the addCredential command (passing target/serverUrl if known) and wait for its result, then call this command again if a credential was added. Always confirm the deployment plan (from planDeployment) with the user before calling this command.",
```
dotNomad, line 106: Question: There is a setting to "Enable Connect Cloud" that can be disabled (`positPublisher.enableConnectCloud`) I wonder if we need to craft these a bit to that setting. Perhaps the description is totally fine but the tooling needs to distinguish. 🤔

extensions/vscode/package.json
```diff
@@ -322,6 +451,18 @@
...
+          "when": "false"
+        },
+        {
+          "command": "posit.publisher.agent.deployContent",
+          "when": "false"
+        },
+        {
+          "command": "posit.publisher.agent.addCredential",
+          "when": "false"
+        },
```
dotNomad, line 465: Good catch here. I was thinking to myself that I needed to check if these were disabled in my review.

extensions/vscode/package.json
```diff
@@ -646,6 +787,117 @@
           "type": "object",
           "properties": {}
         }
+      },
+      {
+        "name": "publish-content_planDeployment",
```
dotNomad, line 792: Ah here are the [`languageModelTools`](https://code.visualstudio.com/api/references/contribution-points#contributes.languageModelTools)

test/extension-contract-tests/src/contracts/llm-tools.test.ts
```diff
@@ -18,38 +26,185 @@ vi.mock("src/llm/tooling/troubleshoot/configurationTroubleshootTool", () => ({
...
+    commands.registerCommand as unknown as { mock: { calls: any[][] } }
+  ).mock.calls.find((c) => c[0] === id);
+  if (!call) {
+    throw new Error(`No command registered with id "${id}"`);
+  }
+  return call[1];
+}
+
+function register() {
+  registerLLMTooling(makeContext() as any, {} as any, {} as any);
```
dotNomad: I typically try to avoid `as` where I can since it defeats the purpose of TypeScript, there are a few in this file (and it is a testing file), but this one is particularly rough. Is there anything we can do to avoid so many type assertions?

extensions/vscode/src/views/deployProgress.ts
```diff
@@ -363,11 +378,19 @@ export function runDeployWithProgress(options: DeployProgressOptions): void {
...
           }
         }
+
+        return {
+          status: "failed",
+          message,
+          errCode: lastErrCode,
+          logsUrl: lastLogsUrl,
+          dashboardUrl: lastDashboardUrl,
+        } as const;
```
dotNomad, line 433: I'm seeing that `onError(err)` above calls `handleDeployError` for Connect. Looking at that codepath it looks like when `handleDeployError` prompts the user for a retry, refreshes the credentials, then initiates a deployment this reports that as a failure.

extensions/vscode/src/views/homeView.ts
```diff
@@ -1850,20 +1921,33 @@ export class HomeViewProvider implements WebviewViewProvider, Disposable {
...
+        undefined,
+        startingServerType,
+        authMethodHint,
       );
-      if (credential) {
-        this.refreshCredentials();
+      if (!credential) {
+        return { status: "canceled" };
       }
+      this.refreshCredentials();
```
dotNomad: This should be `awaited` right?

`deployContent` is supposed to be called right after the credential is added and we are reporting the added status before it is made, potentially.

extensions/vscode/src/multiStepInputs/newCredential.ts
```diff
@@ -98,19 +103,31 @@ export async function newCredential(
...
     const platformList = getPlatformList();
 
-    // If only one platform is available (Connect Cloud disabled), skip the picker
-    // and go directly to Connect credential flow
-    if (platformList.length === 1) {
-      serverType = ServerType.CONNECT;
+    // Skip the picker when there's only one platform available (Connect
+    // Cloud disabled), or the caller already knows the target.
+    const resolvedServerType =
+      startingServerType ??
```
dotNomad, line 144: Looks like the `startingServerType` can bypass the `getPlatformList()` which knows whether Connect Cloud is enabled or not through `positPublisher.enableConnectCloud`.

Should we add protection here or perhaps in the `addCredentialTool` to avoid that? It looks like it would be better here, but that may confuse the model.

A user who has turned off Connect Cloud could get dropped into a Connect Cloud auth flow based on the model's interpretation right now.

dotNomad (commented):
I still have about half of this to go through, but I had a couple findings already that I think we need to address.

extensions/vscode/package.json
mconflitti-pbc: Yeah i think the tool can be aware it can deploy to either but when the actions are taken it should definitely respect the settings and not try to go down a dead path.

test/extension-contract-tests/src/contracts/llm-tools.test.ts
mconflitti-pbc: let me take another pass at this

extensions/vscode/src/views/homeView.ts
mconflitti-pbc: good catch! yes i believe you are right

extensions/vscode/src/views/deployProgress.ts
mconflitti-pbc: hm let me look closer, but it should probably still let the agent know the initial deploy failed and why even if a new one is initiated as a result.

extensions/vscode/src/multiStepInputs/newCredential.ts
mconflitti-pbc: yeah i wasn't aware you could turn off connect cloud originally so this definitely needs more work. I think in this part of the code we can make sure we always call platformlist first and ensure what the agent tries to select is in that list otherwise error out

mconflitti-pbc:
Screenshot 2026-08-06 at 11 46 44 AM

Agent tool now respects and bubbles up to user a message about the connect cloud setting

extensions/vscode/src/workspaces.ts
```diff
@@ -1,7 +1,35 @@
...
+ * before reading or writing any file derived from the directory.
+ */
+export function resolveWithinWorkspace(
+  root: string,
+  relDir: string,
+): ResolvedWorkspacePath {
+  const absPath = nodePath.resolve(root, relDir);
+  const rel = nodePath.relative(root, absPath);
+  // We can't just check rel.startsWith("..") because a directory literally
+  // named ".." would be a false positive.
```
dotNomad: Just a bit of comment clarity

```suggestion
  // We can't just check rel.startsWith("..") because a directory starting
  // with "..", like "..config", would be a false positive.
```

extensions/vscode/src/workspaces.ts
```diff
@@ -1,7 +1,35 @@
...
+export type ResolvedWorkspacePath =
+  { ok: true; absPath: string } | { ok: false; error: string };
+
+/**
+ * Resolve a (possibly untrusted, e.g. agent-supplied) relative directory
+ * against the workspace root, rejecting any path that escapes the
+ * workspace via ".." segments or an absolute path. Callers must run this
+ * before reading or writing any file derived from the directory.
+ */
+export function resolveWithinWorkspace(
```
dotNomad, line 19: Good idea pulling this out.

extensions/vscode/src/views/deployProgress.ts
```diff
@@ -363,11 +423,20 @@ export function runDeployWithProgress(options: DeployProgressOptions): void {
...
       } finally {
+        externalCancelListener?.dispose();
         onComplete();
       }
     },
-  );
+  ) as Promise<DeployOutcome>;
```
dotNomad: We should get rid of this to avoid missing TypeScript compilation checks.

The easiest way to do this is:
- `export async function runDeployWithProgress(`
- `return await window.withProgress`
- and say that `async (progress, token): Promise<DeployOutcome>` returns the `DeployOutcome` promise.

That gets us the correct type annotations without the `as` without sacrificing type checking.

Additionally that means we can remove all of the `as const` assertions on the returns, but that is more of a nit, not a requirement.

extensions/vscode/src/llm/index.ts
```diff
@@ -18,5 +53,48 @@ export function registerLLMTooling(
...
+        deployTool.run(
+          normalizeAgentCommandArgs<DeployContentInput>(raw, [
+            "directory",
+            "entrypoint",
+            "credentialName",
+            "title",
+            "contentType",
+            "deploymentName",
+            "configurationName",
+          ]) as DeployContentInput,
```
dotNomad: Can we validate that we have the the `DeployContentInput` values here? We know `normalizeAgentCommandArgs` gives back a `Partial`, but it could be helpful to throw an error here before running the tool since the missing fields can be pretty gnarly

- a missing entrypoint will deploy the first inspection target
- a missing credential name will get "No stored credential named 'undefined'"
- a missing directory becomes the workspace root which could be rough for nested projects

extensions/vscode/src/llm/agentCommandArgs.ts
```diff
@@ -0,0 +1,30 @@
...
+ * as if it were the object form. Normalize either shape into the object
+ * `run()` expects.
+ */
+export function normalizeAgentCommandArgs<T extends object>(
+  raw: unknown[],
+  argNames: readonly string[],
+): Partial<T> {
+  const first = raw[0];
+  if (typeof first === "object" && first !== null && !Array.isArray(first)) {
+    return first as Partial<T>;
```
dotNomad: This actually is a `Partial<T>` so no reason for the `as`!

```suggestion
    return first;
```

extensions/vscode/src/llm/agentCommandArgs.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  const input: Record<string, unknown> = {};
+  argNames.forEach((name, index) => {
+    if (raw[index] !== undefined) {
+      input[name] = raw[index];
+    }
+  });
+  return input as Partial<T>;
```
dotNomad, line 29: This one is unavoidable with `raw`, but the fact that it is a `Partial` helps since we could get nothing back.

dotNomad (commented):
Caught a few more things, mostly around TypeScript `as` - now on to the tools specifically.

extensions/vscode/src/llm/index.ts
dotNomad: Oh I'm realizing now that this only covers the Positron command path. We could do checking in the actual running of the tool 🤔

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
```diff
@@ -0,0 +1,256 @@
...
+        error: `Credential "${credential.name}" targets Connect Cloud, which is disabled (positPublisher.enableConnectCloud). Ask the user to re-enable it, or deploy with a Connect credential instead.`,
+      };
+    }
+
+    const expectedProductType = getProductType(credential.serverType);
+
+    // Resolve the configuration: reuse an existing one, or create a new one.
+    let configName = input.configurationName;
+    if (configName) {
+      const existingConfig = await loadConfiguration(configName, relDir, root);
```
dotNomad: This probably needs a `try` `catch` since `loadConfigFromFile` can `throw` if there is a `TomlError`

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
```diff
@@ -0,0 +1,256 @@
...
+    private readonly state: LLMToolingState,
+    private readonly homeViewProvider: LLMToolingHomeView,
+    private readonly clientVersion: string,
+  ) {}
+
+  async invoke(
+    options: LanguageModelToolInvocationOptions<DeployContentInput>,
+    token: CancellationToken,
+  ): Promise<LanguageModelToolResult> {
+    const result = await this.run(options.input, token);
```
dotNomad: This probably needs a `try` `catch` too since there are a few calls that could throw.

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
```diff
@@ -0,0 +1,256 @@
...
+            : undefined,
+        clientVersion: this.clientVersion,
+      });
+    }
+
+    const outcome = await this.homeViewProvider.deployProject(
+      deploymentName,
+      input.credentialName,
+      configName,
+      relDir,
```
dotNomad: This is still the raw `input.directory` right? We should normalize it so `findContentRecord` doesn't miss anything and accidentally deploy a second piece of content.

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
```diff
@@ -0,0 +1,256 @@
...
+    if (!configName) {
+      const inspections = await inspectProject({
+        projectDir: absDir,
+        entrypoint: input.entrypoint,
+        relativeDir: relDir,
+      });
+      const match =
+        inspections.find(
+          (i) => i.configuration.entrypoint === input.entrypoint,
+        ) ?? inspections[0];
```
dotNomad: Raising that `?? inspections[0]` will select an unknown type and no entrypoint which can write a config with an empty entrypoint. Could be worth raising rather than defaulting.

dotNomad (commented):
This is extremely close! The `addCredentialTool` and `planDeploymentTool` looked great from my review. I think there are a few things we can harden in the `deployContentTool` though.

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
mconflitti-pbc: yeah that is fair!

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
mconflitti-pbc: good call, though if there is an error, the agent would just reply that something failed I think so nothing super catastrophic. PRobably better to catch and provide a meaningful error message!

dotNomad (approved):
Awesome adjustments. I think we got this to a pretty great place - hardened TypeScript and good tool error behavior.

Thank you for going back and forth on this a bit. 🎉

PR #4335, opened 2026-08-10
fix(llm): reload credentials before deployment

mconflitti-pbc:
### Summary
- Reload Publisher credentials before `deployContent` reports `needs-credential`.
- Prevent a stale in-memory cache from disagreeing with `planDeployment` during extension startup.
- Add regression coverage for the cache-miss/reload path.

### Verification
- `git diff --check`
- Package JSON validation
- Focused Vitest test attempted, but dependencies are not installed in this checkout (`vitest: command not found`).

@joshyam-k @dotNomad please review.

### DIFF
--- a/extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
+++ b/extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
@@ -157,7 +157,15 @@ export class DeployContentTool implements LanguageModelTool<DeployContentInput> 
       const absDir = resolvedDir.absPath;
       const relDir = relativeProjectDir(absDir, root);
 
-      const credential = this.state.findCredential(credentialName);
+      let credential = this.state.findCredential(credentialName);
+      if (!credential) {
+        // planDeployment reads SecretStorage directly, while PublisherState
+        // may still be loading its in-memory cache during extension startup.
+        // Refresh before reporting that a credential is missing so the two
+        // agent tools cannot disagree about an existing stored credential.
+        await this.state.refreshCredentials();
+        credential = this.state.findCredential(credentialName);
+      }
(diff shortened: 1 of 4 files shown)


extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
```diff
@@ -157,7 +157,15 @@ export class DeployContentTool implements LanguageModelTool<DeployContentInput>
       const absDir = resolvedDir.absPath;
       const relDir = relativeProjectDir(absDir, root);
 
-      const credential = this.state.findCredential(credentialName);
+      let credential = this.state.findCredential(credentialName);
```
joshyam-k, line 160: can we just leverage `this.state.credentialsService.list();` from the block below and search through it? something like

```ts
let credential = this.state.findCredential(credentialName);
if (!credential) {
  const all = await this.state.credentialsService.list();
  credential = all.find((c) => c.name === credentialName);
  if (!credential) {
    return {
      status: "needs-credential",
      ...
    };
  }
}
```

I guess this is mostly desirable if we want this to happen more silently than it would if we run `refreshCredentials`

joshyam-k (approved):
one restructure suggestion but also fine as is as long as we resolve the lint issue

extensions/vscode/src/llm/tooling/deploy/deployContentTool.ts
mconflitti-pbc: there may be something here that we can address on a follow up. the refresh only happens on a cache miss and will do the actual token refresh or prompt sign in if deployment fails. i think this is a good tradeoff for now though open to changing it in the future.

PR #4298, opened 2026-07-21
Add Positron API integration tests and CI workflow

jonvanausdeln:
## Summary

Closes https://github.com/posit-dev/positron/issues/14963

Adds a Positron-only integration test suite that runs the Publisher extension inside a **real Positron build** via [`@posit-dev/positron-test-electron`](https://github.com/posit-dev/positron-test-electron), plus a CI workflow that runs it on every PR and push to `main`.

Part of the rollout tracked in posit-dev/positron#14531 (pattern established in quarto-dev/quarto#1058).

## What's covered

Publisher's real Positron-API surface is interpreter discovery (`src/utils/vscode.ts`): `acquirePositronApi()` feature detection plus `positron.runtime.getPreferredRuntime()` with VSCode fallbacks. The starter tests exercise exactly that against the live API — paths unreachable by the plain VSCode suite or the mock-based contract tests:

- **`extension.test.ts`** — Positron injects the `acquirePositronApi` global, the API reports a version, and Publisher activates in the Positron extension host.
- **`interpreter-discovery.test.ts`** — Publisher's `getPythonInterpreterPath()` / `getRInterpreterPath()` resolve the same interpreter Positron reports as its preferred runtime, covering API acquisition, the retry loop, and `~/` runtime-path expansion (posit-dev/positron#12942). The tests wait for Positron's async runtime discovery before asserting, so they measure Publisher's behavior rather than discovery timing.

## Harness

- `scripts/run-positron-tests.mjs` — downloads/caches a Positron build (stable by default, `POSITRON_CHANNEL=daily` supported) and runs the Mocha entry point (`src/test/positron/index.ts`) inside it. Bundled extensions stay enabled because runtime discovery is provided by the bundled Python and Ark extensions.
- Run locally with `npm run test-positron` (**macOS only** for now — an upstream `positron-test-electron` limitation; Windows/Linux follow once extractable archives are published).
- `esbuild.tests.mjs` / `.vscode-test.mjs` — the plain suite (`out/test/suite/`) and Positron suite (`out/test/positron/`) are now separated so `npm test` doesn't pick up Positron-only tests.
- `src/test/positron/README.md` documents how to run and add tests.

## CI

`.github/workflows/positron-api-tests.yaml` (macos-latest, Python 3.12 + R 4.4 for runtime discovery, date-keyed cache of the Positron download, launched through [`posit-dev/setup-positron`](https://github.com/posit-dev/setup-positron)), wired into `pull-request.yaml` and `main.yaml`.

[...]

PR #4346, opened 2026-08-18
ci(positron): run Positron API tests on ubuntu-latest

jonvanausdeln:
Moves the `Positron-API-Tests` workflow from `macos-latest` to `ubuntu-latest`. Linux Actions minutes bill at roughly 1/10 the macOS rate.

Part of posit-dev/positron#15281. The suite itself landed in #4298.

## Changes

Beyond the runner line, four things were required:

1. **`xvfb-run -a` wrap.** Positron is an Electron app and needs a display server on Linux. `posit-dev/setup-positron` is a thin composite step that only runs the command it is handed, so it provides none. Without this the run fails with `Missing X server or $DISPLAY` and then dies with SIGSEGV.
2. **`@posit-dev/positron-test-electron` → `^0.0.3`.** 0.0.2's platform table is `darwin-arm64`/`darwin-x64` only; `linux-x64` and `win32-x64` descriptors land in 0.0.3 (the latest published version). The bump is in `extensions/vscode/package.json` with the root workspace `package-lock.json` updated to match.
3. **`posit-dev/setup-positron@v1`** instead of `@main`, so upstream changes to the action can't break CI here.
4. **`${{ runner.os }}` in the Positron download cache key**, on both the restore and the save step. The cache layout is `.positron-test/<channel>/<version>/<platform>`, so under an OS-less key a stale same-day macOS entry would restore, set `cache-hit=true`, suppress the save step, and the Linux build would never get cached that day.

Also:

- Added `workflow_dispatch` alongside `workflow_call` so the suite can be triggered manually (`e2e.yaml` already does both). This is what made the pre-merge verification below possible.
- Replaced the "macOS only" notes in `scripts/run-positron-tests.mjs`, `src/test/positron/README.md`, and `extensions/vscode/CLAUDE.md` with the `xvfb-run` recipe for headless Linux runs. Unrelated `darwin` platform switches in extension source are untouched.

## Verification

Dispatched on a fork before opening this PR ([run](https://github.com/jonvanausdeln/publisher/actions/runs/32176545448), commit `4b9ac8d8`). A green job alone doesn't prove the tests ran, so from the log:

```
Run xvfb-run -a npm run test-positron
  ✔ Positron injects the acquirePositronApi global
  ✔ Publisher activates in Positron
  ✔ resolves the Python interpreter from Positron's preferred runtime (2051ms)
  ✔ resolves the R interpreter from Positron's preferred runtime
  4 passing (3s)
Extension host test runner exit code: 0
Exit code:   0
```

Both interpreter-discovery tests pass, so `actions/setup-python@v7` and `r-lib/actions/setup-r@v2` resolve on ubuntu and Positron's runtime discovery finds both. `positron-channel: stable` works on Linux — the 0.0.3 README's "stable not available on Linux/Windows yet" caveat is stale. Wall clock was ~5 minutes.

No CHANGELOG entry: CI-only, no user-facing behavior change.

### DIFF
--- a/.github/workflows/positron-api-tests.yaml
+++ b/.github/workflows/positron-api-tests.yaml
@@ -1,13 +1,14 @@
 name: Positron-API-Tests
-on: [workflow_call]
+on:
+  workflow_call:
+  workflow_dispatch:
 permissions:
   contents: read
 env:
   POSITRON_CHANNEL: stable
 jobs:
   test:
-    # @posit-dev/positron-test-electron currently supports macOS only.
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
       - uses: actions/setup-node@v7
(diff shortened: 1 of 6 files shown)


dotNomad (approved):
This is looking great!

PR #4348, opened 2026-08-19
Fix OAuth browser sign-in fallback compatibility

### LINKED ISSUE #4347: Fall back to legacy token auth when OAuth device flow is unavailable
## Summary

Publisher browser sign-in fails against Connect versions that advertise OAuth but do not support RFC 8628 device authorization. Instead of falling back to the existing browser-based token claim flow, Publisher reports:

> OAuth sign-in did not complete: This Connect server does not advertise the OAuth device flow. You can use an API key instead.

This affected a workshop using Publisher v2.9.17. Upgrading Connect to a version with device-flow support made sign-in work, but the upgrade masked a Publisher backward-compatibility regression.

## What happened

1. Publisher discovers `/.well-known/oauth-authorization-server`, available on Connect 2026.02.0 and later.
2. Any successful OAuth discovery causes **Sign in with a browser** to select the OAuth path unconditionally.
3. In a web/remote Positron environment, Publisher cannot use the local loopback callback. If no usable Workbench redirect relay is detected, or that relay cannot be used, Publisher selects the device-code flow.
4. Connect 2026.02.0 through 2026.05.x advertise OAuth but do not advertise `device_authorization_endpoint`; device flow requires Connect 2026.06.0 or later.
5. `OAuthClient.startDeviceAuth()` throws because the endpoint is absent. The credential flow then returns to the auth picker with API key as the only functioning alternative.

The legacy token claim flow is currently selected only when OAuth discovery returns no metadata. It is not reconsidered when the available OAuth transports are incompatible with the current editor environment and Connect version.

## Expected behavior

Browser authentication should preserve this compatibility cascade:

1. OAuth authorization code using a viable callback transport
2. OAuth device flow, only when `device_authorization_endpoint` is advertised
3. Legacy browser token claim
4. Manual API key

A Connect server advertising OAuth should not by itself make the legacy token flow unreachable.

## Relevant code

- `extensions/vscode/src/multiStepInputs/newConnectCredential.ts`: `dispatchAuthMethod()` routes browser sign-in to OAuth whenever `oauthMetadata` is non-null.
- `extensions/vscode/src/auth/oauth/activator.ts`: `authorize()` falls back to `deviceCodeFlow()` for web/remote environments and after unsuccessful Workbench/loopback transports without first checking device-flow capability.
- `extensions/vscode/src/auth/oauth/client.ts`: `startDeviceAuth()` correctly rejects metadata without `device_authorization_endpoint`, producing the observed error.

## Test gap

Activator tests use OAuth metadata that always contains `device_authorization_endpoint`. Client tests separately confirm that an absent device endpoint throws, but there is no cross-layer test covering:

- OAuth metadata present
- `device_authorization_endpoint` absent
- web/remote environment, unavailable Workbench relay, rejected Workbench redirect URI, or failed loopback transport

Add coverage for that compatibility matrix and assert that the credential flow proceeds to legacy browser token authentication rather than API-key-only recovery.

## Version context

- Last known-good workshop: July 28, before the OAuth Publisher change shipped.
- OAuth implementation merged July 30 and shipped beginning with Publisher v2.9.14.
- Failure reproduced with Publisher v2.9.17.
- Upgrading Connect to a release supporting the device endpoint allowed the same Publisher flow to complete.

mconflitti-pbc:
## Summary

- preserve the browser authentication cascade when Connect advertises OAuth without every required capability
- fall back from unavailable Workbench, remote, loopback, or device transports to legacy browser token authentication
- classify Workbench relay polling connection failures as nonterminal while keeping denial and timeout terminal
- preserve expired OAuth credentials and provide actionable guidance when reauthentication cannot use OAuth
- add compatibility-matrix, negative-path, and order-independent test coverage

Fixes #4347

## Verification

- `npx vitest run src/auth/oauth src/multiStepInputs/newConnectCredential.test.ts` — 90 tests passed
- focused tests pass with shuffled execution (`--sequence.shuffle --sequence.seed=17`)
- `npx tsc --noEmit --project tsconfig.json`
- focused ESLint checks
- Prettier and `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Browser sign-in now falls back to the legacy token flow when Connect advertises OAuth without dynamic client registration, or when device authorization is selected or needed as a fallback but the server does not advertise it. This includes forced device-code mode. (#4347)
+
 ## [2.10.0]
 
 ### Added
--- a/extensions/vscode/src/auth/oauth/activator.test.ts
+++ b/extensions/vscode/src/auth/oauth/activator.test.ts
@@ -104,7 +104,10 @@ vi.mock("./workbench", async () => {
   };
 });
 
-import { ConnectOAuthActivator } from "./activator";
+import {
+  ConnectOAuthActivator,
+  OAuthTransportUnavailableError,
+} from "./activator";
 import { OAuthMetadata } from "./types";
 import { WorkbenchRelayError } from "./workbench";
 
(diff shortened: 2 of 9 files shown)


joshyam-k (approved):
This all makes sense to me!

Issue #3323, opened 2025-12-03
feat: Publish scripts as Quarto documents


brooklynbagel:
## Description

Connect supports rendering scripts (`.py` and `.R`) files as Quarto documents, see https://docs.posit.co/connect/user/scripts/. There's some front-matter that needs to be added that makes Quarto recognize them as renderable, see https://quarto.org/docs/computations/render-scripts.html#syntax and the scripts are renders using Jupyter or Knitr. However, the Publisher extension does not know how to handle these types of files. For example, if I try to publish the following R script

```R
#' ---
#' title: "R script file"
#' ---

# Randomly generate data that acts as 30-day trailing unit sales.
today <- Sys.Date()
data <- data.frame(
  Date = today - 1:30,
  Sales = sample(75:100, 30, replace = TRUE)
)

# Write a CSV for downstream use.
write.csv(data, "sales.csv", row.names = FALSE)

#' Download [CSV](sales.csv)
```

Publisher generates the following configuration

```toml
# Configuration file generated by Posit Publisher.
# Please review and modify as needed. See the documentation for more options:
# https://github.com/posit-dev/publisher/blob/main/docs/configuration.md
product_type = 'connect'
'$schema' = 'https://cdn.posit.co/publisher/schemas/posit-publishing-schema-v3.json'
type = 'unknown'
entrypoint = 'rscript.R'
validate = true
files = [
  '/rscript.R',
  '/renv.lock',
  '/.posit/publish/render-rscript-reprex-545N.toml',
  '/.posit/publish/deployments/deployment-OVT9.toml'
]
title = 'render-rscript-reprex'
```

Where the `type` config should be `quarto-static`, and an `[r]` and `[quarto]` block are missing. 

### Note

It does appear that Publisher knows how to handle `.py` handles rendered with Jupyter.

```py
# %% [markdown]
# ---
# title: "Python script file"
# ---

# %%
import datetime
import pandas as pd
import random

# Randomly generate data that acts as 30-day trailing unit sales.
today = datetime.date.today()
data = pd.DataFrame({
    'Date': pd.date_range(
        today - datetime.timedelta(days=30),
        today - datetime.timedelta(days=1),
    ),
    'Sales': [random.randint(75, 100) for _ in range(1, 31)],
})


# Write a CSV for downstream use.
data.to_csv("sales.csv", index=False)

# %% [markdown]
# Download [CSV](sales.csv)
```

Publisher generates the configuration

```toml
[...]

PR #4374, opened 2026-09-04
Prompt for content type when detection can't determine one

### LINKED ISSUE #2666: Bad experience: Suggesting the user to manually set the content type
## Description

When trying to publish content for which Publisher fails to get the content type, the following message shows up:
`"Please set the framework you are using, for example type = 'python-shiny'. Edit the Configuration." `

As if users knew what the valid values for this field are, and what the implications of picking one are.

### LINKED ISSUE #2833: app_mode: unknown is a bad experience
If you specify an entrypoint file and Publisher can't infer what "app mode" it should be on Connect, it generates `app-mode: unknown`. Technically you can edit the TOML to change this, but in practice it is not something mortals can do. 

My use case was deploying a Python app that uses a framework based on FastAPI, so we can support it. But it doesn't import FastAPI, so the detection does not work. 

We believe that having a published TOML schema, and using the TOML plugin, you can get hints to help create valid configuration. I tried to lean into that. So, I start trying to type over "unknown", and a red squiggly line appears under it and tells me what the valid values are. Cool. I never would have guessed that it should be "python-fastapi" otherwise. So I type that, and then... the whole TOML file gets red squigglies! On hover, now it says `"python" is a required property` So I start typing `python =` on a new line and the squigglies go away. I guess that this must be the python version, so I type "3.12", but now `python` has a squiggly under it that says `"3.12" is not of type "object"`. Ok, so... what is an object in TOML? I type `python = {}` and the squigglies go away. But surely that can't be right, I haven't specified anything!

The only reasonable workaround I found was to (after deleting all of `.posit`) put a fake import in my entrypoint file (`import fastapi` in this case), running the detection again to get the right `python-fastapi` and other bits in place, and then deleting the fake import. That's not great.

nealrichardson:
## Summary

When Publisher can't detect a content type for an entrypoint, it now prompts you to pick one from the list of valid content types instead of silently writing `type = "unknown"` to the config. The chosen type is treated as if it had been detected automatically, including filling in the required `[python]`/`[r]`/`[quarto]` sections.

Closes #2833
Closes #2666

## Test plan

- [x] `tsc --noEmit`
- [x] `eslint` (project-wide, zero warnings)
- [x] `vitest run` (full suite, 1991 tests)
- [x] Manual: create a deployment for an entrypoint Publisher can't detect (e.g. a FastAPI app with no `import fastapi`) and confirm the content-type picker appears and produces a valid config


nealrichardson:
I tried this locally, and it works, but the contents of the menu, the list of options of app modes, needs some work, so let me iterate on this some more.

Screenshot 2026-09-14 at 2 41 04 PM

It also doesn't know anything about scripts, which was one case that bit me on this feature.

Screenshot 2026-09-14 at 2 41 28 PM

nealrichardson:
This is ready for review now. 

https://github.com/user-attachments/assets/e338d41a-aa12-4313-b341-4c906ebd5047

It doesn't fully satisfy the script use case because it doesn't inject the frontmatter if missing, but I think we should tackle that separately. See #3323.

extensions/vscode/src/inspect/manualContentTypeRanking.ts
```diff
@@ -0,0 +1,107 @@
...
+  allValidContentTypes,
+  ContentType,
+} from "src/api/types/configurations";
+import { ScriptLanguage } from "./index";
+
+// Content types the manual picker suggests first for a given entrypoint
+// extension, ranked by likelihood, plus (for .R/.r and .py) the language to
+// use for the "Script" entry. Extensions not listed here get no "Suggested"
+// group; every type still appears under "All content types" regardless. See
+// PICKER-PLAN.md for the reasoning behind this mapping.
```
fizzboop: is the `PICKER-PLAN.md` a file that should have been added? i don't see it in the code base or in this PR.

extensions/vscode/src/api/types/configurations.ts
```diff
@@ -113,18 +113,63 @@ export const contentTypeStrings = {
...
   contentTypeLabelOverrides[type] ?? contentTypeStrings[type] ?? type;
 
+// Short descriptions shown alongside the label in the manual content-type
+// picker only (see getManualContentTypeQuickPicks). Distinct from
+// contentTypeStrings so the label and description are never the same string.
+// ContentType.QUARTO is intentionally omitted: it is a legacy alias for
+// ContentType.QUARTO_STATIC (both map to Connect's "quarto-static" app mode,
+// see bundler/appMode.ts) kept only so old configs still parse, and is never
+// offered as a choice.
+export const contentTypePickerDescriptions: Partial<
```
fizzboop, line 151: `QUARTO_STATIC` ("Quarto Document") is the only offered content type without a picker description. was that intentional?

Image

extensions/vscode/src/inspect/manualContentTypeRanking.ts
```diff
@@ -0,0 +1,107 @@
...
+  ".js": { types: [ContentType.NODEJS] },
+  ".ts": { types: [ContentType.NODEJS] },
+  ".md": { types: [ContentType.QUARTO_STATIC] },
+};
+
+// Content types the manual picker offers as choices. ContentType.QUARTO is
+// excluded: it is a legacy alias for ContentType.QUARTO_STATIC (both map to
+// Connect's "quarto-static" app mode, see bundler/appMode.ts) kept only so
+// old configs still parse, never as something to newly write.
+export const manualContentTypeChoices = allValidContentTypes.filter(
```
fizzboop, line 58: 🤖 mentioned the following:
>  The manual content-type picker unconditionally offers "Quarto Document" even for `.py/.R` script entrypoints. Selecting it writes a `[quarto]` config section with no engine and no `[python]` section, silently producing a config that fails to render on Connect

extensions/vscode/src/inspect/manualContentTypeRanking.ts
nealrichardson: Sigh, this is a Claude turd, I had it write a plan doc and then build it, there should be no references to it

extensions/vscode/src/api/types/configurations.ts
nealrichardson: Every description it suggested was wrong because quarto-static isn't actually static. I guess we should say something about how it is rendering on the server. That's the note maybe?

extensions/vscode/src/inspect/manualContentTypeRanking.ts
nealrichardson: huh ok I'll take a look, thanks for catching that!

extensions/vscode/src/api/types/configurations.ts
nealrichardson: Added that below, L167

extensions/vscode/src/inspect/manualContentTypeRanking.ts
nealrichardson: Claude added something for this but I'm not sure about it, there are generally lots of ways to select the wrong content type and that would cause failure to render on Connect, that's kinda why the app modes exist. This flow at least gives you a chance to get it right because the current "unknown" fallback *always* fails.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Data files read by code in a Quarto or R Markdown document are now included in the deployment. Publisher scans fenced code chunks (`{r}`, `{python}`, `{ojs}`, and others) and inline code for referenced files, so a report that calls something like `read_csv("data/penguins.csv")` no longer fails to render on the server the first time it is deployed. Only paths that resolve to a file inside the project directory are added. (#4388)
+- When Publisher can't automatically determine the content type for an entrypoint, it now prompts you to pick one from the list of valid content types instead of writing `type = "unknown"` to the configuration and leaving you to edit the TOML by hand. Picking a type fills in the same required fields (e.g. an empty `[python]` or `[r]` section) that automatic detection would have set. For a `.py` or `.R` entrypoint, the list now also offers a "Script" option that configures Connect to render it as a Quarto document, and ranks the choices most likely to match your entrypoint's file extension first. (#2833, #2666)
```
nealrichardson:
```suggestion
- When Publisher can't automatically determine the content type for an entrypoint, it prompts you to pick one from the list of valid content types. Picking a type fills in the same required fields (e.g. an empty `[python]` or `[r]` section) that automatic detection would have set. For a `.py` or `.R` entrypoint, the list now also offers a "Script" option that configures Connect to render it as a Quarto document, and ranks the choices most likely to match your entrypoint's file extension first. (#2833, #2666)
```

extensions/vscode/src/inspect/manualContentTypeRanking.ts
fizzboop: absolutely, it's definitely an improvement from how it was originally.

fizzboop (approved):

PR #4381, opened 2026-09-10
fix: Sync `server_settings` types/fixtures with new Connect fields

fizzboop:
## Intent
Connect recently [added](https://github.com/posit-dev/connect/pull/42900) a required status field and per-installation publishable field to the `server_settings/{python,r,quarto,nodejs}` API responses. Publisher's `connect-api` types and `mock-connect` test fixtures predated those additions, which broke the `connect-contract-tests/validate-fixtures` CI job on main and blocked unrelated dependabot PRs (#4368, #4369).

## Type of Change
- - [x] Bug Fix 
- - [ ] New Feature 
- - [ ] Breaking Change 
- - [ ] Documentation 
- - [ ] Refactor 
- - [ ] Tooling 

## Approach
Added `status/publishable` to the `connect-api` TypeScript types and updated the `mock-connect` fixtures and test mocks to match. No behavior changes beyond exposing the new fields on the typed client response.

## User Impact
None directly — this only fixes broken CI and stale internal types/fixtures. It does unblock future Publisher features that want to read runtime status/publishable.

## Automated Tests
Updated test/mock-connect fixtures and `connect-api/connectPublish` unit tests to cover the new required fields; ran `just validate-fixtures` and both affected test suites locally.

## Directions for Reviewers
Confirm `just validate-fixtures` passes and that the fixture values (`status`, `publishable`) look reasonable against the current Connect OpenAPI spec.

## Checklist
- [ ] I have updated the _root_ [CHANGELOG.md](../CHANGELOG.md) to cover notable changes. (N/A — internal test/type fix, no user-facing change)


packages/connect-api/src/types.ts
```diff
@@ -341,25 +341,37 @@ export interface SchedulerSettings {
...
 
+/** The configured, licensed, and discovered state of a runtime. */
+export interface RuntimeStatus {
+  enabled: boolean;
+  licensed: boolean;
+  available: boolean;
+  usable: boolean;
+}
+
 export interface PyInstallation {
```
hekhuisk, line 352: Not relevant to your changes, but is there is a reason we don't just have 
```
interface BaseInstallation {
  version: string;
  cluster_name: string;
  image_name: string;
}
```
for `PyInstallation`, `RInstallation`, `QuartoInstallation`, `NodejsInstallation` to extend?

extensions/vscode/src/publish/connectPublish.test.ts
```diff
@@ -363,10 +363,28 @@ function makeSettings(overrides?: Partial<AllSettings>): AllSettings {
...
-    nodejs: { installations: [], enabled: true, ...overrides?.nodejs },
+    python: {
+      installations: [],
+      api_enabled: true,
+      status: { enabled: true, licensed: true, available: true, usable: true },
+      ...overrides?.python,
+    },
+    r: {
+      installations: [],
+      status: { enabled: true, licensed: true, available: true, usable: true },
```
hekhuisk, line 374: nit: You could pull `installations` and `status` out into a separate const that is spread over each of these.

packages/connect-api/src/types.ts
fizzboop: i'm not entirely sure if there is. 
@dotNomad would that be possible or am i missing context on why we can't go that route?

dotNomad: I assume this is because the Connect API side isn't inheriting from a single type. Each of the installation `type` objects are their own with nothing enforcing that they share `version`, `cluster_name`, or `image_name`.

Not saying we can't do that here, but that is probably the reason the code is like that.

dotNomad (approved):
This looks like a very reasonable change to me. I'll leave it up to you @fizzboop if you want to create a `BaseInstallation` to reduce a bit of duplication, or if you want it to represent the Connect API extremely closely.

packages/connect-api/src/types.ts
fizzboop: ah okay, i'm leaning on leave it as is. i can probably add a [comment](https://github.com/posit-dev/publisher/pull/4381/commits/317df3fbc5ce6dd64b25dd43723839c2d6f47abe) just in case anyone else runs into this.

extensions/vscode/src/publish/connectPublish.test.ts
fizzboop: added in commit [1e9d0dd](https://github.com/posit-dev/publisher/pull/4381/commits/1e9d0dd3d78a931743fbe375817dd2fb9a0a701e)

hekhuisk (approved):

dotNomad (approved):
The test changes make sense to me. Thanks for pushing on that.

PR #4390, opened 2026-09-15
Scan code chunks for files on disk

### LINKED ISSUE #4388: resourceFinder doesn't scan code chunks in .qmd/.rmd, so data files read by code are missing from the bundle
## Problem

`resourceFinder` (added in #3149 for #2594) auto-adds referenced files to config files, but for `.qmd`/`.rmd` it only parses markdown content — image syntax, `` tags, `.html` links, and the YAML `resource_files:` key. Code chunks are never parsed.

So today:

- `` in a `.qmd` → found ✅
- `read_csv("data/penguins.csv")` in a `.qmd` code chunk → not found ❌

The result is that the first deploy of a Quarto report that reads a data file fails to render on the server, because the data file was never added to the bundle. Users then have to select files manually, or fall back to rsconnect, which has a straightforward argument for adding files to the bundle. This has been hitting users repeatedly in live use via the Posit Assistant publishing skill and the Positron publish command — the agent retries several times before working out that files need attaching, and often never does.

Notably, standalone `.r` files are handled the way we want: `scanR` (`extensions/vscode/src/inspect/helpers/resourceFinder.ts:234-257`) strips `#` comments, extracts all quoted string literals, and keeps the ones that resolve to an existing file — with no extension filter. That's essentially the behavior we want applied inside `.qmd`/`.rmd` code chunks.

## Ask

Extract fenced code chunks from `.qmd`/`.rmd` (and `.md`, where relevant) and run the same string-literal extraction plus path check over their contents, so that files a document's code references are added to the deployment. Re-implementing more of `rmarkdown::find_external_resources()` is fine if that's what it takes — the goal is that users get the files their code references included whenever it's reasonably possible.

## Pointers

- `extensions/vscode/src/inspect/helpers/resourceFinder.ts:8-15` — `supportedExtensions`
- `scanMarkdown` (`:130-194`) — splits YAML front matter from markdown inline; applies `markdownImgRe`, `markdownLinkRe`, `htmlImgInMarkdownRe` (`:22-26`). No chunk handling; there's no existing front-matter/markdown/chunk splitter helper.
- `scanR` (`:234-257`) + `processMatches`/`addResource` (`:331`) — the logic to reuse
- `parseYAMLResourceFiles` (`:380-431`)
- Callers that push results into `cfg.files`: `extensions/vscode/src/inspect/detectors/rmarkdown.ts:189-191` and the equivalent in `quarto.ts`
- Tests: `extensions/vscode/src/inspect/helpers/resourceFinder.test.ts`

## Suggested scope

- Handle fenced chunks of any engine (` ```{r} `, ` ```{python} `, ` ```{ojs} `, plus plain ` ```r `), and ideally inline code (`` `r ...` ``).
- Keep false positives low the same way `scanR` does: only add a path that resolves to a file that actually exists within the project directory.
- Don't add files already covered by config files / existing exclusions.
- Add unit tests covering: `read_csv("data/x.csv")` in an `{r}` chunk, `pd.read_csv("data/x.csv")` in a `{python}` chunk, a relative path that doesn't exist (must not be added), and a path escaping the project dir (must not be added).

Implementation should reuse `processMatches`/`addResource` rather than duplicating path handling, respect the project-directory boundary and existing exclusion behavior, and stay focused — no refactors beyond what's needed.

A complementary workaround for authors who control the document — listing files under `resource_files:` in the Quarto front matter — already works today, but shouldn't be a requirement.

jonkeane:
Add a resource scanner inspired by https://github.com/rstudio/rmarkdown/blob/main/R/html_resources.R

Scans code for references to files, includes them by default.

Finally got this working locally, and it's amazing:

image

Closes #4388


m-- (approved):
played with this locally and on my pet example it's a treat. thank you 🚀

Issue #4017, opened 2026-04-16
bug: Without Quarto locally we don't detect scripts


dotNomad:
If we cannot utilize `quarto inspect` our fallback behavior does not detect [scripts](https://docs.posit.co/connect/user/scripts/) (`.py` or `.R` scripts for Connect) well.

We rely on quarto inspect to recognize an .R file wth the `#' ---` frontmatter and the detector will use that to determine the quarto-static config.

We do have a fallback, but if there is no _quarto.yml then the .R file get rejected by the genNonInspectConfig()

https://github.com/posit-dev/publisher/blob/70e6ec85516a1887c25c6b98ff6c039457ae5de6/extensions/vscode/src/inspect/detectors/quarto.ts#L449-L457

Even without `_quarto.yml` a `.R` entrypoint doesn't get identified.

PR #4392, opened 2026-09-18
Detect and insert Quarto script frontmatter

nealrichardson:
## Summary

Closes both remaining gaps tracked under #3323:

- **#4017** — a `.R`/`.py` script that already has the Quarto frontmatter Connect needs to render it is now detected automatically, even without a working `quarto` CLI or a `_quarto.yml` file.
- **#3323** — choosing the "Script" option for a `.R`/`.py` entrypoint in the manual content-type picker now inserts the required Quarto frontmatter into the file automatically, instead of leaving the user to hand-edit it.

Frontmatter detection (both directions) is anchored to the start of the script (after an optional shebang and/or blank lines), and insertion preserves an existing shebang line rather than displacing it.

There are tests of course, and here's a screen recording of the basic flow working for R and Python scripts. When I run it on a script file that doesn't have front matter, it gets inserted for me. I then deleted the publisher toml and retried just to show that we do detect the front matter as a script if it is present--I just think it's unlikely that anyone will already have it.

https://github.com/user-attachments/assets/ac226d7f-6bca-4cdf-9074-e643d5f43a45


extensions/vscode/src/inspect/helpers/quartoScriptFrontmatter.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Detect whether a bare R or Python script already has the Quarto
+ * frontmatter block Connect requires to render it (see
+ * https://quarto.org/docs/computations/render-scripts.html). R scripts use
+ * roxygen-style `#'` comments; Python scripts use a `# %% [markdown]` cell.
+ * In both cases the block must be delimited by a `---` line on each side,
+ * and must be the first thing in the script (after an optional shebang line
+ * and/or blank lines) — a later, unrelated comment block that happens to be
+ * delimited by `---` lines doesn't count as Quarto frontmatter.
+ */
```
jonkeane, line 18: I tried this out locally and I was a _little bit surprised_ that it edited my script file to insert this block (and then saved it to disk). I more expected it would inject in flight and not change my script on disk. I'm ok for this to change the script on disk, but think I would like at least a note (or addition to the header comment explaining why?)

nealrichardson: I think if it doesn't save then deploy will fail, it needs to be in the code that is shipped to Connect. 

Thanks for taking a look. I had meant to keep this draft until I ran it myself locally but Claude and I got into an argument about something else. I'll try it out and see how it feels to me.

nealrichardson:
Ok this is ready for review now.

extensions/vscode/src/inspect/helpers/quartoScriptFrontmatter.test.ts
```diff
@@ -0,0 +1,257 @@
...
+import { describe, expect, test } from "vitest";
+import * as yaml from "js-yaml";
+import {
+  buildQuartoScriptFrontmatter,
+  hasQuartoScriptFrontmatter,
+  insertQuartoScriptFrontmatter,
+  type ScriptLanguage,
+} from "./quartoScriptFrontmatter";
+
+// Strip the comment prefix off each frontmatter line and parse what's left as
```
zackverham: I think maybe this comment is out of place - looks like it should be for the function below

extensions/vscode/src/inspect/helpers/quartoScriptFrontmatter.ts
zackverham: did we come to a consensus on this one? It makes sense that we need to save the frontmatter change to disk, I do agree that maybe having some type of "inserted by publisher" note would probably be good if nothing else than to be a good citizen with other people's work. 

Other option I suppose would be to do some temp file shenanigans and modify a copy of the original, but that feels like overkill to me. I only bring it up because if there's a bug in here and publisher accidentally mangles someone's working file, that would be pretty bad.

